Issue #92, opened 2015-08-21
Implement Full Outer Join


atifaziz:
FullOuter Join is a relatively common operation that unfortunately is not implemented in linq. It would be great if MoreLinq could provide an implementation.

---
Originally reported on Google Code with ID 92

Reported by @fsateler on 2014-08-27 15:50:35

fsateler:
I had a PR on [bitbucket](https://bitbucket.org/morelinq/morelinq/pull-requests/1/add-fullouterjoin-method/diff). I plan to update it to provide a FullGroupJoin instead. Comments welcome!

fsateler:
Work in progress at branch full-outer branch on my fork https://github.com/fsateler/MoreLINQ/tree/full-outer

atifaziz:
@fsateler [Your work](https://github.com/fsateler/MoreLINQ/tree/full-outer) is in a [branch](https://github.com/fsateler/MoreLINQ/tree/full-outer) that branches off [the one](https://github.com/fsateler/MoreLINQ/tree/mono-todelimited) for #105 and therefore assumes that PR #104 is merged. Your `full-outer` branch should really branch off from **master** at  e0d370e69cb6c8b9f2b38acbef9a220558f3cfcd. I would fix this before any further work in order to make it possible to merge in the end when it's ready.

fsateler:
@atifaziz yes, I haven't rebased on master because otherwise it doesn't build on my linux box. When I file the PR I will rebase on top of master.

atifaziz:
@fsateler [Your implementation](https://github.com/fsateler/MoreLINQ/blob/4f869d8034534dc150eafc76cb480b57649aa5de/MoreLinq/FullGroupJoin.cs) seems very close to the one [posted on StackOverflow](http://stackoverflow.com/a/13503860/6682) by @sehe. Did you take inspiration from it or use it as a starting point and then adapted the code to arrive at a full group join? If so, I just want to make sure that credit is at least paid where it's due (e.g. as code comment).

fsateler:
@atifaziz I'm not sure. I certainly took inspiration from somewhere on the internet (and very likely stackoverflow), but I don't remember the original source.

It seems likely I used this as a starting point when I needed it, so I'll add a comment to the code.

atifaziz:
@fsateler :+1: I think something as simple as the following will do:

    // Inspiration & credit: http://stackoverflow.com/a/13503860/6682

[...]

PR #105, opened 2015-08-22
Mark non cls compliant methods as such

fsateler:
This enables compilation on mono.

The order of the methods was not the same, so I first made the order deterministic. This should minimize diffs in the future.

### DIFF
--- a/MoreLinq/ToDelimitedString.g.cs
+++ b/MoreLinq/ToDelimitedString.g.cs
@@ -328,7 +328,7 @@ namespace MoreLinq
         /// </remarks>
         /// <param name="source">The sequence of items to delimit. Each is converted to a string using the
         /// simple ToString() conversion.</param>
-
+        [CLSCompliant(false)]
         public static string ToDelimitedString(this IEnumerable<sbyte> source)
         {
             return ToDelimitedString(source, null);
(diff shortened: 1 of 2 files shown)


atifaziz:
Thanks for the PR however there are a couple of issues with it.

First, it bundles two unrelated things ([Non-CLS compliant members](https://github.com/fsateler/MoreLINQ/commit/ba142270afca016e8badc72e3e0636b0aaf0aedb) and [deterministic order of ToDelimitedString members](https://github.com/fsateler/MoreLINQ/commit/9da685ed95fda04a379e5c2c58bce992153123b0)) into one PR. This makes it hard to merge and review each independently as they are in the same branch. 

The second and perhaps bigger issue is that I think these should be addressed in [v1](https://github.com/morelinq/MoreLINQ/tree/v1) before being merged into the [master](https://github.com/morelinq/MoreLINQ/tree/master) for v2 because [v1 is still the officially released version](https://www.nuget.org/packages/morelinq/1.1.1).

I'm closing this PR but I'd be happy to take your changes if you could make them on the v1 branch as two separate PRs.

P.S. [Your work](https://github.com/fsateler/MoreLINQ/tree/full-outer) for #92 is in a [branch](https://github.com/fsateler/MoreLINQ/tree/full-outer) that branches off [this one](https://github.com/fsateler/MoreLINQ/tree/mono-todelimited) and therefore assumes this PR has been merged into `master`. It will make it very hard to merge when it's ready so this is a head's up to fix your branching before you work too deep into it.

fsateler:
@atifaziz OK I will rebase on v1 and publish separately.

atifaziz:
Thanks @fsateler and I've tested the non-CLS compliance (you're indeed right) and opened it up as issue #106.

Issue #106, opened 2015-08-23
Some ToDelimitedString members are non-CLS compliant


atifaziz:
When compiling using the C# 6 compiler shipping with VS 2015, the compiler issues the following warnings:

```
Microsoft (R) Visual C# Compiler version 1.0.0.50618
Copyright (C) Microsoft Corporation. All rights reserved.

ToDelimitedString.g.cs(110,72): warning CS3001: Argument type 'IEnumerable<sbyte>' is not CLS-compliant
ToDelimitedString.g.cs(132,72): warning CS3001: Argument type 'IEnumerable<sbyte>' is not CLS-compliant
ToDelimitedString.g.cs(539,72): warning CS3001: Argument type 'IEnumerable<ulong>' is not CLS-compliant
ToDelimitedString.g.cs(443,73): warning CS3001: Argument type 'IEnumerable<ushort>' is not CLS-compliant
ToDelimitedString.g.cs(465,73): warning CS3001: Argument type 'IEnumerable<ushort>' is not CLS-compliant
ToDelimitedString.g.cs(480,71): warning CS3001: Argument type 'IEnumerable<uint>' is not CLS-compliant
ToDelimitedString.g.cs(502,71): warning CS3001: Argument type 'IEnumerable<uint>' is not CLS-compliant
ToDelimitedString.g.cs(517,72): warning CS3001: Argument type 'IEnumerable<ulong>' is not CLS-compliant
```

The Mono (3.12.0) C# compiler gives similar warnings:

```
MoreLINQ/ToDelimitedString.g.cs(110,72): warning CS3001: Argument type `System.Collections.Generic.IEnumerable<sbyte>' is not CLS-compliant
MoreLINQ/ToDelimitedString.g.cs(132,72): warning CS3001: Argument type `System.Collections.Generic.IEnumerable<sbyte>' is not CLS-compliant
MoreLINQ/ToDelimitedString.g.cs(443,73): warning CS3001: Argument type `System.Collections.Generic.IEnumerable<ushort>' is not CLS-compliant
MoreLINQ/ToDelimitedString.g.cs(465,73): warning CS3001: Argument type `System.Collections.Generic.IEnumerable<ushort>' is not CLS-compliant
MoreLINQ/ToDelimitedString.g.cs(480,71): warning CS3001: Argument type `System.Collections.Generic.IEnumerable<uint>' is not CLS-compliant
[...]

PR #107, opened 2015-08-23
Make ToDelimitedString overloads order deterministic

fsateler:
This prevents spurious diffs because the ordering of the declared methods changed

PR #108, opened 2015-08-23
Mark non cls compliant methods as such

fsateler:
This builds on top of #107 in order to minimize diff lines.


fsateler:
This fixes #106

PR #109, opened 2015-08-31
Fix null check in EquiZip

### LINKED ISSUE #94: Patch for /MoreLinq/EquiZip.cs
There was a copy-paste error in the null-checking mechanism.

---
Originally reported on Google Code with ID 94

Reported by @taneltumanski on 2014-10-07 10:29:05

<hr>
 * *Attachment: [EquiZip.cs.patch](https://storage.googleapis.com/google-code-attachments/morelinq/issue-94/comment-0/EquiZip.cs.patch)*

yln:
Fix copy-paste error for null-checking by manually applying changes from patch.

Fixes #94.

### DIFF
--- a/MoreLinq.Test/MoreLinq.Test.csproj
+++ b/MoreLinq.Test/MoreLinq.Test.csproj
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>MoreLinq.Test</RootNamespace>
     <AssemblyName>MoreLinq.Test</AssemblyName>
-    <TargetFrameworkVersion>v3.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <FileUpgradeFlags>
     </FileUpgradeFlags>
@@ -59,6 +59,7 @@
     <Compile Include="FoldTest.cs" />
     <Compile Include="IndexTest.cs" />
     <Compile Include="KeyValuePair.cs" />
+    <Compile Include="NullArgumentTest.cs" />
     <Compile Include="SplitTest.cs" />
     <Compile Include="GroupAdjacentTest.cs" />
     <Compile Include="PairwiseTest.cs" />
--- a/MoreLinq.Test/NullArgumentTest.cs
+++ b/MoreLinq.Test/NullArgumentTest.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Data;
+using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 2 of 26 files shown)


yln:
Github has a nice mechanism for milestone management. If you create a Release 2.0 milestone and assign issues, I can take a look and see if I can help.

atifaziz:
Thanks for this PR. The reason this hadn't been addressed until now (it's a dead simple fix after all) is that this bug should have been picked up by tests for argument checks and those are missing. It would be great to get those in so a copy-paste error by some future maintainer will get caught in time.

atifaziz:
> Github has a nice mechanism for milestone management. If you create a Release 2.0 milestone and assign issues, I can take a look and see if I can help.

These were tagged as [Milestone-Release2.0](https://github.com/morelinq/MoreLINQ/issues?q=label%3AMilestone-Release2.0) on Google Code and have migrated over here. I haven't converted that label into a *GitHub milestone* per say but I'll take a look. The good news is that there's only [one issue open](https://github.com/morelinq/MoreLINQ/labels/Milestone-Release2.0) and I'd rather get the release out without adding new features at the moment. If there's any issues that need to be added, those would be along the lines of omissions, fixes or simple to-dos.

yln:
@atifaziz Can you take a look, now?
This generates tests cases that make sure that all parameters which are of a reference type throw a `ArgumentNullException` (with the correct parameter name). It also checks that all parameters annotated with `[CanBeNull]` really accept `null` as an incoming argument.

Note that I already fixed a few issues found by the tests: 05db824, c0c97ac76, c6172fc5889, 05b96ab54, 6c4d73a4c59a, d33b7d33af0c.

yln:
Some rationale: I don't think that testing null checks manually is worth anyone's time. Also, it doesn't prevent copy & paste errors: if I copy the method header/stub, I will also copy the test file stub.

Other automated alternatives:
* [Code Contracts](https://msdn.microsoft.com/en-us/library/dd264808%28v=vs.110%29.aspx)
* PostSharp, or other AOP
* Roslyn-based solution (similar to current, but using Roslyn API's instead of Reflection API's)

atifaziz:
@yln Ideally your work would have gone into a new issue/PR/branch around a test strategy for all null checking that would have been discussed in advance. Instead its linked to this PR which is specific to testing null arg checking for the `EquiZip` method group. Nonetheless, I'm going to merge your work as an exception instead forcing a workflow for two main reasons: (1) you made granular and atomic commits with clear messages that made it easy for me to review the changes and (2) I could following your line of thought and appreciate the strategy and (solid) implementation. You also managed to fix a number of other issues as a result, so thanks for going the distance and adding considerable value!

The one change I'd like to make is rebase your commits on top of the v1 branch because cherry-picking and back-porting to v1 would be more painful at this stage. That would mean doing a 1.1 release in addition to 2.0. It also has the consequence that you'll have to re-apply c6172fc5889de5cf92b9b28e6a5872ffd761592d and c0c97ac764b24b8e06b821b052fc86beca9a9902 (and possibly others) after merging with the master line because those commits addressed methods that were not in v1.

As for alternatives…

> * [Code Contracts](https://msdn.microsoft.com/en-us/library/dd264808%28v=vs.110%29.aspx)
> * Roslyn-based solution

These sound like an overkill. Specifically, run-time reflection is the right solution because you're testing against the compiled artefact and the test framework supports dynamic enumeration of test cases. Had that not been the case, Roslyn would've come handy to generate the test code, albeit at the expense of considerable complexity due to the number elements and concepts one needs to deal with in the AST and semantic model API.

> * PostSharp, …

Not everyone will have PostSharp, a commercial product, to build the solution on their box.

> …or other AOP

I would argue that AOP is in effect with your solution already except its being applied by the test framework through attributes like [`TestCaseSource`](http://www.nunit.org/index.php?p=testCaseSource&r=2.5).

atifaziz:
@yln I have rebased and published your relevant commits against on top of v1 as the new branch named [`null-checks`](https://github.com/morelinq/MoreLINQ/tree/null-checks). For all further work, please create PR against that branch.

PR #112, opened 2015-09-09
Null checks

### LINKED ISSUE #111: CanBeNull attribute breaks source packages
`CanBeNull` (introduced in 646e4424767ab2055731d5134bedd50a99451868 and with PR #109) is somewhat problematic. While it's not publicly exposed, it will break solutions that use the *source* packages.

The [null test case generator](https://github.com/morelinq/MoreLINQ/blob/b4bc5a7dd504c526642ef73afad828e9e64e36e1/MoreLinq.Test/NullArgumentTest.cs#L70-L81) that uses it should instead enforce the policy that all comparer type of parameters should be nullable, otherwise it's a bug. That covers the 90% uses of the attribute. The remaining are overloads of `ToDelimitedString` where the delimiter parameter can be null. We can (unfortunately) hard-wire that exception unless there's a better alternative altogether to using a marker attribute.

yln:
Removes `[CanBeNull]` and adapts tests.
Changes `MaxBy`/`MinBy` to accept null for `IComparer<T>` and fix docs.
Fixes #111.

### DIFF
--- a/MoreLinq.Test/MaxByTest.cs
+++ b/MoreLinq.Test/MaxByTest.cs
@@ -39,10 +39,9 @@ namespace MoreLinq.Test
         }
 
         [Test]
-        [ExpectedException(typeof(ArgumentNullException))]
         public void MaxByNullComparer()
         {
-            SampleData.Strings.MaxBy(x => x.Length, null);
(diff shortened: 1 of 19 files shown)


atifaziz:
Thanks & in review.

PR #114, opened 2015-09-19
V2 null checks

### LINKED ISSUE #110: Missing null arg checks and tests
Several methods are missing null arg checking and corresponding unit tests. This issue was created from larger efforts started by @yln as part of issue #94 and more specifically PR #109.

The work on this issue is currently taking place on the [null-checks](https://github.com/morelinq/MoreLINQ/tree/null-checks) branch that branches off v1.

At b7e623bec24bba7f9aa75b6876ca8828b12f7025, 227 tests were passing:

```
NUnit-Console version 2.6.2.12296
Copyright (C) 2002-2012 Charlie Poole.
Copyright (C) 2002-2004 James W. Newkirk, Michael C. Two, Alexei A. Vorontsov.
Copyright (C) 2000-2002 Philip Craig.
All Rights Reserved.

Runtime Environment - 
   OS Version: Microsoft Windows NT 6.2.9200.0
  CLR Version: 2.0.50727.8662 ( Net 3.5 )

ProcessModel: Default    DomainUsage: Single
Execution Runtime: net-3.5
...................................................................................................................................................................................................................................
Tests run: 227, Errors: 0, Failures: 0, Inconclusive: 0, Time: 0.8109916 seconds
  Not run: 0, Invalid: 0, Ignored: 0, Skipped: 0
```

As of b4bc5a7dd504c526642ef73afad828e9e64e36e1, 440 tests are passing (some of which may be duplicates to hand-written null check tests):

```
NUnit-Console version 2.6.2.12296
Copyright (C) 2002-2012 Charlie Poole.
Copyright (C) 2002-2004 James W. Newkirk, Michael C. Two, Alexei A. Vorontsov.
Copyright (C) 2000-2002 Philip Craig.
All Rights Reserved.

Runtime Environment - 
   OS Version: Microsoft Windows NT 6.2.9200.0
  CLR Version: 2.0.50727.8662 ( Net 3.5 )

ProcessModel: Default    DomainUsage: Single
Execution Runtime: net-3.5
........................................................................................................................................................................................................................................................................................................................................................................................................................................................
Tests run: 440, Errors: 0, Failures: 0, Inconclusive: 0, Time: 0.8280372 seconds
  Not run: 0, Invalid: 0, Ignored: 0, Skipped: 0
```

yln:
Fixes #110

### DIFF
--- a/MoreLinq.Test/NullArgumentTest.cs
+++ b/MoreLinq.Test/NullArgumentTest.cs
@@ -141,7 +141,7 @@ namespace MoreLinq.Test
         {
             if (type == typeof (int)) return 7; // int is used as size/length/range etc. avoid ArgumentOutOfRange for '0'.
             if (type == typeof (string)) return "";
-            if (typeof(IEnumerable<int>).IsAssignableFrom(type)) return new[] { 1, 2, 3 }; // Provide non-empty sequence for MinBy/MaxBy.
+            if (type == typeof(IEnumerable<int>)) return new[] { 1, 2, 3 }; // Provide non-empty sequence for MinBy/MaxBy.
             if (type.IsArray) return Array.CreateInstance(type.GetElementType(), 0);
             if (type.IsValueType || HasDefaultConstructor(type)) return Activator.CreateInstance(type);
             if (typeof(Delegate).IsAssignableFrom(type)) return CreateDelegateInstance(type);
--- a/MoreLinq.Test/NullArgumentTest.cs
+++ b/MoreLinq.Test/NullArgumentTest.cs
@@ -128,7 +128,7 @@ namespace MoreLinq.Test
         private static bool CanBeNull(ParameterInfo parameter)
         {
             var nullableTypes = new[] { typeof (IEqualityComparer<>), typeof (IComparer<>) };
-            var nullableParameters = new[] { "ToDataTable.expressions", "ToDelimitedString.delimiter", "Trace.format" };
+            var nullableParameters = new[] { "Assert.errorSelector", "ToDataTable.expressions", "ToDelimitedString.delimiter", "Trace.format" };
 
             var type = parameter.ParameterType;
             type = type.IsGenericType ? type.GetGenericTypeDefinition() : type;


yln:
The following is just my (subjective) opinion. I am fine either way.

0. I agree with "an overload with additional parameters should behave as if those parameters are optional".
1. As I see it, the default behavior is throwing a `InvalidOperationException` and technically the default value is not null, but `delegate { return null; }`, but that is just an implementation detail. (The fact that the default behavior is implemented via a delegate that returns null bothers me a little too.)
2. Why provide two ways to do the same thing?  
`source.Assert(predicate)` and `source.Assert(predicate, null)`  
For this method `errorSelector` will almost never be a computed value (which would force us to put an if/else in our code).
3. I prefer the "everything-non-null" approach (unless explicitly stated). I can't tell from the method signature that null is okay here. It's not even stated in the docs. Can the `predicate` be null too? (I agree that this particular method wouldn't make much sense then, but in general ...)

I agree that there is the exception for `IEqualityComparer<>` and `IComparer<>`, because we need to be consistent with the .NET framework (a design decision made 10? years ago).
I disagree with the other exceptions (maybe 2.x is a good point to introduce minor breaking changes to clean them up?) and especially with adding new exceptions.

atifaziz:
Ah, I'm sorry, I completely misread that you were talking about `Assert` and not `AssertCount` because none of your points was lining up to what I was seeing in the code. Anyway, since we agree on the first point, let's move on to the remaining ones…

> The fact that the default behavior is implemented via a delegate that returns null bothers me a little too

This was to avoid duplication in the code by way of having the policy at one location.

> Why provide two ways to do the same thing?

You could see it that way but the actual intention was that faced with a faulty `errorSelector` implementation (one returning a `null` for exception), `Assert` did the right thing. We can now argue whether hiding, making or patching a faulty implementation is a good thing or not. I do see that `AssertCount` and `Assert` are inconsistent here as the former will fault with `NullReferenceException` when it gets to throwing the exception and the delegate has returned a `null`.

> I can't tell from the method signature that null is okay here. It's not even stated in the docs.

We've talked about this. Optionality is not covered by the type system but in this case, it's about covering a faulty implementation that didn't seem worth bloating docs for. Why? Well, the `errorSelector` enables three things: (1) keep raising `InvalidOperationException` but change the message, (2) change the raise exception type or (3) both. If the delegate implementation intended to change the message, the exception type or both and yet by error returned `null` then the defaulted implementation of `Assert` would yield a different result. Either it would be caught by unit tests or observed behaviour at runtime. That all said…

> I prefer the "everything-non-null" approach

I can go with this for the return value of `errorSelector` but the entire delegate should remain optional with respect to the other overload(s).

atifaziz:
Tests look good:

```
NUnit-Console version 2.6.2.12296
Copyright (C) 2002-2012 Charlie Poole.
Copyright (C) 2002-2004 James W. Newkirk, Michael C. Two, Alexei A. Vorontsov.
Copyright (C) 2000-2002 Philip Craig.
All Rights Reserved.

Runtime Environment - 
   OS Version: Microsoft Windows NT 6.2.9200.0
  CLR Version: 2.0.50727.8662 ( Net 3.5 )

ProcessModel: Default    DomainUsage: Single
Execution Runtime: net-3.5
..................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Tests run: 722, Errors: 0, Failures: 0, Inconclusive: 0, Time: 3.5310994 seconds
  Not run: 0, Invalid: 0, Ignored: 0, Skipped: 0
```

@yln Thank for all your help with this and keen observations. I'm merging your latest contributions and we can continue the thread on `Assert` in another issue if you think it's important to reconsider for 2.0.

P.S. I pasted this response in the wrong box, and sadly, it became part of the commit message detail. :see_no_evil:

atifaziz:
:warning: I've done the unthinkable and re-pushed (read *modified history*) the merge as ba2274fdd483abdb013d9d0842d5d3504786ce81 with the commit message fixed. There are only two known forks at this point in time so I've exceptionally gone ahead and made the change rather than live with the silly commit error forever as part of the repo history.

:speak_no_evil:

PR #116, opened 2015-09-30
V2 null consistency

yln:
Remove "nullable" exceptions (other than `IEqualityComparer` and `IComparer`) for consistency. Refine implementations and update docs.
- Assert(errorSelector): now consistent with AssertCount
- ToDataTable(expressions): params array, so no reason why this should ever be null; there wasn't even a test to cover the null case
- Trace(format): format string should never be null (never computed, fixed per call-site); also simplify implementation (`string.Format(format, args)` handles null just fine)
- ToDelimitedString(delimiter): Make it more explicit what happens; also update docs.

### DIFF
--- a/MoreLinq.Test/AssertTest.cs
+++ b/MoreLinq.Test/AssertTest.cs
@@ -59,7 +59,7 @@ namespace MoreLinq.Test
         }
 
         [Test]
-        [ExpectedException(typeof(InvalidOperationException))]
+        [ExpectedException(typeof(NullReferenceException))]
         public void AssertSequenceWithInvalidElementsAndCustomErrorReturningNull()
         {
             new[] { 2, 4, 6, 7, 8, 9 }.Assert(n => n % 2 == 0, _ => null).Consume();
--- a/MoreLinq.Test/NullArgumentTest.cs
+++ b/MoreLinq.Test/NullArgumentTest.cs
@@ -128,7 +128,7 @@ namespace MoreLinq.Test
         private static bool CanBeNull(ParameterInfo parameter)
         {
             var nullableTypes = new[] { typeof (IEqualityComparer<>), typeof (IComparer<>) };
-            var nullableParameters = new[] { "ToDataTable.expressions", "ToDelimitedString.delimiter", "Trace.format" };
+            var nullableParameters = new[] { "ToDelimitedString.delimiter", "Trace.format" };
 
             var type = parameter.ParameterType;
             type = type.IsGenericType ? type.GetGenericTypeDefinition() : type;
--- a/MoreLinq.Test/NullArgumentTest.cs
+++ b/MoreLinq.Test/NullArgumentTest.cs
@@ -128,13 +128,11 @@ namespace MoreLinq.Test
         private static bool CanBeNull(ParameterInfo parameter)
         {
             var nullableTypes = new[] { typeof (IEqualityComparer<>), typeof (IComparer<>) };
-            var nullableParameters = new string[0];
 
             var type = parameter.ParameterType;
             type = type.IsGenericType ? type.GetGenericTypeDefinition() : type;
-            var param = parameter.Member.Name + "." + parameter.Name;
 
-            return nullableTypes.Contains(type) || nullableParameters.Contains(param);
+            return nullableTypes.Contains(type);
         }
 
         private static object CreateInstance(Type type)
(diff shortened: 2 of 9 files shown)


yln:
I guess this was auto-closed because the base-branch went away. Any comments on the changes?

atifaziz:
That was my bad. I was too eager to close the branch `v2-null-checks` branch after merging with `master`, not realising you still had `v2-null-consistency` open over it.

atifaziz:
I'm not too happy with d75ed05d805e0af8eb45b9932b4e825b4303f168 and we've talked about this at some length in #114. Overloads with mere additional arguments should act as if they are optional and provide identical behaviour to those overloads where the arguments are omitted. Throwing an exception means the result cannot be computed because it would be undefined and that's not true.

The goal here is not to simplify the test code.

yln:
`source.ToDelimitedString()` is not equal to `source.ToDelimitedString(null)`, it is equal to `source.ToDelimitedString(CultureInfo.CurrentCulture.TextInfo.ListSeparator)`.

I can't imagine any scenario in which I would want to call `source.ToDelimitedString(null)`. `delimiter` will most likely not be a computed value. Even if it is a computed value, it is much better if the user defines what the fallback for null is. 

Which code is more explicit/less surprising?
```c#
var delimiter = ComputeDelimiter();
var x = source.ToDelimitedString(delimiter);
```

```c#
var delimiter = ComputeDelimiter() ?? ",";
var x = source.ToDelimitedString(delimiter);
```
As a casual reader I have no chance of knowing that delimiter will be replaced with `CultureInfo.CurrentCulture.TextInfo.ListSeparator` (which even varies by system) in variant 1.
(Yes, of course, I would also argue for disallowing `ComputeDelimiter()` to return null.)


The rest of the changes are okay?

fsateler:
I'm missing a changes file where it is documented this breaking behavior change (is this done only on the nuspec file at release time?).

FWIW, I agree with @yln that `ToDelimitedString(null)` makes no sense. I don't think "default value for a parameter" needs to be equal to "default value for the parameter type".

yln:
>I'm missing a changes file where it is documented this breaking behavior change (is this done only on the nuspec file at release time?).

I can provide that if you guys tell me which file and format it should have.

fsateler:
@yln I don't think such a file exists currently.

atifaziz:
> `source.ToDelimitedString()` is not equal to `source.ToDelimitedString(null)`, it is equal to `source.ToDelimitedString(CultureInfo.CurrentCulture.TextInfo.ListSeparator)`.

`source.ToDelimitedString()` is equal to `source.ToDelimitedString(null)` as both exhibit the same *runtime* behaviour as if you wrote `source.ToDelimitedString(CultureInfo.CurrentCulture.TextInfo.ListSeparator)` explicitly.

> As a casual reader I have no chance of knowing that delimiter will be replaced with CultureInfo.CurrentCulture.TextInfo.ListSeparator (which even varies by system) in variant 1.

Casual readers have no chance of knowing what `DateTime.Now.ToString()`, `DateTime.Now.ToString(null)` or `string.Format(DateTime.Now)` do if they haven't the slightest clue or ever read a line or two of the docs; or define *casual reading*. Yes, they give you string representations of the date & time but the exact format and output is defined by the environment. In general, such code is found in presentation & UI layers. In other cases where you need certainty, you supply arguments explicitly.

The `ToDelimitedString` overloads accepting a delimiter allow `null` in much the same way that many text-formatting methods in BCL permit a `null` for format specification, `IFormatProvider` and so forth. If you want to force a delimiter then we should remove the simple overload that takes no arguments because maintaining consistency in the optionality of the parameters across the overloads is more about contractual correctness than implementation tactics.

That all said, is `ToDelimitedString()` at all useful? Not really. I actually added it just because it could be done & `TextInfo.ListSeparator` was available as a reasonable choice. I personally only ever use it when doing casual experimentation or demos in, say, LINQPad. Even then I usually I prefer to specify the delimiter in case an experiment is so successful that I end up taking it to production. I don't want to have to unnecessarily revisit the code & add delimiters back. When I added the zero-arity overload, I maintained the optionality of the argument for consistency. So again, I think the real question to ask is whether to keep or drop the `ToDelimitedString` nullary overload, not who in their right mind would compute the delimiter with a possibility of `null`. I would personally opt to drop it but I don't find it's doing any harm at the moment either, especially not worth delaying the 2.0 release & this PR has thrown a :wrench: at that (which is a shame because given your help & contributions, I wanted to reciprocate by getting 2.0 out the door).

On a further aside on optionality, do bear in mind that there is also the spirit & idea in LINQ (especially with [standard query operators](https://msdn.microsoft.com/en-us/library/bb394939.aspx)) to be able to project or translate quoted queries into foreign systems. The simpler & non-explicit overloads do help there where the exact runtime semantics are understood to be defined by the target environment. I'm certainly not suggested that  `ToDelimitedString` was designed with that scenario in mind or that I'd like to maintain it as-is to exclude potential opportunities.

yln:
Okay, you convinced me. I don't have a strong opinion any more. ;)

PR #117, opened 2015-09-30
Scan<TSource>: do not throw on empty sequence

### LINKED ISSUE #97: Scan should not throw on empty sequences
The `Scan<T>` implementation throws when an empty sequence is found. I believe the correct behavior should be to yield break in that case.

This would mean a different behavior from the `Scan<T, TState>` implementation, but it is already different. Without a seed, an `IEnumerable` of size N will yield an `IEnumerable` of the same size, while the version with seed adds an extra element.

So making a 0-length `IEnumerable` result in a 0-length `IEnumerable` seems like the right behavior to me.

---
Originally reported on Google Code with ID 97

Reported by @fsateler on 2014-12-30 13:15:32

fsateler:
Instead, yield break and thus return an empty sequence.

Fixes #97

### DIFF
--- a/MoreLinq/Scan.cs
+++ b/MoreLinq/Scan.cs
@@ -65,7 +65,7 @@ namespace MoreLinq
             using (var i = source.GetEnumerator())
             {
                 if (!i.MoveNext())
-                    throw new InvalidOperationException("Sequence contains no elements.");
+                    yield break;
 
                 var aggregator = i.Current;
 


atifaziz:
Thanks. I rebased this on to `v1` as 7ca36290dafb5f616c616bc8eb07ef61c8dfbdbd as the fix would be useful there for a maintenance release. I'll later merge it also into `master`. You can delete your `patch-1` branch.

fsateler:
@atifaziz thanks!

PR #118, opened 2015-10-01
Add FullGroupJoin method

### LINKED ISSUE #92: Implement Full Outer Join

fsateler:
Fixes #92 

Reviews welcome!

### DIFF
--- a/MoreLinq/FullGroupJoin.cs
+++ b/MoreLinq/FullGroupJoin.cs
@@ -1,4 +1,4 @@
-﻿#region License and Terms
+#region License and Terms
 // MoreLINQ - Extensions to LINQ to Objects
 // Copyright (c) 2008 Jonathan Skeet. All rights reserved.
 // 
--- a/MoreLinq/FullGroupJoin.cs
+++ b/MoreLinq/FullGroupJoin.cs
@@ -30,6 +30,9 @@ namespace MoreLinq
         /// </summary>
         /// <remarks>
         /// This operator uses deferred execution and streams the results.
+        /// The results are yielded in the order of the elements found in the first sequence
+        /// followed by those found only in the second. In addition, the callback responsible
+        /// for projecting the results is supplied with sequences which preserve their source order.
         /// </remarks>
         /// <typeparam name="TFirst">The type of the elements in the first input sequence</typeparam>
         /// <typeparam name="TSecond">The type of the elements in the first input sequence</typeparam>
@@ -56,6 +59,9 @@ namespace MoreLinq
         /// </summary>
         /// <remarks>
         /// This operator uses deferred execution and streams the results.
+        /// The results are yielded in the order of the elements found in the first sequence
+        /// followed by those found only in the second. In addition, the callback responsible
+        /// for projecting the results is supplied with sequences which preserve their source order.
         /// </remarks>
         /// <typeparam name="TFirst">The type of the elements in the first input sequence</typeparam>
         /// <typeparam name="TSecond">The type of the elements in the first input sequence</typeparam>
@@ -95,16 +101,19 @@ namespace MoreLinq
         {
             comparer = comparer ?? EqualityComparer<TKey>.Default;
 
-            var alookup = first.ToLookup(firstKeySelector, comparer);
-            var blookup = second.ToLookup(secondKeySelector, comparer);
-            var keys = alookup.Select(p => p.Key).Union(blookup.Select(p => p.Key), comparer);
+            var alookup = OrderPreservingLookup<TKey,TFirst>.CreateForJoin(first, firstKeySelector, comparer);
+            var blookup = OrderPreservingLookup<TKey, TSecond>.CreateForJoin(second, secondKeySelector, comparer);
 
-            var join = keys.Select(key => resultSelector(key, alookup[key], blookup[key]));
+            var seenKeys = new HashSet<TKey>(comparer);
 
-            // We do the iteration + yield return to delay executing the earlier ToLookup
-            // until the result is enumerated
-            foreach (var item in join) {
-                yield return item;
+            var allKeys = alookup.Select(p => p.Key).Concat(blookup.Select(p => p.Key));
+
+            foreach (var key in allKeys) {
+                if (seenKeys.Contains(key)) {
+                    continue;
+                }
+                seenKeys.Add(key);
+                yield return resultSelector(key, alookup[key], blookup[key]);
             }
         }
     }
--- a/MoreLinq/FullGroupJoin.cs
+++ b/MoreLinq/FullGroupJoin.cs
@@ -112,7 +112,7 @@ namespace MoreLinq
             }
 
             foreach (var b in blookup) {
-                if (!seenKeys.Add(b.Key))
+                if (seenKeys.Contains(b.Key))
                     continue;
                 yield return resultSelector(b.Key, alookup[b.Key], b);
             }
--- a/MoreLinq/FullGroupJoin.cs
+++ b/MoreLinq/FullGroupJoin.cs
@@ -114,7 +114,8 @@ namespace MoreLinq
             foreach (var b in blookup) {
                 if (seenKeys.Contains(b.Key))
                     continue;
-                yield return resultSelector(b.Key, alookup[b.Key], b);
+                // We can skip the lookup because we are iterating over keys not found in the first sequence
+                yield return resultSelector(b.Key, Enumerable.Empty<TFirst>(), b);
             }
         }
     }
--- a/MoreLinq/FullGroupJoin.cs
+++ b/MoreLinq/FullGroupJoin.cs
@@ -102,19 +102,13 @@ namespace MoreLinq
 
             var alookup = Lookup<TKey,TFirst>.CreateForJoin(first, firstKeySelector, comparer);
             var blookup = Lookup<TKey, TSecond>.CreateForJoin(second, secondKeySelector, comparer);
-
-            bool firstNotEmpty = alookup.Count > 0;
-            HashSet<TKey> seenKeys = null;
-            if (firstNotEmpty) {
-                seenKeys = new HashSet<TKey>(comparer);
-                foreach (var a in alookup) {
-                    yield return resultSelector(a.Key, a, blookup[a.Key]);
-                    seenKeys.Add(a.Key);
-                }
+            
+            foreach (var a in alookup) {
+                yield return resultSelector(a.Key, a, blookup[a.Key]);
             }
 
             foreach (var b in blookup) {
-                if (firstNotEmpty && seenKeys.Contains(b.Key))
+                if (alookup.Contains(b.Key))
                     continue;
                 // We can skip the lookup because we are iterating over keys not found in the first sequence
                 yield return resultSelector(b.Key, Enumerable.Empty<TFirst>(), b);
--- a/MoreLinq/FullGroupJoin.cs
+++ b/MoreLinq/FullGroupJoin.cs
@@ -20,7 +20,6 @@ namespace MoreLinq
     using System;
     using System.Collections.Generic;
     using System.Linq;
-    using System.Text;
 
     // Inspiration & credit: http://stackoverflow.com/a/13503860/6682
     static partial class MoreEnumerable
--- a/MoreLinq/FullGroupJoin.cs
+++ b/MoreLinq/FullGroupJoin.cs
@@ -1,0 +1,111 @@
+﻿#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


MoreLinq/FullGroupJoin.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This operator uses deferred execution and streams the results.
+        /// </remarks>
+        /// <typeparam name="TFirst">The type of the elements in the first input sequence</typeparam>
+        /// <typeparam name="TSecond">The type of the elements in the first input sequence</typeparam>
+        /// <typeparam name="TKey">The type of the key to use to join</typeparam>
+        /// <typeparam name="TResult">The type of the elements of the resulting sequence</typeparam>
+        /// <param name="first">First sequence</param>
+        /// <param name="second">Second secuence</param>
```
sholland1: misspelled sequence as "secuence"

MoreLinq/FullGroupJoin.cs
```diff
@@ -0,0 +1,111 @@
...
```
sholland1: misspelled sequence as "secuence"

fsateler:
@sholland1 Thanks, fixed in a followup commit. I can rebase and squash when this is deemed ok for inclusion.

atifaziz:
Most of this looks good to me, except like [GroupJoin][GroupJoin], I would preserve the order:

> **GroupJoin** preserves the order of the elements of *outer*, and for each element of *outer*, the order of the matching elements from *inner*.

I think this is easy to do. Instead of union-ing the keys & then iterating over them, use a `HashSet<TKey>` to collect visited keys. Then walk over the left keys (adding to the hash set) and then unvisited right keys.

  [GroupJoin]: https://msdn.microsoft.com/en-us/library/vstudio/system.linq.enumerable.groupjoin(v=vs.100).aspx

atifaziz:
Also want to add that while [`ToLookup`][ToLookup] is not documented to return the keys in order, the [current implementation does][ToLookupRefSrc]. However, if we don't want to count on the implementation changing from under our feet in some future version or patch, it would be perfectly acceptable to copy the [implementation from CoreFX][CoreFxToLookup] since it is published under the [MIT License][corefxlic] and that's [compatible with Apache License, Version 2.0][ALv2-compat] under which this [project is licensed][ALv2].


  [ToLookup]: https://msdn.microsoft.com/en-us/library/vstudio/system.linq.enumerable.tolookup(v=vs.100).aspx
  [ToLookupRefSrc]: http://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,2056
  [CoreFxToLookup]: https://github.com/dotnet/corefx/blob/6f1c2a86fb8fa1bdaee7c6e70a684d27842d804c/src/System.Linq/src/System/Linq/Enumerable.cs#L3230-L3403
  [corefxlic]: https://github.com/dotnet/corefx/blob/6f1c2a86fb8fa1bdaee7c6e70a684d27842d804c/LICENSE
  [ALv2-compat]: http://apache.org/legal/resolved.html#category-a
  [ALv2]: https://github.com/morelinq/MoreLINQ/blob/ec4bbd3c7ca61e3a98695aaa2afb23da001ee420/COPYING.txt

fsateler:
OK, I don't think having a second implementation of `ToLookup` is the best course of action. I will see if I can come up with something that can guarantee ordering.

atifaziz:
> I don't think having a second implementation of ToLookup is the best course of action

You didn't say why. Curious to hear the downsides you are thinking of.

fsateler:
@atifaziz I think that either we should (a) expose this Lookup implementation because it is generally useful, or (b) don't use it because it is overkill for the use at hand. Wording suggestions for the order promise in the xmldoc welcome, everything I have tried reads terribly confusing.

atifaziz:
@fsateler We can always expose the `ToLookup` implementation in a subsequent version (and open it as another pending issue). Right now its inclusion is a question of getting `FullOuterJoin` complete & closed.

Does your comment about overkill relate to preserving key order or including the `ToLookup` implementation? A less efficient alternative would be to sort your keys after the union but the downside of that approach is that those who don't care about the order of keys being preserved will always pay the cost of sort.

> Also, wording suggestions for the order promise in the xmldoc welcome

Use the same as `GroupJoin`?

> **GroupJoin** preserves the order of the elements of *outer*, and for each element of *outer*, the order of the matching elements from *inner*.

Reads very clear.

fsateler:
@atifaziz I meant overkill to include a full Lookup implementation just to guard against possibly non-order-preserving implementations. This can be implemented without implementing a whole new Lookup class.

            var agroups = first.GroupBy(firstKeySelector, comparer);
            // Use dictionary because lookups are immutable
            var bgroups = second.GroupBy(secondKeySelector, comparer).Index().ToDictionary(p => p.Value.Key);

            foreach (var firstGroup in agroups) {
                IEnumerable<TSecond> secondGroup = Enumerable.Empty<TSecond>();
                KeyValuePair<int, IGrouping<TKey, TSecond>> pair;
                if (bgroups.TryGetValue(firstGroup.Key, out pair)) {
                    secondGroup = pair.Value;
                    bgroups.Remove(firstGroup.Key);
                }
                yield return resultSelector(firstGroup.Key, firstGroup, secondGroup);
            }
            foreach (var secondGroup in bgroups.Values.OrderBy(p => p.Key)) {
                yield return resultSelector(secondGroup.Value.Key, Enumerable.Empty<TFirst>(), secondGroup.Value);
            }

At this point I think the discussion on this has no further benefits (ie, if I didn't explain myself clearly/couldn't convince you then it is unlikely I will ;). I defer to you as maintainer: are you ok with the above implementation (or a variation thereof), or would you rather use a custom `Lookup`? I will update the PR accordingly.


As for the xmldoc,

> Use the same as GroupJoin?

The problem is that there are 2 properties we need to make clear: 1. order is preserved on keys, as found on first and then on second, and 2. within each group, order is preserved as found in the original sequence. The `GroupJoin` documentation does not feel like it explains the first property very well (because in `GroupJoin` the keys not on inner will be lost!). And whatever I write feels very clunky.

Perhaps something like:

> The results are yielded in the order they are found in the `first` sequence, and then elements found only on `second` sequence. Within each result, the ordering of the elements of each sequence are preserved.

atifaziz:
@fsateler The implementation you presented has the downside I mentioned: it makes everyone pay the cost of a sort. That's not acceptable when [a better solution is already available][corefx-lookup]. I would therefore opt for including `ToLookup` from CoreFX in the interest of time & effort, not to mention that it's field-tested. We don't need to expose it as that can be a separate issue opened if & when needed.

> overkill to include a full Lookup implementation just to guard against possibly non-order-preserving implementations

Except when you put code in production that relies on an undocumented & incidental behaviour and the result changes from underneath you (without recompilation) with some update, patch or platform change, then I can reassure you that the the countless hours one will spend hunting down the obscure cause of the issue will be painful. [Ivan Danilov claims that `ToLookup` in .NET 4.5.1 on Windows 8 does not preserve the order](http://stackoverflow.com/questions/5489987/linq-full-outer-join#comment49887095_30289550) but I haven't verified that and never mind it anyhow. The [CoreFX `ToLookup` implementation][corefx-lookup] weighs less than 200 LOC and that doesn't seem like an overkill given the waste of human life spent debugging that it'll prevent. :) What might seem like an overkill is to try and implement it yourself along with unit tests and what not, unless of course you have the time to spare & have a better implementation in mind.

Also bear in mind that all these decisions can be reverted should they turn out to be excessive.


  [corefx-lookup]: https://github.com/dotnet/corefx/blob/6f1c2a86fb8fa1bdaee7c6e70a684d27842d804c/src/System.Linq/src/System/Linq/Enumerable.cs#L3230-L3403

atifaziz:
> The results are yielded in the order they are found in the `first` sequence, and then elements found only on `second` sequence. Within each result, the ordering of the elements of each sequence are preserved.

How about this?

> The results are yielded in the order of the elements found in the first sequence followed by those found only in the second. In addition, the callback responsible for projecting the results is supplied with sequences which preserve their source order.

The main change is to the following that could be misleading:

> Within each result, the ordering of the elements of each sequence are preserved.

The projector could change the order, for example, if it does a sort before returning the joined result. The only thing that can be said is that the projector will receive those sequences in their source order.

fsateler:
@atifaziz I like the xmldoc change. Will apply.

fsateler:
@atifaziz 

> What might seem like an overkill is to try and implement it yourself along with unit tests and what not

That was my worry precisely. Even if the initial implementation is copied over from CoreFX, bugs always happen and the implementations are likely to diverge over time (ie, maintainance time is never zero).

But I see how you would prefer to borrow a battle tested implementation that is known to work.

I will include the lookup implementation (plus tests) in an updated PR. Would you prefer a new PR or that I force-push to this one?

atifaziz:
> bugs always happen and the implementations are likely to diverge over time (ie, maintainance time is never zero).

Yes I agree, but here's how I think that risk is low. The implementation is less than 200 LOC and I could understand it in under 5 minutes. The entries are added and maintained as a hybrid between a hashtable for fast lookup and nodes linked by insertion order within the same table. If there's a bug, the implementation isn't convoluted by its cleverness and so I'm confident it can be fixed in a reasonable amount of time without an overhaul. On the other hand, if the bug has been fixed in CoreFX then one can replay the changes since CoreFX is open source and one can follow the history.

> Would you prefer a new PR or that I force-push to this one?

Why force-push at all? Just keep building on top of the commits in this PR. There's no dirty history here. No work should be lost. No need to squash or rebase.

fsateler:
I have copied the Lookup implementation, but changed the name to OrderPreservingLookup, to distinguish it from the BCL implementation.

@atifaziz would you prefer me to add a new file for the MIT license, append to the COPYING file or just keep as a comment in the new Lookup class?

Also, the MIT license terms require distributors to keep the copyright notice, so the MS copyright notice should end up in the nuget package somehow. Hints on how to do that welcome.

fsateler:
@atifaziz I have addressed your indications in a follow-up commit.

atifaziz:
Thanks for tackling that fast. However, could you please break e5a37b3a8727e786a30e5613b0b3dd41e1d8c0ef into 4 commits (as they are unrelated), each with the comment that you've put in the details of that one rolled-up commit. Besides making the commits atomic, it also makes reviewing easier.

fsateler:
@atifaziz done.

I realized during rebasing that I had misconfigured my git client and had the wrong email address (one that should only be used for a specific internal project) for several commits. I have altered the email information (which causes the commit hashes to change), but the commits are otherwise identical.

atifaziz:
@fsateler Splendid. :+1: Will review.

atifaziz:
@fsateler Can we remove the UTF-8 BOM from the following files (as it is safe to do)?

- `MoreLinq\FullGroupJoin.cs`
- `MoreLinq\Lookup.cs`
- `MoreLinq.Test\FullGroupJoinTest.cs`

fsateler:
OK, will do as well.

If only visual studio allowed to be configured to never use BOMs...

fsateler:
@atifaziz I have updated the branch with all changes.

I'm not sure if I like how the code looks like after commit https://github.com/fsateler/MoreLINQ/commit/761e445d8f8c0a2c85085a41cc79fe28563114ed, it looks quite convoluted for a case I'm not sure is common (empty first sequence).

atifaziz:
I like how [a first suggestion for optimization](https://github.com/morelinq/MoreLINQ/pull/118#commitcomment-14040817) has led to [some](https://github.com/fsateler/MoreLINQ/commit/fe498fbf76fafda224c751200e787d0293e1a1dc) [other](https://github.com/fsateler/MoreLINQ/commit/06c3aefa52d7ba74efc3f35e93ab069ad38f4e6d) [sweet](https://github.com/fsateler/MoreLINQ/commit/761e445d8f8c0a2c85085a41cc79fe28563114ed) ones. Great job. :+1: And I agree that it gets ugly by 761e445d8f8c0a2c85085a41cc79fe28563114ed but perhaps because we don't need the `HashSet` at all. The first sequence already has the key-set built via the first `Lookup` and we use that for containment checks when looping over keys of the second sequence. In other words:

```c#
var alookup = Lookup<TKey,TFirst>.CreateForJoin(first, firstKeySelector, comparer);
var blookup = Lookup<TKey, TSecond>.CreateForJoin(second, secondKeySelector, comparer);

foreach (var a in alookup)
    yield return resultSelector(a.Key, a, blookup[a.Key]);

foreach (var b in blookup) {
    if (alookup.Count > 0 && alookup.Contains(b.Key))
        continue;
    // We can skip the lookup because we are iterating over keys not found in the first sequence
    yield return resultSelector(b.Key, Enumerable.Empty<TFirst>(), b);
```

So gone is another whole set of allocations! The only reason for the `alookup.Count > 0` micro-optimization now is that `Contains` doesn't short-circuit the hashing & lookup if empty. If we move that into the implementation of `Contains` itself then I think we have `FullOuterJoin` reading short, simple, sweet & optimal!

fsateler:
Ah, excellent suggestion :+1: ! (it seems so obvious now :) ) Now the code looks much, much better.

I decided to short-circuit the Lookup.Contains function instead of GetGrouping. It is sufficient for our purposes, and it doesn't complicate anything.

fsateler:
I'm afraid I've never manipulated a nuget package before. I think I added it correctly, but please verify.

fsateler:
Better late than never, I put the copyright information in the right element now.

Issue #58, opened 2015-08-21
MoreEnumerable.AtLeast and MoreEnumerable.Exactly


atifaziz:
Count() enumerates the entire sequence before returning, AssertCount() asserts that there are exactly a specific number of elements and throws an exception if not, Any() tells me if there are any elements in the sequence at all, but none of these provide the simple behaviour of telling that there are or are not at least x number of elements or exactly x number of elements.

AtLeast(int x) should iterate the sequence enough to count that there are at least x elements. If there are at least x elements, it returns true, otherwise false.

Exactly(int x) should iterate the sequence just enough to count upto x or x+1, if x+1 exists. If the sequence is exactly x elements long, it returns true, otherwise false.

---
Originally reported on Google Code with ID 58

Reported by `jeffmagic` on 2010-01-29 21:28:26

sholland1:
The `Exactly` method isn't implemented on this yet.
I was thinking I could implement `Exactly` after finishing up `AtLeast`.

atifaziz:
@sholland1 Kindly submit that as a separate & new issue, cross-referencing this one for background & quoting if necessary. Thanks.

Arithmomaniac:
The current AtLeast has an optimization for ICollection. It should have one for IReadOnlyCollection instead/as well.

PR #120, opened 2015-10-10
Implement AtLeast extension method #58

sholland1:
If this is good for this method, I can implement the Exactly method, too.

### DIFF
--- a/MoreLinq.Test/AtLeastTest.cs
+++ b/MoreLinq.Test/AtLeastTest.cs
@@ -1,4 +1,4 @@
-﻿#region License and Terms
+#region License and Terms
 // MoreLINQ - Extensions to LINQ to Objects
 // Copyright (c) 2008 Jonathan Skeet. All rights reserved.
 // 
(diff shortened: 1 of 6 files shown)


sholland1:
I added the optimization for ICollections.
Is there anything else I could do to wrap this up?

sholland1:
You're right. This stack overflow [answer](http://stackoverflow.com/a/7969468) corroborates this.

atifaziz:
@sholland1 Could you also remove the UTF-8 BOM marker from the following files?

* `MoreLinq\AtLeast.cs`
* `MoreLinq.Test\AtLeastTest.cs`

sholland1:
I have finished implementing each of the improvements you suggested.

sholland1:
I used the existing `UnenumerableList` class to improve the test.
While I was looking through the test project, I found the `InfiniteSequence` class. I used it to refactor these tests a little bit.
In the future, it'd probably be worth it to separate the tests and test helpers.

PR #121, opened 2015-10-15
Add overload of ForEach that provides the source element index to the action

tg73:
Includes unit tests.

### DIFF
--- a/MoreLinq.Test/ForEachTest.cs
+++ b/MoreLinq.Test/ForEachTest.cs
@@ -35,7 +35,7 @@ namespace MoreLinq.Test
         [ExpectedException(typeof(ArgumentNullException))]
         public void ForEachNullAction()
         {
-            new[] { 1, 2, 3 }.ForEach(null);
+            new[] { 1, 2, 3 }.ForEach((Action<int>)null);
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


fsateler:
This is trivially implemented with `.Index().ForEach(...)`, so I'm not sure how useful this method is. OTOH, it provides parity with `Select<TSrc,int,TRes>` overload...

atifaziz:
@fsateler is absolutely right. `Index` was added to MoreLINQ to solve that problem once and for all. It not only gets tiring to provide an index-supplying version of callbacks for each operator but it also explodes the number of overloads and unit tests to write for each operator. `Index` is ready to use on anything. However, `ForEach` is simple and I don't see any harm in accepting this given that it's not already overloaded with overloads, :laughing: don't really expect it to be and @tg73 has already taken the trouble to provide the implementation along with unit tests.

@tg73 I'll be re-basing this on top `v1` branch (unless you want to do that?) as we can release that right away and then merge it into `master` to make it part of v2.

tg73:
@atifaziz Sounds good to me, I'm happy for you to do the rebase.

atifaziz:
Rebased on `v1` as bfaf3a9df851c3325189d4aed8a7a3aaec20a812 and published in [version 1.4](https://www.nuget.org/packages/morelinq/1.4.0), including [portable version](https://www.nuget.org/packages/MoreLinq.Portable/1.4.0).

PR #126, opened 2015-11-26
Implement Exactly method

sholland1:


JamesFaix:
Have you considered an "Exactly-Where" overload like 
    ```Exactly<T>(this IEnumerable<T> collection, Func<T, Boolean> condition, int count)```

jnyrup:
@JamesFaix I considered such an overload myself, when I wrote a similar method. 
I looked at other extension methods and saw that very LINQ methods took more than one argument.
I came to the conclusion that `list.Where(condition).Exactly(count)` was at least as or even more readable than `list.Exactly(condition, count)`.

atifaziz:
This was implemented in #197.

PR #129, opened 2016-01-19
GroupAdjacent: Do not use Select to project key and element

fsateler:
Do it inside the loop. Also, avoid allocating a new KeyValuePair element, as it is not needed.

On a simple naive test like `longList.GroupAdjacent(i => i.Key).Consume();`, it achieves a 30% performance improvement in my system. Real-World effect will probably be lower (as selector are more complicated and are not consumed in a tight loop), but I still think it is a worthwhile improvement.

### DIFF
--- a/MoreLinq/GroupAdjacent.cs
+++ b/MoreLinq/GroupAdjacent.cs
@@ -181,25 +181,25 @@ namespace MoreLinq
             Debug.Assert(elementSelector != null);
             Debug.Assert(comparer != null);
 
-            using (var iterator = source.Select(item => new KeyValuePair<TKey, TElement>(keySelector(item), elementSelector(item)))
-                                        .GetEnumerator())
+            using (var iterator = source.GetEnumerator())
             {
                 var group = default(TKey);
                 var members = (List<TElement>) null;
 
                 while (iterator.MoveNext())
                 {
-                    var item = iterator.Current;
-                    if (members != null && comparer.Equals(group, item.Key))
+                    var key = keySelector(iterator.Current);
+                    var element = elementSelector(iterator.Current);
+                    if (members != null && comparer.Equals(group, key))
                     {
-                        members.Add(item.Value);
+                        members.Add(element);
                     }
                     else
                     {
                         if (members != null)
                             yield return CreateGroupAdjacentGrouping(group, members);
-                        group = item.Key;
-                        members = new List<TElement> { item.Value };
+                        group = key;
+                        members = new List<TElement> { element };
                     }
                 }
 


atifaziz:
The changes look fairly innocent so I don't see much in the way of merging this but I'm curious about your perf test code (e.g. how big was `longList`?) and configuration. Besides it being always good to share & document some metrics, I'm mostly curious because I'm surprised with the performance improvement being as much as 30%. The `KeyValuePair` allocation doesn't worry me as much as it's a value type and doesn't put any pressure on the GC heap. Also did you test with debug or release build?

fsateler:
The benchmark code is as follows (very ugly, as it evolved as I tested different hypothesis):


	int n = 40000;
	var origList = Enumerable.Range(0, n).Select(i => i);
	
	var l = origList.Interleave(origList, origList, origList, origList).Select(i => new { k = i, val = i*i }).ToList().Take(n*4);
	
	MyExtensions.Benchmark(() => {
		l.GroupAdjacent(i => i.k).Consume();
	});


Where the Benchmark function is:

	public static void Benchmark(Action block, int ntimes = 1000, string name = "block") {
		// Precompile method
		block();
		var sw = new Stopwatch();
        sw.Start();
        for (int i=0; i < ntimes; i++) {
            block();
        }
		
		sw.Stop();
		var perLap = (double)sw.ElapsedMilliseconds/(double)ntimes;
        String.Format($"Time Taken: {sw.ElapsedMilliseconds:n}ms total, {perLap:0.000}ms p/i: {name}", sw.ElapsedMilliseconds, name).Dump();
	}


As you may guess, this is running in linqpad, but with optimizations enabled. I have 2 tabs with the same code, one referencing `morelinq.2.0.0-alpha01` from nuget, the other one my  local build (in release mode).
I have varied the `n` to test different sizes. On my system, my last run with `n = 40000` resulted in 

    nuget: 16,180.00ms total, 16.180ms p/i
    local: 12,515.00ms total, 12.515ms p/i

Different runs add or take ~ 400ms. With `n = 10000` I get:

```
nuget: 3,799.00ms total, 3.799ms p/i
local:  3,145.00ms total, 3.145ms p/i
```

With low values of `n` the impact is still visible (`n = 10`):

```
nuget: 6.00ms total, 0.006ms p/i
local: 4.00ms total, 0.004ms p/i:
```

atifaziz:
Thanks for sharing the details of your performance tests. I'll run those on my end and share any findings that differ.

atifaziz:
So I'm also seeing an increase in performance as well (wasn't doubting that) but around ~20% on my box, which is still good enough. The code is quite tight and from the little digging I did with profiled & instrumented code, my guess is that a large part of the increase is coming from doing away with the interface method call to `IEnumerator.MoveNext`, from `Select`'s iterator, as well as the delegate invocation of the key-value pair projection. As you already noted, the difference is probably negligible in more real-world code.

fsateler:
@atifaziz thanks. Yes, for most code it would probably be negligible, but I've lately been using this over 100k+ length sequences, and final consumption is a simple sum or similar, so the tight loop is not quite a far-fetched scenario. I wouldn't have been looking at this method if the profiler hadn't told me a relatively large amount of time was spent in it ;)

atifaziz:
I can certainly imagine transformations with source data completely in memory (thus becoming CPU-bound) can surface such cases and I'm glad to hear that real-world code is driving the optimizations. :+1:

There are probably many such optimization that can be made in MoreLINQ. In general, my approach has been to reuse what's already there for achieving correctness quickly. Then with tests in place, once can always get comfortable about optimizing if there's a burning need (make it work, right and then fast).

Thanks for the PR.

PR #157, opened 2016-03-10
SortedMerge, ToDelimitedString, Shuffle, SelectTrimmedNotEmpty and UTs.

smzinovyev:
Added: IEnumerable<T> SortedMerge<T, TKey>(this IEnumerable<IEnumerable<T>> source, Func<T, TKey> keySelector, OrderByDirection direction, IComparer<TKey> comparer)

Added internal class ListExtensions with the following methods:
List<T> AddSorted<T>(this List<T> list, T item)
List<T> AddSorted<T>(this List<T> list, T item, IComparer<T> comparer)
List<T> AddSorted<T>(this List<T> list, T item, IComparer<T> comparer, OrderByDirection direction)
List<T> AddSorted<T>(this List<T> list, T item, OrderByDirection direction)
List<T> AddSorted<T, TKey>(this List<T> list, T item, Func<T, TKey> keySelector, IComparer<TKey> comparer, OrderByDirection direction)
int BinarySearchBy<T, TKey>(this List<T> list, T item, Func<T, TKey> keySelector, IComparer<TKey> comparer, OrderByDirection direction)

Added UTs for:
OrderedMerge
SortedMerge
AddSorted

Added overloaded methods for ToDelimitedString in order to initialize StringBuilder with right capacity.
Added Shuffle method.
Added SelectTrimmedNotEmpty method.

Added UTs for Shuffle and SelectTrimmedNotEmpty

Updated Readme.md with a sample of using MoreLinq in DNX 4.5.1. A console app built in dnx451 can use MoveLinq v1.4.0

Declared ListExtensions class as internal. Kept UTs for it. So, signed test projects with the key and added them to InternalVisibleTo assemblies. Not sure about a build process, so added duplicate snk file. But my preferences is to keep snk file on sln file level and use link to the snk file in each project.

PR #154, opened 2016-03-04
Added:

smzinovyev:
IEnumerable<T> SortedMerge<T, TKey>(this IEnumerable<IEnumerable<T>> source, Func<T, TKey> keySelector, OrderByDirection direction, IComparer<TKey> comparer) where TKey : IComparable<TKey>

List<T> AddSorted<T>(this List<T> list, T item)
List<T> AddSorted<T>(this List<T> list, T item, IComparer<T> comparer)
List<T> AddSorted<T>(this List<T> list, T item, IComparer<T> comparer, OrderByDirection direction)
List<T> AddSorted<T>(this List<T> list, T item, OrderByDirection direction)
List<T> AddSorted<T, TKey>(this List<T> list, T item, Func<T, TKey> keySelector, IComparer<TKey> comparer, OrderByDirection direction)
int BinarySearchBy<T, TKey>(this List<T> list, T item, Func<T, TKey> keySelector, IComparer<TKey> comparer, OrderByDirection direction)

Added UTs for:

OrderedMerge
SortedMerge
AddSorted

PR #158, opened 2016-03-10
Updated Readme.md with a sample of using MoreLinq in DNX 4.5.1. A console app built in dnx451 can use MoveLinq v1.4.0

smzinovyev:
ok. one by one...


atifaziz:
@smzinovyev I appreciate your good intentions to contribute to MoreLINQ, but considering PR #154 and #157 and now this, my guess is that you're having trouble understanding how commits, branches and PRs are used to make _atomic_ contributions that can be easily reviewed and merged. And while I could try and explain it, it would cost me a lot of time and [there's plenty of material out there that does a good job of explaining](https://guides.github.com/introduction/flow/). I would therefore suggest to study the material and then return here with new and separate PRs. Rolling back to then forward again is not the right approach and pollutes/confuses the history of the project unnecessarily with trials & errors. Create independent branches, make commits that belong to the purpose of those branches and make PRs against those branches.

smzinovyev:
I had a good intention, but yes, I am not an expert in git and it seems the process is more important than final source code. This library is good, but I need some extra methods in it. So, will keep them in my own separate utility lib.

--
Sergiy Zinovyev

Issue #125, opened 2015-11-04
Unique / Duplicate operators


JamesFaix:
There are several variations on the theme of duplicate record detection that I find myself needing quite often when validating Excel files.

The first method, UniqueKeys just returns all keys (from a given key selector delegate) that only occur once in a collection.

```
 public static IEnumerable<TKey> UniqueKeys<TSource, TKey>(this IEnumerable<TSource> me,
        Func<TSource, TKey> keySelector, IEqualityComparer<TKey> keyComparer) {
        //check for null arguments

        return me.GroupBy(keySelector, keyComparer)
            .Where(g => g.Count() == 1)
            .Select(g => g.Key);
    }
```

Likewise, a DuplicateKeys method would be the same, just with "Where(g => g.Count() == 1)" replaced with "Where(g => g.Count() > 1)".

Building on these, are UniqueElements and DuplicateElements methods,

```
 public static IEnumerable<TSource> DuplicateElements<TSource, TKey>(this IEnumerable<TSource> me,
        Func<TSource, TKey> keySelector, IEqualityComparer<TKey> keyComparer) {
         //check for null args
        return me.ExceptByAll(me.UniqueKeys(keySelector, keyComparer), keySelector, keyComparer);
    }
```

The DuplicateKeys method can be used, for example, to find a distinct list of ID numbers that are used more than once, while DuplicateElements can be used to return a full list of rows where duplicate ID numbers are used.

The "Elements" methods are dependent on ExceptAll or ExceptByAll being implemented.  (see #124 )

atifaziz:
This is now covered by [`CountBy`](http://morelinq.github.io/2.0/ref/api/html/Overload_MoreLinq_MoreEnumerable_CountBy.htm) since [2.0](https://github.com/morelinq/MoreLINQ/releases/tag/v2.0.0), e.g.:

```c#
var xs = new[] { 7, 4, 6, 5, 1, 2, 3, 8, 6, 4, 5, 1, 2 };

var duplicates = // 4, 6, 5, 1, 2
    from e in xs.CountBy(x => x)
    where e.Value > 1
    select e.Key;
    
var unique = // 7, 3, 8
    from e in xs.CountBy(x => x)
    where e.Value == 1
    select e.Key;
```

Issue #124, opened 2015-11-04
UnionAll, IntersectAll, ExceptAll


JamesFaix:
LINQ's built-in "Union", "Except" and "Intersect" are somewhat odd among LINQ operators, as most others deal simply with enumerable collections, while these three treat enumerables as proper mathematical sets.  Although this nicely mirrors SQL's Union, Intersect, and Except, LINQ does not provide an alternative like SQL's "All" keyword that will allow duplicate elements to be returned.

A working implementation can be found here: http://www.siepman.nl/blog/post/2013/08/19/Except-intersect-and-union-without-distinct.aspx

They could possibly also be implemented as overloads of the existing set operators, but with an optional boolean argument that can toggle "normal" distinct results vs "all" duplicate results.

Example method signature:
    public static IEnumerable<T> Union<T>(this IEnumerable<T> first, IEnumerable<T> second, bool all = false);  
This would return the results of first.Union(second) if (all == false) and all duplicates in the result if (all == true).  

These operators could also be combined with MoreLINQ's ExceptBy (and prospecitve IntersectBy, see #63 ).

I have found using an ExceptByAll type method essential for checking a DataColumn for duplicate values, and reporting back the row numbers of those duplicates.

fsateler:
There is Concat, which is the same as SQL UNION ALL.

ExceptAll would be "left join where right hand is null", and IntersectAll woul be the same as "inner join, discard right hand side". But neither of these options expresses the intent clearly.

JamesFaix:
I agree that a standard UnionAll is redundant with Concat.  As for ExceptAll and IntersectAll being expressible with the help of Join, this is true, but as you mentioned, it does not express the intent clearly and LINQ Joins are very wordy to start with.  

Several LINQ functions can also be reduced to others, like Last() to Skip(Math.Min(0, Count()-2)).Take(1), or Any(predicate) to Where(predicate).Count() > 1.  Clarity of intent is a great way to eliminate errors before they happen.

Additionally I think there is room for ExcepAllBy, IntersectAllBy, and UnionAllBy variations, and the expressions that these could be reduced to (if in fact they can without semantic differences) would be a lot more complicated.

thelemontree10:
Hi,

Has the newest library has this function yet ? I have read through all but see nowhere.

Thank you.

Issue #63, opened 2015-08-21
IntersectBy Operator


atifaziz:
Similar to `DistinctBy` but with the functionality of `Intersect `

so Something like this:

```c#
var emails = GetAllEmails();
var uniqueEmails = emails.DistinctBy(x => x.Email).ToList();
var duplicates = emails.IntersectBy(uniqueEmails, x => x.EmailAddress);
```

Thanks

---
Originally reported on Google Code with ID 63

Reported by `jalchr` on 2011-02-14 09:48:06

fsateler:
I'm not sure there is much advantage over the alternative:

    var emails = GetAllEmails();
    var uniqueEmails = emails.DistinctBy(x => x.Email).ToList();
    var duplicates = emails.Where(e => uniqueEmails.Contains(e.EmailAddress));

Is the increase in readability significant?

JamesFaix:
If you put Contains() inside the predicate of Where(), it will get called once per iteration, which should put execution time around O(n^2).  An IntersectBy operator can be optimized to run a lot faster against large collections.

fsateler:
Actually, that would be m×n, but that can be solved by using a hashset instead of a list.

JamesFaix:
You are correct on both counts.  

But more to the point, we're now looking at a combination of about 5 function calls that require the client to worry about implementation details for optimizing performance, which could all be replaced by one sleek extension method.  This is what LINQ is for.

JamesFaix:
```
public static IEnumerable<TSource> IntersectBy<TSource, TKey>(this IEnumerable<TSource> first,
            IEnumerable<TSource> second,
            Func<TSource, TKey> keySelector,
            IEqualityComparer<TKey> keyComparer = null) {

            if (first == null) throw new ArgumentNullException("first");
            if (first == null) throw new ArgumentNullException("second");
            if (first == null) throw new ArgumentNullException("keySelector");
            
            keyComparer = keyComparer ?? EqualityComparer<TKey>.Default;

            var keys = new HashSet<TKey>(second.Select(keySelector), keyComparer);
            foreach (var item in first) {
                var k = keySelector(item);
                if (keys.Contains(k)) {
                    yield return item;
                    keys.Remove(k);
                }
            }
        }
```

fsateler:
@JamesFaix that has the problem of iterating the source list twice 

```csharp
public static IEnumerable<TSource> IntersectByImpl<TSource, TKey>(
[...]

PR #163, opened 2016-04-21
Add ExceptBy overloads & IntersectBy

JamesFaix:
Added overloads to ExceptBy and added IntersectBy.  Modified several
tests in ExceptByTest to avoid generic type ambiguity caused by new
overloads.  Created IntersectByTest, which has test corresponding to
each test in ExceptByTest.

This addresses issue #63, and provides a foundation for addressing issues #124 and #125.

### DIFF
--- a/MoreLinq/ExceptBy.cs
+++ b/MoreLinq/ExceptBy.cs
@@ -24,6 +24,54 @@ namespace MoreLinq
     static partial class MoreEnumerable
     {
         /// <summary>
+        /// Returns a distinct set of elements from the first collection,
+        /// whose keys are not in the second collection.
+        /// </summary>
+        /// <typeparam name="T">The type of source, keys, and result elements.</typeparam>
+        /// <param name="first">The source collection.</param>
+        /// <param name="second">The key collection.</param>
+        /// <param name="keySelector">The key selector.</param>
+        /// <returns>
+        /// Distinct set of elements from first collection,
+        /// whose keys are not in second collection.
+        /// </returns>
+        public static IEnumerable<T> ExceptBy<T>(this IEnumerable<T> first,
+            IEnumerable<T> second,
+            Func<T, T> keySelector) {
+
+            if (first == null) throw new ArgumentNullException("first");
+            if (second == null) throw new ArgumentNullException("second");
+            if (keySelector == null) throw new ArgumentNullException("keySelector");
+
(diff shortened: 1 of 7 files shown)


MoreLinq/ExceptBy.cs
```diff
@@ -24,6 +24,54 @@ namespace MoreLinq
...
+        /// <param name="first">The source collection.</param>
+        /// <param name="second">The key collection.</param>
+        /// <param name="keySelector">The key selector.</param>
+        /// <returns>
+        /// Distinct set of elements from first collection,
+        /// whose keys are not in second collection.
+        /// </returns>
+        public static IEnumerable<T> ExceptBy<T>(this IEnumerable<T> first,
+            IEnumerable<T> second,
+            Func<T, T> keySelector) {
```
fsateler: Is this key selector really necessary? Doesn't look like it to me. Would removing this paramtere make the type inference problems go away?

JamesFaix: Funny you should say that, actually.  

I don't think this overload is particularly useful, its really just a special case of when ```TKey``` and ```TSource``` are the same. In this case ```keySelector``` is a projection from one T to another T, not the identity function.  

Originally I did not have this overload (or the corresponding one on IntersectBy). After adding the overload with this signature:

```IEnumerable<TSource> ExceptBy<TSource, TKey>(this IEnumerable<TSource> first, IEnumerable<TKey> second, Func<TSource, TKey> keySelector)```

several of the unit tests for ExceptBy that were using String as both TSource and TKey gave me ambiguous type errors.  I added the ```ExceptBy<T>``` overload (and explicit type parameters in several methods on ExceptByTest) to get rid of those errors.

I believe my original overload is very useful, but I would like to get rid of ```ExceptBy<T>``` if possible.  I've been using this new ```ExceptBy<TSource, TKey>``` overload in practice now for a while, and I've never had ambiguous type issues.  I will see if there is a way to alter the tests to remove the ambiguity without adding any additional overloads.

JamesFaix:
I think I've overoverloaded ExceptBy and IntersectBy.  Perhaps the overloads where `second` is of type `TKey` should be renamed to ExceptKeys and IntersectKeys.  

The overloads with one type parameter just replace the ambiguous overload issue with ambiguity as to whether `second` is supposed to be a set of values or keys.

Issue #80, opened 2015-08-21
Except and Intersect overloads that take delegates


atifaziz:
Originally reported on Google Code with ID 80

```
Useful when you have one sequence, and you want to get the except/intersect from another
sequence based on a specific projection (like DistinctBy)

Originally sourced from here: 
http://akshayluther.com/2009/08/14/improving-linq-except-and-intersect/

Additional overloads added where the two sequences have different types added - since
the Func doesn't really care if they're the same or not and therefore you can just
check equality inside.

Sample usage:
var toRemove = existingResults.Except(newResults, (x, y) => x.EntityId == y.EntityId
&& x.EntityType == y.EntityType);

Code: 
 public static IEnumerable<TSource> Except<TSource>(this IEnumerable<TSource> first,
IEnumerable<TSource> second, Func<TSource, TSource, bool> comparer)
        {
            return first.Where(x => second.Count(y => comparer(x, y)) == 0);
        }

        public static IEnumerable<TSource> Intersect<TSource>(this IEnumerable<TSource>
first, IEnumerable<TSource> second, Func<TSource, TSource, bool> comparer)
        {
            return first.Where(x => second.Count(y => comparer(x, y)) == 1);
        }

        public static IEnumerable<TFirstSource> Except<TFirstSource, TSecondSource>(this
IEnumerable<TFirstSource> first, IEnumerable<TSecondSource> second, Func<TFirstSource,
TSecondSource, bool> comparer)
        {
            return first.Where(x => second.Count(y => comparer(x, y)) == 0);
        }

        public static IEnumerable<TFirstSource> Intersect<TFirstSource, TSecondSource>(this
IEnumerable<TFirstSource> first, IEnumerable<TSecondSource> second, Func<TFirstSource,
TSecondSource, bool> comparer)
        {
            return first.Where(x => second.Count(y => comparer(x, y)) == 1);
        }
```

Reported by `agrath` on 2013-07-09 22:34:29

JamesFaix:
This could be improved with something like the HashSet implementation of ExceptBy, if HashSet didn't require an IEqualityComparer.  Is there anything like Hashset that takes an Func<T, T, bool> as a comparer parameter?

JamesFaix:
Here is a pretty simple implementation I came up with for these methods.  I've left out argument validation to keep code samples brief.  I also ended each method name in "Delegate" to distinguish them from the similar methods in PR #169, but I do not think these are great final names. 

```
[...]

PR #164, opened 2016-04-22
Issue63 IntersectBy and ExceptBy overloads v2

JamesFaix:
Created IntersectBy operator.
Created similar IntersectKeys and ExceptKeys operators, which take a collection of keys as their second argument.
Created unit tests for IntersectBy, IntersectKeys, ExceptKeys, analogous to existing tests for ExceptBy

### DIFF
--- a/MoreLinq/ExceptAll.cs
+++ b/MoreLinq/ExceptAll.cs
@@ -1,0 +1,199 @@
+﻿#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq {
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    static partial class MoreEnumerable {
+
+        /// <summary>
+        /// Returns the sequence of elements in the first sequence which aren't
+        /// in the second sequence.
+        /// </summary>
+        /// <remarks>
+        /// This is not a set operation like <see cref="System.Linq.Enumerable.Except{TSource}(
+        /// System.Collections.Generic.IEnumerable{TSource}, System.Collections.Generic.IEnumerable{TSource})"/>; 
+        /// if multiple elements in <paramref name="first"/> are equal, all such elements are returned.
+        /// This operator uses deferred execution and streams results from <paramref name="first"/>,
+        /// but the entire set of elements from <paramref name="second"/> is cached as soon as execution begins.
+        /// Duplicate values in <paramref name="second"/> are not relevant and are discarded when <paramref name="second"/> is cached.
+        /// </remarks>
+        /// <typeparam name="TSource">The type of the elements of the input sequences.</typeparam>
+        /// <param name="first">The sequence of potentially included elements.</param>
+        /// <param name="second">The set of elements which may prevent elements in <paramref name="first"/> from being returned.</param>
+        /// <param name="comparer">The element comparer. If <c>null</c>, uses the default equality comparer for <typeparamref name="TSource"/>.</param>
+        /// <returns>The sequence of elements from <paramref name="first"/> that are not in <paramref name="second"/>, 
+        /// including duplicate values from <paramref name="first"/>.</returns>
+        /// <exception cref="System.ArgumentNullException">Thrown if <paramref name="first"/> or <paramref name="second"/> is <c>null</c>. 
+        /// If <paramref name="comparer"/> is <c>null</c>, the default equality comparer for <typeparamref name="TSource"/> is used.</exception>
+        public static IEnumerable<TSource> ExceptAll<TSource>(this IEnumerable<TSource> first,
+            IEnumerable<TSource> second,
+            IEqualityComparer<TSource> comparer) {
+            if (first == null) throw new ArgumentNullException("first");
+            if (second == null) throw new ArgumentNullException("second");
+
+            return ExceptAllKeysImpl(first, second, x => x, comparer);
+        }
+        
+        /// <summary>
+        /// Returns the sequence of elements in the first sequence which aren't
+        /// in the second sequence, according to a given key selector.
+        /// </summary>
+        /// <remarks>
+        /// This is not a set operation like <see cref="MoreLinq.MoreEnumerable.ExceptBy{TSource, TKey}(
+        /// System.Collections.Generic.IEnumerable{TSource}, System.Collections.Generic.IEnumerable{TSource}, System.Func{TSource, TKey})"/>;  
+        /// if multiple elements in <paramref name="first"/> have equal keys, all such elements are returned.
+        /// This operator uses deferred execution and streams results from <paramref name="first"/>,
+        /// but the entire set of keys from <paramref name="second"/> is cached as soon as execution begins.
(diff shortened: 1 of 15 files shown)


JamesFaix:
see also: [v1](https://github.com/morelinq/MoreLINQ/pull/163)

JamesFaix:
The last several commits address issues #124 , #125, and #80

MoreLinq/ExceptAll.cs
```diff
@@ -0,0 +1,186 @@
...
+    using System.Linq;
+
+    static partial class MoreEnumerable {
+
+        /// <summary>
+        /// Returns the sequence of elements in the first sequence which aren't
+        /// in the second sequence.
+        /// </summary>
+        /// <remarks>
+        /// This is a sequence operation; if multiple elements in <paramref name="first"/> are equal,
```
fsateler: I would add a note here that this is unlike `Except` from LINQ (I was wondering why we wanted `ExceptAll`...

MoreLinq/ExceptAll.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// Returns the sequence of elements from the first collection,
+        /// whose keys are not in the second collection, according to a given key selector.
+        /// </summary>
+        /// <typeparam name="TSource">The type of source and result elements.</typeparam>
+        /// <typeparam name="TKey">The type of the key.</typeparam>
+        /// <param name="first">The sequence of potentially included elements.</param>
+        /// <param name="second">The sequence of keys which may prevent elements in
+        /// <paramref name="first"/> from being returned.</param>
+        /// <param name="keySelector">The mapping from source element to key.</param>
+        /// <param name="keyComparer">The key comparer. If <c>null</c>, uses the default TKey equality comparer.</param>
```
fsateler: Missing <typeparamref> for annotating TKey.

MoreLinq/ExceptAll.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// <returns>
+        /// Sequence of elements from first sequence,
+        /// whose keys are not in second sequence.
+        /// </returns>
+        /// <remarks>
+        /// This is a sequence operation; if multiple elements in <paramref name="first"/> have
+        /// equal keys, all such elements are returned.
+        /// This operator uses deferred execution and streams the results, although
+        /// a set of keys from <paramref name="second"/> is immediately selected and retained.
```
fsateler: This is not true. The `IEnumerable` generated by the compiler does not execute anything until the first `MoveNext` call (which is why you have to do the null checks in the wrapper functions instead of the Impl one).

MoreLinq/ExceptBy.cs
```diff
@@ -150,11 +154,10 @@ static partial class MoreEnumerable {
...
             foreach (var element in first) {
                 var key = keySelector(element);
-                if (keys.Contains(key)) {
-                    continue;
-                }
-                yield return element;
-                keys.Add(key);
+                if (!keys.Contains(key)) {
+                    yield return element;
+                    keys.Add(key);
```
fsateler: [Add returns a boolean indicating whether the element was already present](https://msdn.microsoft.com/en-us/library/bb353005(v=vs.110).aspx) So no need to invoke both Contains and Add

MoreLinq/IntersectAll.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <typeparam name="TKey">The type of the key.</typeparam>
+        /// <param name="first">The sequence of potentially included elements.</param>
+        /// <param name="second">The sequence of keys which may allow elements in
+        /// <paramref name="first"/> to be returned.</param>
+        /// <param name="keySelector">The mapping from source element to key.</param>
+        /// <param name="keyComparer">The key comparer. If <c>null</c>, uses the default TKey equality comparer.</param>
+        /// <returns>Sequence of elements from first sequence
+        /// whose keys are in second sequence, according to the given key selector.</returns>
+        /// <remarks>This is a sequence operation; if multiple elements in <paramref name="first"/> have
+        /// equal keys, all such elements are returned.
```
fsateler: What about `second` duplicates? I think it should be mentioned that duplicates there are discarded.

MoreLinq/IntersectBy.cs
```diff
@@ -146,10 +146,10 @@ partial class MoreEnumerable {
...
 
             var keys = new HashSet<TKey>(second, keyComparer);
-            foreach (var item in first) {
-                var k = keySelector(item);
+            foreach (var element in first) {
+                var k = keySelector(element);
                 if (keys.Contains(k)) {
-                    yield return item;
+                    yield return element;
                     keys.Remove(k);
```
fsateler: [HashSet.Remove also reports success](https://msdn.microsoft.com/en-us/library/bb342694(v=vs.110).aspx). I know, not in your patch, but if you are already modifying...

fsateler:
@JamesFaix I think it is better if you split the changes into separate pull requests, as it makes reviewing the changes easier.

BTW, there are a few member names with lowercase names, the convention is using CapitalizedFirst.

Also, please check the xmldoc comments. I left a few comments on a couple, but they apply to most.

JamesFaix:
Thanks for all the feedback.  Sorry about the pull request bloat, I'm pretty new to Git.

I want to make sure I understand all your comments correctly.  Are you suggesting:
0. Generally, review XML docs for inconsistencies and missing members.
1. Add a very explicit `remark` to all the "All" methods, so they don't get confused with the standard set operators.
2. Add explicit comments about `second` duplicates being discarded on "All" methods.
3. Simplify `ExceptByImpl`, since `HashSet.Add` returns a `bool` that can be used instead of `Contains`.

I am a little confused about this one:
```
/// This operator uses deferred execution and streams the results, although
/// a set of keys from <paramref name="second"/> is immediately selected and retained.
```

_This is not true. The IEnumerable generated by the compiler does not execute anything until the first MoveNext call (which is why you have to do the null checks in the wrapper functions instead of the Impl one)._

Are you saying that the `HashSet` storing `second` is also deferred?

How is that related to placing null checks?

fsateler:
0. Correct, and for using references where appropriate.
1. I would find it useful, yes (a naive user might wonder why it exists at all). I think a simple "Like `Except`, but preserves distinct elements in the first sequence" or something like that should make it clear enough.
2. Yes, right after you document that duplicate elements in first are preserved.
3. Yes.

> Are you saying that the `HashSet` storing `second` is also deferred?

Yes, it will be executed upon fetching the first element of the resulting sequence. That is:

```
var q = a.ExceptAll(b);
// b has not been iterated over yet
q.GetEnumerator().MoveNext(); // b is iterated fully here
```

> How is that related to placing null checks?

The null checks are placed in the wrapper functions because you want the exceptions to occur when you create the sequence, not when you iterate over it, which might be in a completely separate module. If they are placed in the Impl methods, then you would get the exception upon enumerating, not upon function call. You can test it on LINQPad:

```
IEnumerable<object> Foo(int a) {
    if (a == 0) throw new Exception("Exception!");
    yield return null;
}

var q = Foo(0); // no exception here
q.Dump(); // Exception here
```

JamesFaix:
Got it. Thanks.  I was putting the null checks on the wrapper methods to minimize the number of checks each call had to perform; I hadn't thought about the execution timing.

JamesFaix:
1. Reverted all but my first 3 commits, which all deal with Except and Intersect variations.  
2. Changed the implementation of ExceptBy as mentioned above. 
3. Reviewed and corrected docs for all the Except and Intersect methods.

fsateler:
@JamesFaix I suggest you do the following:

1. Create a new (local) branch, where you keep all your new changes (ie, those unrelated to this PR).
2. Then do `git rebase -i master` on this branch. On the list git will present to you, just delete the lines referring to commits that are unrelated to this PR (be sure to also delete all the revert commits). After closing then git will leave you at a branch with only those changes.
3. Check it all worked out ok
4. `git push --force`. This will update this PR to have only the important commits, so it will have a few, not 20 :)

Yes, git can be confusing at first ;). Obligatory XKCD reference: http://xkcd.com/1597/

JamesFaix:
@fsateler Thanks for the tip.  That comic summarizes the impression I've been gathering from YouTube and StackOverflow.

fsateler:
@JamesFaix Why did you close the PR? I hope it wasn't due to me being annoying...

fsateler:
heh, I just saw v3 at #169 . Sorry for the noise.

JamesFaix:
No worries.  You've been very helpful, I'm just not fluent in git shell yet.

PR #170, opened 2016-05-07
Add StartsWith and EndsWith

anjdreas:
Similar to `String.StartsWith()` and `.EndsWith()`, except on `IEnumerable<T>`. Can even take strings as arguments, as sequences of chars.

* Add MoreEnumerable.StartsWith()
* Add MoreEnumerable.EndsWith()
* Add 15 test cases for StartsWith
* Add 15 test cases for EndsWith

All tests run OK locally.
I tried to glance through and match the coding conventions, but let me know if you see anything that is off or test cases you would want to add.

### DIFF
--- a/MoreLinq.Test/EndsWithTest.cs
+++ b/MoreLinq.Test/EndsWithTest.cs
@@ -1,0 +1,105 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class EndsWithTest
+    {
+        [TestCase(null, null)]
+        [TestCase(null, new[] {1})]
+        [TestCase(new[] {1}, null)]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void EndsWithThrowsIfFirstOrSecondAreNull(IEnumerable<int> first, IEnumerable<int> second)
+        {
+            first.EndsWith(second);
+        }
+
+        [TestCase(new[] {1, 2, 3}, new[] {2, 3}, Result = true)]
+        [TestCase(new[] {1, 2, 3}, new[] {1, 2, 3}, Result = true)]
+        [TestCase(new[] {1, 2, 3}, new[] {0, 1, 2, 3}, Result = false)]
+        public bool EndsWithWithIntegers(IEnumerable<int> first, IEnumerable<int> second)
+        {
+            return first.EndsWith(second);
+        }
+
+        [TestCase(new[] {'1', '2', '3'}, new[] {'2', '3'}, Result = true)]
+        [TestCase(new[] {'1', '2', '3'}, new[] {'1', '2', '3'}, Result = true)]
+        [TestCase(new[] {'1', '2', '3'}, new[] {'0', '1', '2', '3'}, Result = false)]
+        public bool EndsWithWithChars(IEnumerable<char> first, IEnumerable<char> second)
+        {
+            return first.EndsWith(second);
+        }
+
+        [TestCase("123", "23", Result = true)]
+        [TestCase("123", "123", Result = true)]
+        [TestCase("123", "0123", Result = false)]
+        public bool EndsWithWithStrings(string first, string second)
+        {
+            // Conflict with String.EndsWith(), which has precedence in this case
+            return MoreEnumerable.EndsWith(first, second);
+        }
+
+        [Test]
+        public void EndsWithReturnsTrueIfBothEmpty()
+        {
+            Assert.True(new int[0].EndsWith(new int[0]));
+        }
+
+        [Test]
+        public void EndsWithReturnsFalseIfOnlyFirstIsEmpty()
+        {
+            Assert.False(new int[0].EndsWith(new[] {1,2,3}));
+        }
+
+        [TestCase("", "", Result = true)]
+        [TestCase("1", "", Result = true)]
+        public bool EndsWithReturnsTrueIfSecondIsEmpty(string first, string second)
+        {
+            // Conflict with String.EndsWith(), which has precedence in this case
+            return MoreEnumerable.EndsWith(first, second);
+        }
+
+        [Test]
+        public void EndsWithDisposesBothSequenceEnumerators()
+        {
+            using (var first = TestingSequence.Of(1,2,3))
+            using (var second = TestingSequence.Of(1))
+            {
+                first.EndsWith(second);
+            }
+        }
+
+        [Test]
+        [SuppressMessage("ReSharper", "RedundantArgumentDefaultValue")]
+        public void EndsWithUsesSpecifiedEqualityComparerOrDefault()
+        {
+            var first = new[] {1,2,3};
+            var second = new[] {4,5,6};
+
+            Assert.False(first.EndsWith(second));
+            Assert.False(first.EndsWith(second, null));
+            Assert.False(first.EndsWith(second, new EqualityComparerFunc<int>((f, s) => false)));
+            Assert.True(first.EndsWith(second, new EqualityComparerFunc<int>((f, s) => true)));
+        }
+    }
+}
--- a/MoreLinq.Test/StartsWithTest.cs
+++ b/MoreLinq.Test/StartsWithTest.cs
@@ -1,0 +1,105 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class StartsWithTest
+    {
+        [TestCase(null, null)]
+        [TestCase(null, new[] {1})]
+        [TestCase(new[] {1}, null)]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void StartsWithThrowsIfFirstOrSecondAreNull(IEnumerable<int> first, IEnumerable<int> second)
+        {
+            first.StartsWith(second);
+        }
+
+        [TestCase(new[] {1, 2, 3}, new[] {1, 2}, Result = true)]
+        [TestCase(new[] {1, 2, 3}, new[] {1, 2, 3}, Result = true)]
+        [TestCase(new[] {1, 2, 3}, new[] {1, 2, 3, 4}, Result = false)]
+        public bool StartsWithWithIntegers(IEnumerable<int> first, IEnumerable<int> second)
+        {
+            return first.StartsWith(second);
+        }
+
+        [TestCase(new[] {'1', '2', '3'}, new[] {'1', '2'}, Result = true)]
+        [TestCase(new[] {'1', '2', '3'}, new[] {'1', '2', '3'}, Result = true)]
+        [TestCase(new[] {'1', '2', '3'}, new[] {'1', '2', '3', '4'}, Result = false)]
+        public bool StartsWithWithChars(IEnumerable<char> first, IEnumerable<char> second)
+        {
+            return first.StartsWith(second);
+        }
+
+        [TestCase("123", "12", Result = true)]
+        [TestCase("123", "123", Result = true)]
+        [TestCase("123", "1234", Result = false)]
+        public bool StartsWithWithStrings(string first, string second)
+        {
+            // Conflict with String.StartsWith(), which has precedence in this case
+            return MoreEnumerable.StartsWith(first, second);
+        }
+
+        [Test]
+        public void StartsWithReturnsTrueIfBothEmpty()
+        {
+            Assert.True(new int[0].StartsWith(new int[0]));
+        }
+
+        [Test]
+        public void StartsWithReturnsFalseIfOnlyFirstIsEmpty()
+        {
+            Assert.False(new int[0].StartsWith(new[] {1,2,3}));
+        }
+
+        [TestCase("", "", Result = true)]
+        [TestCase("1", "", Result = true)]
+        public bool StartsWithReturnsTrueIfSecondIsEmpty(string first, string second)
+        {
+            // Conflict with String.StartsWith(), which has precedence in this case
+            return MoreEnumerable.StartsWith(first, second);
+        }
+
+        [Test]
+        public void StartsWithDisposesBothSequenceEnumerators()
+        {
+            using (var first = TestingSequence.Of(1,2,3))
+            using (var second = TestingSequence.Of(1))
+            {
+                first.StartsWith(second);
+            }
+        }
+
+        [Test]
+        [SuppressMessage("ReSharper", "RedundantArgumentDefaultValue")]
+        public void StartsWithUsesSpecifiedEqualityComparerOrDefault()
+        {
+            var first = new[] {1,2,3};
+            var second = new[] {4,5,6};
+
+            Assert.False(first.StartsWith(second));
+            Assert.False(first.StartsWith(second, null));
+            Assert.False(first.StartsWith(second, new EqualityComparerFunc<int>((f, s) => false)));
+            Assert.True(first.StartsWith(second, new EqualityComparerFunc<int>((f, s) => true)));
+        }
+    }
+}
--- a/MoreLinq/EndsWith.cs
+++ b/MoreLinq/EndsWith.cs
@@ -68,7 +68,8 @@ namespace MoreLinq
         {
             if (first == null) throw new ArgumentNullException("first");
             if (second == null) throw new ArgumentNullException("second");
-            if (comparer == null) comparer = EqualityComparer<T>.Default;
+
+            comparer = comparer ?? EqualityComparer<T>.Default;
 
             ICollection<T> secondCollection = second as ICollection<T> ?? second.ToList();
             using (var firstIter = first.TakeLast(secondCollection.Count).GetEnumerator())
--- a/MoreLinq/EndsWith.cs
+++ b/MoreLinq/EndsWith.cs
@@ -1,0 +1,74 @@
+#region License and Terms
+
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2016 Andreas Gullberg Larsen. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MoreLinq
+{
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        ///     Determines whether the end of <paramref name="first" /> matches <paramref name="second" />.
+        /// </summary>
+        /// <typeparam name="T">Type of elements.</typeparam>
+        /// <param name="first">The sequence to check.</param>
+        /// <param name="second">The sequence to compare to.</param>
+        /// <returns>
+        ///     <c>true</c> if <paramref name="first" /> ends with elements equivalent to <paramref name="second" />.
+        /// </returns>
+        /// <remarks>
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.EndsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> using <see cref="EqualityComparer{T}.Default" /> on pairs
+        ///     of elements at the same index.
+        /// </remarks>
+        public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
+        {
+            return EndsWith(first, second, EqualityComparer<T>.Default);
+        }
+
+        /// <summary>
+        ///     Determines whether the end of <paramref name="first" /> matches <paramref name="second" />.
+        /// </summary>
+        /// <typeparam name="T">Type of elements.</typeparam>
+        /// <param name="first">The sequence to check.</param>
+        /// <param name="second">The sequence to compare to.</param>
+        /// <param name="comparer">Equality comparer to use.</param>
+        /// <returns>
+        ///     <c>true</c> if <paramref name="first" /> ends with elements equivalent to <paramref name="second" />.
+        /// </returns>
+        /// <remarks>
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.EndsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> on pairs of elements at the same index.
+        /// </remarks>
+        public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second, IEqualityComparer<T> comparer)
+        {
+            if (first == null) throw new ArgumentNullException("first");
+            if (second == null) throw new ArgumentNullException("second");
+            if (comparer == null) comparer = EqualityComparer<T>.Default;
+
+            ICollection<T> secondCollection = second as ICollection<T> ?? second.ToList();
+            using (var firstIter = first.TakeLast(secondCollection.Count).GetEnumerator())
+            {
+                return secondCollection.All(item => firstIter.MoveNext() && comparer.Equals(firstIter.Current, item));
+            }
+        }
+    }
+}
--- a/MoreLinq/EndsWith.cs
+++ b/MoreLinq/EndsWith.cs
@@ -24,8 +24,8 @@ namespace MoreLinq
     static partial class MoreEnumerable
     {
         /// <summary>
-        /// Determines whether the end of <paramref name="first" /> matches
-        /// <paramref name="second" />.
+        /// Determines whether the end of the first sequence is equivalent to
+        /// the second sequence, using the default equality comparer.
         /// </summary>
         /// <typeparam name="T">Type of elements.</typeparam>
         /// <param name="first">The sequence to check.</param>
@@ -47,8 +47,8 @@ namespace MoreLinq
         }
 
         /// <summary>
-        /// Determines whether the end of <paramref name="first" /> matches
-        /// <paramref name="second" />.
+        /// Determines whether the end of the first sequence is equivalent to
+        /// the second sequence, using the specified element equality comparer.
         /// </summary>
         /// <typeparam name="T">Type of elements.</typeparam>
         /// <param name="first">The sequence to check.</param>
--- a/MoreLinq/EndsWith.cs
+++ b/MoreLinq/EndsWith.cs
@@ -1,5 +1,4 @@
 #region License and Terms
-
 // MoreLINQ - Extensions to LINQ to Objects
 // Copyright (c) 2016 Andreas Gullberg Larsen. All rights reserved.
 //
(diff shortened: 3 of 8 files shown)


MoreLinq.Test/EndsWithTest.cs
```diff
@@ -0,0 +1,79 @@
...
+        [Test]
+        public void EndsWithReturnsFalseIfOnlyFirstIsEmpty()
+        {
+            new int[0].EndsWith(new[] {1,2,3});
+        }
+
+        [Test]
+        public void EndsWithReturnsFalseIfOnlySecondIsEmpty()
+        {
+            new[] {1,2,3}.EndsWith(new int[0]);
```
fsateler: Usually using an empty "needle" returns true for these methods: `"asdf".EndsWith("") => true`

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <param name="first">The sequence to check.</param>
+        /// <param name="second">The sequence to compare to.</param>
+        /// <returns>
+        ///     <c>true</c> if <paramref name="first" /> begins with elements equivalent to <paramref name="second" />.
+        /// </returns>
+        /// <remarks>
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.StartsWith(string)" /> and
+        ///     it calls <see cref="object.Equals(object)" /> on pairs of elements at the same index.
+        /// </remarks>
+        public static bool StartsWith<T1, T2>(this IEnumerable<T1> first, IEnumerable<T2> second)
```
fsateler: I'm not sure it makes sense to have different types for both enumerables... why did you use separate types?

anjdreas: I was contemplating the very same thing, and originally started out with only one.
What I figured was that we limit the number of usecases by doing only 1 type, since object.Equals() can be called on any two types, and in some usecases that may make sense.

So I was going to write an example how it could be convenient to allow comparing a sequence of ints to a sequence of longs, but then I tested 1L.Equals(1) and realized it is in fact false. I was not aware of this. Reading up on Equals(), it's generally assumed that if they are not of the same type, it should return false. The `==` operator, however, does return true, but that would not apply in this implementation anyway.

Long story short, yes, you are absolutely right. It should be only one type. I'll update the PR.

MoreLinq.Test/EndsWithTest.cs
anjdreas: Thanks, I verified that String has the same behavior. Will update to true in this case.

anjdreas: I just realized the implementation was already like you wanted, but the tests were lacking the asserts :-)

anjdreas:
Fixed tests and went with single param type. I also found a couple of other things I fixed.

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            if (first == null) throw new ArgumentNullException("first");
+            if (second == null) throw new ArgumentNullException("second");
+
+            var sourceIter = first.GetEnumerator();
```
atifaziz: This enumerator will leak as it is never disposed. I suggest adding a test to prove that and then wrapping it in `using`. In fact, the test should assert that both enumerators are properly disposed and while `All` is probably doing that for the second one, it's still good to have a test for in case someone refactors the implementation to not use `All`.

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,49 @@
...
+            return second.All(item => sourceIter.MoveNext() && Equals(sourceIter.Current, item));
```
atifaziz: The equality test should use [`Equals`][Equals] of an [`IEqualityComparer<T>`][IEqualityComparer] here and the comparer should be received as an argument to the function. If the comparer passed is a null reference the [default equality comparer](https://msdn.microsoft.com/en-us/library/ms224763(v=vs.110).aspx) should be used. An overload should then exist that sends `null` by default as the comparer argument. Tests should then be added to assert the employment of the comparer.

[Equals]: https://msdn.microsoft.com/en-us/library/ms132154(v=vs.110).aspx
[IEqualityComparer]: https://msdn.microsoft.com/en-us/library/ms132151(v=vs.110).aspx

atifaziz:
I would add an optimization specifically for the case where the sequences are in real fact instances of strings.

MoreLinq/StartsWith.cs
anjdreas: Thanks, I was not aware iterators were supposed to be disposed.
I found some test code where you were using TestSequence that I reused, very good!

MoreLinq/StartsWith.cs
anjdreas: Thanks, this is absolutely better.

MoreLinq.Test/StartsWithTest.cs
```diff
@@ -61,19 +61,20 @@ public bool StartsWithWithStrings(string first, string second)
...
         }
 
-        [Test]
-        public void StartsWithReturnsFalseIfOnlySecondIsEmpty()
+        [TestCase("", "", Result = true)]
+        [TestCase("1", "", Result = true)]
+        public bool StartsWithReturnsTrueIfSecondIsEmpty(string first, string second)
         {
-            new[] {1,2,3}.StartsWith(new int[0]);
+            return first.StartsWith(second);
```
atifaziz: Hmmm, I bet this does not call your implementation of `StartsWith` but [`String.StartsWith`](https://msdn.microsoft.com/en-us/library/baketfxw(v=vs.110).aspx), so while the test may be passing, it's not testing the right thing!

MoreLinq.Test/EndsWithTest.cs
```diff
@@ -61,19 +61,20 @@ public bool EndsWithWithStrings(string first, string second)
...
         }
 
-        [Test]
-        public void EndsWithReturnsFalseIfOnlySecondIsEmpty()
+        [TestCase("", "", Result = true)]
+        [TestCase("1", "", Result = true)]
+        public bool EndsWithReturnsTrueIfSecondIsEmpty(string first, string second)
         {
-            new[] {1,2,3}.EndsWith(new int[0]);
+            return first.StartsWith(second);
```
atifaziz: I reckon you meant to call `EndsWith` here instead. [My earlier comment on a `StartsWith` test not testing the right implementation](https://github.com/morelinq/MoreLINQ/pull/170/commits/948d15a1488b24afe9d3d7c549a7be13afb1715d#r62559356) applies here too.

anjdreas:
I've addressed all your feedback now. The only thing I was not able to add test for was the string optimization case, do you know of a good way to accomplish that?

Thank you for the awesome PR review, I learnt a great deal and it was very helpful to improve the PR.

MoreLinq.Test/StartsWithTest.cs
anjdreas: Good catch, I did not notice. Fixing it now. It's getting kind of subtle with IEnumerable<char> and string overloads, but I can only assume the concrete type has precedence.

MoreLinq.Test/EndsWithTest.cs
anjdreas: Yes I noticed too, already fixed.

atifaziz:
> The only thing I was not able to add test for was the string optimization case, do you know of a good way to accomplish that?

Here's what I had in mind:

```c#
string str1, str2;
if ((str1 = first as string) != null && (str2 =  second as string) != null && comparer == null)
{
    return str1.StartsWith(str2);
}
```

Honestly, though, it feels like YAGNI as it would be doing an extremely rare service to a few corner cases. It can leave it off until someone they comes screaming for it because they can demonstrate that the omission is simply appalling. :)

anjdreas:
Oh, I actually did implement it (8d668f6cdca47d97586e1215bb3fc485b2c8fecd), but I was not able to write a **test** for it, so that is what I was wondering if you knew how to write :-)

At any rate, yes, I was not sure either how widely used this would be, but it's there and I don't think it adds much overhead to the non-string scenarios. Keep it or remove it, I'm fine with either way.

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,49 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2012 Atif Aziz. All rights reserved.
```
atifaziz: Feel free to Copyright this file (including related files) and (presumably original) work to you, including setting the first year of publication to presumably (2016). If you've taken inspiration from a publicly available implementation, then a common courtesy is to leave a comment crediting the source (assuming license compatibility).

atifaziz:
@anjdreas Ah, I somehow missed that. The implementation isn't correct though. It *shouldn't* drop the comparer silently. The test would be easy to add but I think it's safer & wiser right now to simply revert 45fc4a82ba0a6595506fbaf888590f4535521b92 it as a case of YAGNI. The additional requirement that the comparer would have be `null` to induce the optimization, the favoring scenarios are probably just a figment of imagination right now. Sorry for [thinking out loud and misleading there](https://github.com/morelinq/MoreLINQ/pull/170#issuecomment-217923953).

MoreLinq/EndsWith.cs
```diff
@@ -29,21 +29,24 @@ static partial class MoreEnumerable
...
         /// <returns>
         ///     <c>true</c> if <paramref name="first" /> ends with elements equivalent to <paramref name="second" />.
         /// </returns>
         /// <remarks>
         ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.EndsWith(string)" /> and
-        ///     it calls <see cref="object.Equals(object)" /> on pairs of elements at the same index.
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> on pairs of elements at the same index.
         /// </remarks>
-        public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
+        public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second, IEqualityComparer<T> comparer = null)
```
atifaziz: The `comparer` should be defaulted via an *overloaded* of `EndsWith`.

MoreLinq/StartsWith.cs
```diff
@@ -29,21 +29,24 @@ static partial class MoreEnumerable
...
         /// <returns>
         ///     <c>true</c> if <paramref name="first" /> begins with elements equivalent to <paramref name="second" />.
         /// </returns>
         /// <remarks>
         ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.StartsWith(string)" /> and
-        ///     it calls <see cref="object.Equals(object)" /> on pairs of elements at the same index.
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> on pairs of elements at the same index.
         /// </remarks>
-        public static bool StartsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
+        public static bool StartsWith<T>(this IEnumerable<T> first, IEnumerable<T> second, IEqualityComparer<T> comparer = null)
```
atifaziz: The `comparer` should be defaulted via an *overloaded* of `StartsWith`.

MoreLinq/EndsWith.cs
```diff
@@ -29,21 +29,24 @@ static partial class MoreEnumerable
...
         {
             if (first == null) throw new ArgumentNullException("first");
             if (second == null) throw new ArgumentNullException("second");
 
+            comparer = comparer ?? EqualityComparer<T>.Default;
+
             using (var firstIter = first.Reverse().GetEnumerator())
```
atifaziz: The implementation of `EndsWith` is redundant with `StartsWith` and adds no value because anyone can call `Reverse` on both input sequences and use `StartsWith`. `EndWith` should either be removed or implemented doing the *real work* more effectively.

MoreLinq/EndsWith.cs
anjdreas: Sure, no problem! 
Out of curiosity, is this a convention in this library or in .NET in general?
I personally prefer arguments with default null values, which reduces the number of overloads when the parameter list grows.

anjdreas:
Reverted the String optimization, replaced with overloads.

anjdreas:
I just noticed the Nuspecs folder, do you want a .nuspec for each method as well? Or one for both? Or no nuspec?

MoreLinq/EndsWith.cs
atifaziz: >  is this a convention in this library or in .NET in general?

It's the convention used by LINQ (have a look at [`SequenceEqual`](https://msdn.microsoft.com/en-us/library/system.linq.enumerable.sequenceequal(v=vs.110).aspx), which has overloads, [one where the default comparer will be used](https://msdn.microsoft.com/en-us/library/bb348567(v=vs.110).aspx) and [another where the caller can supply one](https://msdn.microsoft.com/en-us/library/bb342073(v=vs.110).aspx) and .NET in general. Above all, this is not about personal preferences.

An often forgotten idea about LINQ is being able to use expression trees to project a query into another system. An overload without any comparer permits a `first.StartsWith(second)` to be projected into another system whose defaults may be something else or may have a different representation for expressing comparison semantics.

anjdreas: That is a very good point, about LINQ and expression trees. Thanks for that.

MoreLinq/EndsWith.cs
anjdreas: You are probably right, but from a consumer stand point, I'd find it awkward if there was a StartsWith but no EndsWith method. It comes down to convenience and convention for me.

MoreLinq/EndsWith.cs
fsateler: See also [FxCop rule CA1026](https://msdn.microsoft.com/en-us/library/ms182135.aspx). Not all languages targeting the .NET framework allow default parameters.

MoreLinq/EndsWith.cs
fsateler: I think you should probably use [TakeLast](http://morelinq.github.io/2.0/ref/api/html/M_MoreLinq_MoreEnumerable_TakeLast__1.htm), which would also avoid the need to cache the entire sequence in memory.

atifaziz: @fsateler You would probably have to commit the second sequence to memory in that case and one would imagine that it's usually smaller. However, you end up paying that cost up front if, for example, the first sequence is smaller in size. In any event, it would be better than the current implementation.

@anjdreas Sure but then you have to go the distance. :) What should be avoided is duplicate code. If it's about mere convenience using a naïve implementation then such a one-liner convenience extension method that builds on simply on top of other methods should be added to the user code.

anjdreas: So this discussion seems relevant: https://github.com/dotnet/corefx/issues/4750

At first I thought, hey let's check if it's IList then use indexing to optimize, but I had not considered that IEnumerable can change during the enumeration. Apparently reverse is not trivial at this point.

anjdreas: How about the first approach suggested in this comment?
https://github.com/dotnet/corefx/issues/4750#issuecomment-172285275

Optimize TakeLast() for IList sequences, with the following:
```
int i = source.Count - 1;
while (i >= 0)
{
  yield return source[i];
  int count = source.Count;
  if (--i >= count)
    i = count - 1;
}
```

Alternatively, throw an exception if the source changes, like foreach does?

Then EndsWith() could follow @fsateler's suggestion of using TakeLast(), optimized for IList.

anjdreas: The snippet was for reverse, but you get the point.

atifaziz: > but I had not considered that IEnumerable can change during the enumeration.

@anjdreas This is not really a concern for `StartsWith`/`EndWith` as they fold/reduce the sequence into a Boolean result and has immediate execution semantics.

anjdreas:
Please see today's commits.

I have now optimized EndsWith for IList and LinkedList cases, as well as using TakeLast() instead of Reverse() for IEnumerable case to reduce storage of first.

Also, see my above question about nuspec.

anjdreas:
Any thoughts on the latest commits?

fsateler:
To be honest I'm not too keen on providing alternative implementations of LINQ methods.

MoreLinq/EndsWith.cs
```diff
@@ -55,16 +59,68 @@ public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
...
+                {
+                    if (!firstIter.MoveNext() || !comparer.Equals(firstIter.Current, secondIter.Current))
+                        return false;
+                }
+
+                return true;
+            }
+        }
+
+        private static bool ListImpl<T>(IEqualityComparer<T> comparer, IList<T> firstList, IList<T> secondList)
```
fsateler: This is the ListImpl of what operation? Namespacing is necessary as `MoreEnumerable` is still a single class.

MoreLinq/EndsWith.cs
```diff
@@ -55,16 +59,68 @@ public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
...
+        {
+            using (var firstIter = IListEx.Reverse(firstList).GetEnumerator())
+            using (var secondIter = IListEx.Reverse(secondList).GetEnumerator())
```
fsateler: I fail to see the use of `Reverse` here: we already know this is a list, so we might as well do a for loop using indexes.

MoreLinq/EndsWith.cs
```diff
@@ -55,16 +59,68 @@ public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
         ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.EndsWith(string)" /> and
         ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> on pairs of elements at the same index.
         /// </remarks>
+        [SuppressMessage("ReSharper", "InvokeAsExtensionMethod", Justification = "Avoid ambiguity with Reverse(IEnumerable<T>).")]
+        [SuppressMessage("ReSharper", "PossibleMultipleEnumeration", Justification = "Count() will use the concrete implementation of Length/Count, if available.")]
```
fsateler: This one seems to not be necessary anymore?

MoreLinq/EndsWith.cs
```diff
@@ -55,16 +59,68 @@ public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
...
+            {
+                return newSecondList.All(item => firstIter.MoveNext() && comparer.Equals(firstIter.Current, item));
+            }
+        }
+
+        private static bool LinkedListImpl<T>(IEqualityComparer<T> comparer, LinkedList<T> firstLinkedList,
+            LinkedList<T> secondLinkedList)
+        {
+            using (var firstIter = LinkedListEx.Reverse(firstLinkedList).GetEnumerator())
+            using (var secondIter = LinkedListEx.Reverse(secondLinkedList).GetEnumerator())
```
fsateler: Similar to `ListImpl` below, if we know this is a `LinkedList`, we might as well iterate backwards from the `Last` property.

MoreLinq/EndsWith.cs
```diff
@@ -55,16 +59,68 @@ public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
...
+            // In order to properly dispose enumerator for second and only enumerate once,
+            // we must make a full copy of second to a list.
+            var newSecondList = second.ToList();
+            using (var firstIter = first.TakeLast(newSecondList.Count).GetEnumerator())
+            {
+                return newSecondList.All(item => firstIter.MoveNext() && comparer.Equals(firstIter.Current, item));
+            }
+        }
+
+        private static bool LinkedListImpl<T>(IEqualityComparer<T> comparer, LinkedList<T> firstLinkedList,
```
fsateler: Also namespace this function (ie, `EndsWithLinkedListImpl`)

anjdreas:
@fsateler 
> To be honest I'm not too keen on providing alternative implementations of LINQ methods.

Are you referring to Reverse(IList) and Reverse(LinkedList) ? Do you want to rename them or do you not want to provide them as public extension methods?

fsateler:
It seems to me the `EndsWith` implementation has become a bit more complex than needed: I'd use the same implementation (the non-specialized one) for all cases, with a small change. Instead of invoking `ToList` directly on `second`, do as follows:

```csharp
var secondCollection = second as ICollection<T> ?? second.ToList();
using (var firstIter = first.TakeLast(newSecondList.Count).GetEnumerator()) {
    secondCollection.All(item => firstIter.MoveNext() && comparer.Equals(firstIter.Current, item));
}
```

Because we only cared about the `Count` property, we only need `ICollection<T>`.

This would have the benefit of obviating the need of the `Reverse` implementations.

fsateler:
Small correction. I said:

>  I'd use the same implementation (the non-specialized one) for all cases

It should have said:

>  I'd use the same implementation (the non-specialized one) for all cases unless a significant benefit is found

fsateler:
@anjdreas Making public `Reverse` implementations has the potential of breaking user code, as users might have relied on the different implementation (ie, they changed the list during iteration).

anjdreas:
@fsateler 
> It seems to me the EndsWith implementation has become a bit more complex than needed

I agree with this. Your implementation should be good enough for most cases. I do not see any typical use cases for EndsWith() where the second sequence is extremely long, that warrants optimization of IList and LinkedList cases. And it can always be added later a good usecase for it pops up.

> Making public Reverse implementations has the potential of breaking user code, as users might have relied on the different implementation (ie, they changed the list during iteration).

That is a valid point. I noticed they pointed out the same argument in the CoreFx discussion.

How about we rename the extension methods to `ReverseFast()`? We no longer need them for the `EndsWith()` implementation, but it seems to me a natural helper for MoreLINQ to provide to people dealing with large sequences. Or should they just be removed altogether?

fsateler:
> How about we rename the extension methods to `ReverseFast()`? We no longer need them for the `EndsWith()` implementation, but it seems to me a natural helper for MoreLINQ to provide to people dealing with large sequences. Or should they just be removed altogether?

I am not sure. However, if they are added, it would be best to have a separate PR to discuss its merits.

anjdreas:
@fsateler I agree.

anjdreas:
What a mess, but I think the PR is getting just about ready now. Tests are OK.
Squashed and rebased on master.

anjdreas:
Simplified EndsWith and removed Reverse() extension methods.

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>
+        ///     <c>true</c> if <paramref name="first" /> begins with elements equivalent to <paramref name="second" />.
+        /// </returns>
+        /// <remarks>
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.StartsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> using <see cref="EqualityComparer{T}.Default"/> on pairs of elements at the same index.
+        /// </remarks>
+        public static bool StartsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
+        {
+            return StartsWith(first, second, EqualityComparer<T>.Default);
```
atifaziz: This is redundant with the defaulting of the comparer in the other overload. This puts the decision to use `EqualityComparer<T>.Default` in more than one place. Just replace with `null`.

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        ///     Determines whether the beginning of <paramref name="first" /> matches <paramref name="second" />.
```
atifaziz: For sake of consistency with the rest of the project, could you remove the indentation in XML doc comments.

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        /// <typeparam name="T">Type of elements.</typeparam>
+        /// <param name="first">The sequence to check.</param>
+        /// <param name="second">The sequence to compare to.</param>
+        /// <returns>
+        ///     <c>true</c> if <paramref name="first" /> begins with elements equivalent to <paramref name="second" />.
+        /// </returns>
+        /// <remarks>
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.StartsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> using <see cref="EqualityComparer{T}.Default"/> on pairs of elements at the same index.
```
atifaziz: This may seem like nitpicking, but as in the rest of the project, XML doc comment lines should be wrapped at 80 columns whenever possible (using good judgement with long `<see>` elements; best not to break them but wrap early if they will exceed). If it helps, use the [Intend Guides](https://visualstudiogallery.msdn.microsoft.com/e792686d-542b-474a-8c55-630980e72c30) extension if you are using Visual Studio as your editor.

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </returns>
+        /// <remarks>
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.StartsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> on pairs of elements at the same index.
+        /// </remarks>
+        public static bool StartsWith<T>(this IEnumerable<T> first, IEnumerable<T> second, IEqualityComparer<T> comparer)
+        {
+            if (first == null) throw new ArgumentNullException("first");
+            if (second == null) throw new ArgumentNullException("second");
+            if (comparer == null) comparer = EqualityComparer<T>.Default;
```
atifaziz: I would separate this from the args check section with an extra line otherwise it can mislead one (e.g. during review) to think this is another arg check when it is a defaulting of an arg.

MoreLinq/EndsWith.cs
```diff
@@ -0,0 +1,74 @@
...
+        ///     <c>true</c> if <paramref name="first" /> ends with elements equivalent to <paramref name="second" />.
+        /// </returns>
+        /// <remarks>
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.EndsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> using <see cref="EqualityComparer{T}.Default" /> on pairs
+        ///     of elements at the same index.
+        /// </remarks>
+        public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
+        {
+            return EndsWith(first, second, EqualityComparer<T>.Default);
```
atifaziz: My earlier

MoreLinq/EndsWith.cs
```diff
@@ -0,0 +1,74 @@
...
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.EndsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> on pairs of elements at the same index.
+        /// </remarks>
+        public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second, IEqualityComparer<T> comparer)
+        {
+            if (first == null) throw new ArgumentNullException("first");
+            if (second == null) throw new ArgumentNullException("second");
+            if (comparer == null) comparer = EqualityComparer<T>.Default;
+
+            ICollection<T> secondCollection = second as ICollection<T> ?? second.ToList();
```
atifaziz: Explicit declaration of `secondCollection` variable is redundant here. Replace with `var`.

MoreLinq/EndsWith.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </returns>
+        /// <remarks>
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.EndsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> on pairs of elements at the same index.
+        /// </remarks>
+        public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second, IEqualityComparer<T> comparer)
+        {
+            if (first == null) throw new ArgumentNullException("first");
+            if (second == null) throw new ArgumentNullException("second");
+            if (comparer == null) comparer = EqualityComparer<T>.Default;
```
atifaziz: [Ditto](https://github.com/morelinq/MoreLINQ/pull/170/commits/7cf104484815157c82e3bbc52d7e1b90cb0d3643#r63272983)

MoreLinq/EndsWith.cs
```diff
@@ -0,0 +1,74 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
```
atifaziz: Imports should go inside the namespace declaration below.

MoreLinq/StartsWith.cs
```diff
@@ -0,0 +1,70 @@
...
+        ///     This is the <see cref="IEnumerable{T}" /> equivalent of <see cref="string.StartsWith(string)" /> and
+        ///     it calls <see cref="IEqualityComparer{T}.Equals(T,T)" /> using <see cref="EqualityComparer{T}.Default"/> on pairs of elements at the same index.
+        /// </remarks>
+        public static bool StartsWith<T>(this IEnumerable<T> first, IEnumerable<T> second)
+        {
+            return StartsWith(first, second, EqualityComparer<T>.Default);
+        }
+
+        /// <summary>
+        ///     Determines whether the beginning of <paramref name="first" /> matches <paramref name="second" />.
```
atifaziz: Method summaries should be descriptive and avoid referring to any parameters because the rest of the documentation details may not be available for context, e.g., in a summary page like [table of content of methods](https://morelinq.github.io/2.0/ref/api/html/T_MoreLinq_MoreEnumerable.htm). Some like the following would be better but feel free to re-word.

> Determines whether the starting elements of the first sequence match those of a second sequence.

Also, the two overloads have *identical* summaries. Please [see docs of other MoreLINQ methods](https://morelinq.github.io/2.0/ref/api/html/T_MoreLinq_MoreEnumerable.htm) and [LINQ](https://msdn.microsoft.com/en-us/library/system.linq.enumerable(v=vs.100).aspx) itself for inspiration.

Ditto for `EndsWith`.

anjdreas:
Modified per your comments. I should maybe not have rebased to make it easier to see the diff, but I only added new commits after 'Add EndsWith'.

atifaziz:
@anjdreas Appreciate you working through all the details.

anjdreas:
@atifaziz Awesome, I very much appreciated the time you all put into the PR review. I learnt new things and your suggestions greatly improved the implementation. Thanks!

PR #173, opened 2016-05-11
Add a new default overload

Montago:
Slice(int) automatically cuts until the end of the array.

### DIFF
--- a/MoreLinq/Slice.cs
+++ b/MoreLinq/Slice.cs
@@ -51,6 +51,11 @@ namespace MoreLinq
                 ? SliceImpl(asList, startIndex, count)
                 : SliceImpl(sequence, startIndex, count);
         }
+        
+        public static IEnumerable<T> Slice<T>(this IEnumerable<T> sequence, int startIndex)
+        {
+            return sequence.Slice(startIndex, sequence.Count - startIndex);
+        }
 
         private static IEnumerable<T> SliceImpl<T>(IEnumerable<T> sequence, int startIndex, int count)
         {


MoreLinq/Slice.cs
```diff
@@ -51,6 +51,11 @@ public static IEnumerable<T> Slice<T>(this IEnumerable<T> sequence, int startInd
                 ? SliceImpl(asList, startIndex, count)
                 : SliceImpl(sequence, startIndex, count);
         }
+        
+        public static IEnumerable<T> Slice<T>(this IEnumerable<T> sequence, int startIndex)
```
fsateler: Missing xmldoc

MoreLinq/Slice.cs
```diff
@@ -51,6 +51,11 @@ public static IEnumerable<T> Slice<T>(this IEnumerable<T> sequence, int startInd
...
+        {
+            return sequence.Slice(startIndex, sequence.Count - startIndex);
```
fsateler: Missing argument null checks, and `IEnumerable` does not have a `Count` property. Moreover, just invoking `Count()` would iterate the sequence twice.

atifaziz:
This overload doesn't make sense as it's redundant with `Skip` that's already there in LINQ. `Skip` is optimised for lists. `Slice(index, count)` solely exists as an optimisation of `Skip` + `Take` for lists.

Montago:
In that case there's no need for Slice either, since it's just a wrapper for Skip + Take

PR #175, opened 2016-05-19
Ability to create batches based on a first and last match predicate

karthik25:
Hello,

To begin with MoreLINQ is an awesome idea and is really helpful! Regarding this pull request, I had to create a method that would batch a list of items based on a "first" match and "last" match, where other items in between does not satisfy either of the matches. After I wrote this method, I thought it would be a good addition to the MoreLINQ repository! 

Here is an example of how this could be put in to use:

```csharp
public class Activity
{
    public int Id { get; set; }
    public string Name { get; set; }
    public bool IsEnabled { get; set; }
}

public class BatchTester
{
    public static void Main(string[] args)
    {
        // Generate 9 entries w/ Name Activity 1, Activity 2, Activity 3,... and so on
        var activities = Enumerable.Range(0, 9)
                                   .Select(
                                        i => new Activity
                                              {
                                                  Id = i, 
                                                  Name = "Activity " + ((i%3) + 1), 
                                                  IsEnabled = true
                                              });
        Func<Activity, bool> firstMatch = activity => activity.Name.EndsWith("1");
        Func<Activity, bool> lastMatch = activity => activity.Name.EndsWith("3");
        var batches = activities.Batch(firstMatch, lastMatch);
        foreach (var batch in batches)
        {
            var combined = string.Join(", ", batch.Select(b => b.Name));
            // This would print Activtiy 1, Activity 2, Activity 3
            Console.WriteLine(combined);
        }
     }
}
```

Do check this out (test & implementation) and let me know your thoughts! Thanks for the opportunity!!

### DIFF
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -103,5 +103,61 @@ namespace MoreLinq.Test
         {
             new BreakingSequence<object>().Batch(1);
         }
+
+        [Test]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void BatchNullSequenceBasedOnPredicates()
+        {
+            Func<int, bool> firstMatch = i => i == 1;
+            Func<int, bool> lastMatch = i => i == 3;
+            MoreEnumerable.Batch(null, firstMatch, lastMatch).Consume();
+        }
+
+        [Test]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void BatchSequenceBasedOnNullFirstMatchPredicate()
+        {
+            var source = new[] { 1, 2, 3, 1, 2, 3, 1, 2, 3 };
+            Func<int, bool> lastMatch = i => i == 3;
+            source.Batch(null, lastMatch).Consume();
+        }
+
+        [Test]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void BatchSequenceBasedOnNullLastMatchPredicate()
+        {
+            var source = new[] { 1, 2, 3, 1, 2, 3, 1, 2, 3 };
+            Func<int, bool> firstMatch = i => i == 1;
+            source.Batch(firstMatch, null).Consume();
+        }
+
+        [Test]
+        public void BatchSequenceBasedOnPredicateSimple()
+        {
+            Func<int, bool> firstMatch = i => i == 1;
+            Func<int, bool> lastMatch = i => i == 3;
+            var result = new[] { 1, 2, 3, 1, 2, 3, 1, 2, 3 }.Batch(firstMatch, lastMatch);
+            using (var reader = result.Read())
+            {
+                reader.Read().AssertSequenceEqual(1, 2, 3);
+                reader.Read().AssertSequenceEqual(1, 2, 3);
+                reader.Read().AssertSequenceEqual(1, 2, 3);
+                reader.ReadEnd();
+            }
+        }
+
+        [Test]
+        public void BatchSequenceBasedOnPredicateSkipUnmatched()
+        {
+            Func<int, bool> firstMatch = i => i == 1;
+            Func<int, bool> lastMatch = i => i == 3;
+            var result = new[] { 1, 2, 1, 2, 3, 1, 2, 3 }.Batch(firstMatch, lastMatch);
+            using (var reader = result.Read())
+            {
+                reader.Read().AssertSequenceEqual(1, 2, 3);
+                reader.Read().AssertSequenceEqual(1, 2, 3);
+                reader.ReadEnd();
+            }
+        }
     }
 }
--- a/MoreLinq/Batch.cs
+++ b/MoreLinq/Batch.cs
@@ -64,6 +64,21 @@ namespace MoreLinq
             return BatchImpl(source, size, resultSelector);
         }
 
+        /// <summary>
+        /// Batches the sequence based on the first and last match predicates.
+        /// </summary>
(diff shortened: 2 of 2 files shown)


fsateler:
This seems to have weird semantics when the firstMatcher matches twice before the lastMatcher matches: Change `firstMatch` to `activity.Id < 3` , and the result becomes:

> Activity 3, Activity 1, Activity 2, Activity 3
> Activity 1, Activity 2, Activity 3

karthik25:
Hello! 

The intention of this extension method is to be used on a collection in which there is going to be a definite pattern, like in the example I used - every group would have 3 activities happening, say every day. May be adding a date field in the sample `Activity` class would help to implicitly convey the use case? Once it is used for something like the "id" which is almost always going to be unique, I guess the purpose is kind of lost! Correct me if I am wrong!!

I am trying to think of a use case for "id" fields, which would be valid, but I cannot think of any right now...

fsateler:
@karthik25 If that is the use case, what is the purpose of the `lastMatch` parameter? One could define a single parameter that defines batch boundaries (ie, it would return the previous batch and start a new one).

karthik25:
@fsateler  Yes, we can kind of achieve the same with just one predicate. But that would give me a lot of false positives. For example consider the following sequence:

Activity 1, Activity 1, Activity 1, Activity 2, Activity 1, Activity 2, Activity 3

I will get the following batches with this (the first 3 does not represent a pattern).

```
Activity 1
Activity 1
Activity 1, Activity 2
Activity 1, Activity 2, Activity 3
```

With a last match too, I would only get 

`Activity 1, Activity 2, Activity 3`

Which is a valid pattern. 

In short, without the last match predicate I will be getting batches which does not follow a pattern completely!

karthik25:
Thinking more about it, I could make the last match optional and start yielding batches even if it doesn't fit a pattern!

fsateler:
@karthik25 This is why it is important to understand the inuition behind each method, so that one can infer reasonable results.

BTW, MoreLinq [already has a method that splits on a single match function](http://morelinq.github.io/2.0/ref/api/html/M_MoreLinq_MoreEnumerable_Segment__1.htm).

BTW2, I find interesting the ability to select patterns, but I'm unsure if a pattern based on start/end is generic enough.

karthik25:
@fsateler Fair enough, I felt like it will be a great value add for MoreLINQ! With respect to making it more clear, like what this method does, can you pls. provide some suggestions? 

Thanks!

fsateler:
@karthik25 I'm still a bit confused about what it is meant to do, that is why I ask all these questions! For example, with the following sequence:

> Activity 1, Activity 1, Activity 1, Activity 2, Activity 1, Activity 2, Activity 3

Why does it make sense that the result be:

> Activity 1, Activity 2, Activity 3

Instead of the entire sequence? In other words, why was the start point succesively moved each time "Activity 1" appeared? And I mean why in an intuitive sense: why would a naive user expect such a result?

karthik25:
@fsateler :) 

This method is only concerned about "patterns" - that's why only `Activity 1, Activity 2, Activity 3` makes sense. Without a start and end I cannot think of a way to identify patterns.

For example, the following qualify as patterns. Line 2 is missing one of the activities but still follows a "pattern".

```
Activity 1, Activity 2, Activity 3
Activity 1, Activity 3
```

So, in a large set of activities, using this method we will be able find all the groups that did or did not follow a pattern and do further analysis to see what was missed quite often.

atifaziz:
@karthik25 So without looking at the implementation, I'm trying to understand your idea about _capturing patterns as batches_. So you are proposing that we have predicates to identify the start and end of _a pattern_ within a sequence. What do you expect to happen to things falling in-between _those patterns_? Should they simply be dropped?

karthik25:
@atifaziz Yes. That's the intention. Elements after first match and before last match are added as is. Anything else is dropped. 

For example in,

`Activity 1, Activity 1, Activity 1, Activity 2, Activity 3, Activity 1, Activity 3``

First 2 elements are dropped and only 2 batches are yielded `Activity 1, Activity 2, Activity 3` and `Activity 1, Activity 3`.

Based on the discussion we have had so far, I am more and more convinced that this is pretty specialized and does not seem generic and so probably does not belong here! I leave it to you!!

Thanks for the opportunity, learned quite a few new stuff looking at various methods here! Hopefully my next contribution is straight forward!!

atifaziz:
@karthik25 It seems like the concept of what you are proposing is misleading when articulated as _batches belonging to a pattern_. Perhaps it's also the vocabulary so based on what you've explained so far, I would say that what you're after is identifying adjacent sequences that are _bracketed_ and _non-bracketed_. I say _bracketed_ because your so-called pattern is defined by the first and last element. I wouldn't care about dropping the rest to keep the idea simple and general. If someone wants to drop the non-bracketed sequences, they could do that via a downstream `Where`. So given the following sequence of characters:

    abcdef[ghijk]lmnop[qrstuv]wx[y]z

and bracketing defined by predicates `ch == '['` for _first_ and `ch == ']'` for _last_, the output would be sequences of character sequences grouped as follows:

    abcdef, [ghijk], lmnop, [qrstuv], wx, [y], z

If someone were to put a gun to my head and ask me to cook up a name for this in under 10 seconds, it would be dull suggestions like `SegmentByBracketing` or `GroupAdjacentByBracketing`. 😛 

The point is that it's not plain and simple batching and your proposal as it stands now would overload the concept and lead to confusion. I don't know if what I've presented captures the abstract operation more precisely, but if it does, I would try and model it that way and see how it goes. If you find then it has wide and general application, you're welcome to return to discuss further and submit another PR.

Issue #72, opened 2015-08-21
Patch for /trunk/MoreLinq/DistinctBy.cs


atifaziz:
more concise although probably not recommended in pure functional-style

---
Originally reported on Google Code with ID 72

Reported by `tagir.magomedov` on 2012-04-26 19:03:29

---
 * *Attachment: [DistinctBy.cs.patch](https://storage.googleapis.com/google-code-attachments/morelinq/issue-72/comment-0/DistinctBy.cs.patch)*

fsateler:
This patch is wrong:

	var src = new [] { 1,2,2,4 };
	var distinct = src.DistinctBy(i => i);
	distinct.Dump();  // prints 1,2,4
	distinct.Dump(); // prints empty list

This is because `knownKeys` is created not on every iteration but only once when `Distinct` is invoked. This issue should be closed.

atifaziz:
@fsateler Thanks for reviewing and spotting that.

PR #183, opened 2016-07-06
Patch for /trunk/MoreLinq/DistinctBy.cs #72

VladimirRybalko:


atifaziz:
While your implementation is logically equivalent, it's less helpful than the current one in the sense that it causes an extra allocation of a closure for lambda passed to the `Where`.

MoreLinq/DistinctBy.cs
```diff
@@ -15,16 +15,11 @@
 // limitations under the License.
 #endregion
 
-
-#if NO_HASHSET
```
atifaziz: Conditional compilation on `NO_HASHSET` was there when targeting Silverlight but that's not the case so if you want to make a separate PR for removing this as well as the [alternate implementation for Silverlight](https://github.com/morelinq/MoreLINQ/blob/09aa5adcd4a10f693528f5044dd252f893fe962d/MoreLinq/DistinctBy.cs#L90-L100) then I'd be happy to merge that.

Issue #187, opened 2016-07-26
Top/TopBy sorts inverted?


fsateler:
Given the following sequence:

```csharp
new [] {1,2,3,4,5}.Top(3)
```

I expected the result to be `{ 5, 4, 3 }`, yet the result is `{ 1, 2, 3 }`.

atifaziz:
The result is correct [per documentation](https://morelinq.github.io/2.0/ref/api/html/M_MoreLinq_MoreEnumerable_Top__1.htm):

> Combines `OrderBy<TSource, TKey>(IEnumerable<TSource>, Func<TSource, TKey>)`, where each element is its key, and `Take<TSource>(IEnumerable<TSource> , Int32)` in a single operation.

Returns:

> A sequence containing at most top count elements from source, in their ascending order.

Note sure why you expect the result to be `{5, 4, 3}`.

jskeet:
Well, I think it's a reasonable expectation, in that new[] { 1, 2, 3, 4, 5
}.OrderBy(x => x).Take(3) would give { 1, 2, 3 } and new[] { 1, 2, 3, 4, 5
}.OrderByDescending(x => x).Take(3) would give { 5, 4, 3 }.

It seems odd to put them in *ascending* order despite returning the
*highest* items.

Jon

atifaziz:
> …in that `new[] { 1, 2, 3, 4, 5 }.OrderBy(x => x).Take(3)` would give `{ 1, 2, 3 }` …

And therefore `Top` returns the same. May be I haven't had my ☕ yet but what am I missing to see?

jskeet:
My mistake. I misread the original post as saying it was giving the right
results in the wrong order.

I think the problem here is that "top" implies "greatest" - suggesting
OrderByDescending rather than OrderBy.
Tricky.

Jon

fsateler:
On 27 Jul 2016 2:25 a.m., "Jon Skeet" <notifications@github.com> wrote:
>
> I think the problem here is that "top" implies "greatest" - suggesting
> OrderByDescending rather than OrderBy.
> Tricky.

Exactly.  Sorry for being unclear.

Top (3) returning the bottom 3 numbers is confusing. Maybe the confusion
can be resolved by adding an OrderByDirectiom parameter?

atifaziz:
[...]

PR #188, opened 2016-07-29
Rename Top to PartialSort, add OrderByDirection overloads

fsateler:
Rename Top/TopBy to PartialSort/PartialSortBy

Top somewhat implies expectations of sorting in descending order
(ie, the result would be the n largest elements).
PartialSort is a more neutral name, so use that.

Fixes: #187

PartialSort: Add overload taking an OrderByDirection parameter


atifaziz:
I think you missed the rename of the [actual iterator method](https://github.com/morelinq/MoreLINQ/pull/188/commits/7ce9047d0948f7bb865d2292fec7f4bdcea4ece9#diff-0e4966bd1000cde6b315f043f2b16c7cR118).

MoreLinq/PartialSort.cs
```diff
@@ -44,6 +44,27 @@ public static IEnumerable<T> PartialSort<T>(this IEnumerable<T> source, int coun
         }
 
         /// <summary>
+        /// Combines <see cref="OrderBy{T, TKey}(IEnumerable{T}, Func{T, TKey}, IComparer{TKey}, OrderByDirection)"/>,
```
atifaziz: The summary should mention a word about `direction` (without quoting the parameter). For example: “An additional parameter specifies the direction of the sort”. Ditto for other added overloads.

MoreLinq/PartialSort.cs
```diff
@@ -67,6 +88,32 @@ public static IEnumerable<T> PartialSort<T>(this IEnumerable<T> source, int coun
...
+        /// where each element is its key, and <see cref="Enumerable.Take{TSource}"/>
+        /// in a single operation.
+        /// </summary>
+        /// <typeparam name="T">Type of elements in the sequence.</typeparam>
+        /// <param name="source">The source sequence.</param>
+        /// <param name="count">Number of (maximum) elements to return.</param>
+        /// <param name="comparer">A <see cref="IComparer{T}"/> to compare elements.</param>
+        /// <param name="direction">The direction on which to sort the elements</param>
```
atifaziz: > The direction on which to sort the elements

Should instead read, “The direction ~~on~~_in_ which to sort the elements.”

atifaziz:
[README is missing the rename](https://github.com/morelinq/MoreLINQ/blame/a30936b18fa167b94313c61d469cbf0936567bc2/README.md#L77-L78).

fsateler:
Good catch on the README. It is also missing in the nuspec file.

I have force pushed a new version with all your comments addressed.

atifaziz:
> It is also missing in the nuspec file.

I see you found a catch too. 😉 Good eagle eyes. 😄 

Will review later and revert.

PR #190, opened 2016-08-01
Support for AtMost method

jnyrup:
As `AtLeast` is already implemented and there is another PR for `Exactly` I thought that `AtMost` was missing as I find it much more readable than `!AtLeast(count + 1)`.

I'm sorry for the many commits, but I struggled to cope with LF vs CRLF from my Linux machine and tried to fix it from within the GitHub interface. 
Please let me know if is an unacceptable mess.

### DIFF
--- a/MoreLinq.Test/AtMostTest.cs
+++ b/MoreLinq.Test/AtMostTest.cs
@@ -1,0 +1,176 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2016 Jonas Nyrup. All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using LinqEnumerable = System.Linq.Enumerable;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class AtMostTest
+    {
+        [Test]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void AtMostWithNullSequence()
+        {
+            IEnumerable<int> sequence = null;
+            sequence.AtMost(1);
+        }
+
+        [Test]
+        [ExpectedException(typeof(ArgumentOutOfRangeException))]
+        public void AtMostWithNegativeCount()
+        {
+            new[] { 1 }.AtMost(-1);
+        }
+
+        private static IEnumerable<int> GetSequence()
+        {
+            return new InfiniteSequence<int>(0);
+        }
+        [Test]
+        public void AtMostWithEmptySequenceHasAtMostZeroElements()
+        {
+            Assert.IsTrue(GetEmptySequence().AtMost(0));
+        }
+        [Test]
+        public void AtMostWithEmptySequenceHasAtMostOneElement()
+        {
+            Assert.IsTrue(GetEmptySequence().AtMost(1));
+        }
+        private static IEnumerable<int> GetEmptySequence()
+        {
+            return LinqEnumerable.Empty<int>();
+        }
+
+        [Test]
+        public void AtMostWithSingleElementHasAtMostZeroElements()
+        {
+            Assert.IsFalse(GetSingleElementSequence().AtMost(0));
+        }
+        [Test]
+        public void AtMostWithSingleElementHasAtMostOneElement()
+        {
+            Assert.IsTrue(GetSingleElementSequence().AtMost(1));
+        }
+        [Test]
+        public void AtMostWithSingleElementHasAtMostTwoElements()
+        {
+            Assert.IsTrue(GetSingleElementSequence().AtMost(2));
+        }
+        private static IEnumerable<int> GetSingleElementSequence()
+        {
+            return GetSequence().Take(1);
+        }
+
+        [Test]
+        public void AtMostWithManyElementsHasAtMostZeroElements()
+        {
+            Assert.IsFalse(GetManyElementSequence().AtMost(0));
+        }
+        [Test]
+        public void AtMostWithManyElementsHasAtMostOneElement()
+        {
+            Assert.IsFalse(GetManyElementSequence().AtMost(1));
+        }
+        [Test]
+        public void AtMostWithManyElementsHasAtMostThreeElements()
+        {
+            Assert.IsTrue(GetManyElementSequence().AtMost(3));
+        }
+        [Test]
+        public void AtMostWithManyElementsHasAtMostManyElements()
+        {
+            Assert.IsTrue(GetManyElementSequence().AtMost(4));
+        }
+        private static IEnumerable<int> GetManyElementSequence()
+        {
+            return GetSequence().Take(3);
+        }
+
+        //ICollection<T> Optimization Tests
+        [Test]
+        public void AtMostWithEmptySequenceHasAtMostZeroElementsForCollections()
+        {
+            Assert.IsTrue(GetEmptyArray().AtMost(0));
+        }
+        [Test]
+        public void AtMostWithEmptySequenceHasAtMostOneElementForCollections()
+        {
+            Assert.IsTrue(GetEmptyArray().AtMost(1));
+        }
+        private static IEnumerable<int> GetEmptyArray()
+        {
+            return new int[] { };
+        }
+
+        [Test]
+        public void AtMostWithSingleElementHasAtMostZeroElementsForCollections()
+        {
+            Assert.IsFalse(GetSingleElementArray().AtMost(0));
+        }
+        [Test]
+        public void AtMostWithSingleElementHasAtMostOneElementForCollections()
+        {
+            Assert.IsTrue(GetSingleElementArray().AtMost(1));
+        }
+        [Test]
+        public void AtMostWithSingleElementHasAtMostManyElementsForCollections()
+        {
+            Assert.IsTrue(GetSingleElementArray().AtMost(2));
+        }
+        private static IEnumerable<int> GetSingleElementArray()
+        {
+            return GetSingleElementSequence().ToArray();
+        }
+
+        [Test]
+        public void AtMostWithManyElementsHasAtMostZeroElementsForCollections()
+        {
+            Assert.IsFalse(GetManyElementArray().AtMost(0));
+        }
+        [Test]
+        public void AtMostWithManyElementsHasAtMostOneElementForCollections()
+        {
+            Assert.IsFalse(GetManyElementArray().AtMost(1));
+        }
+        [Test]
+        public void AtMostWithManyElementsHasAtMostThreeElementsForCollections()
+        {
+            Assert.IsTrue(GetManyElementArray().AtMost(3));
+        }
+        [Test]
+        public void AtMostWithManyElementsHasAtMostManyElementsForCollections()
+        {
+            Assert.IsTrue(GetManyElementArray().AtMost(4));
+        }
+        private static IEnumerable<int> GetManyElementArray()
+        {
+            return GetManyElementSequence().ToArray();
+        }
+
+        [Test]
+        public void AtMostShouldBeNotEnumerateSequenceForImplementersOfICollection()
+        {
+            var sequence = new UnenumerableList<int>();
+            sequence.AtMost(3);
+        }
+    }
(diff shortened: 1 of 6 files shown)


MoreLinq.Test/AtMostTest.cs
```diff
@@ -0,0 +1,176 @@
...
+        public void AtMostWithEmptySequenceHasAtMostOneElement()
+        {
+            Assert.IsTrue(GetEmptySequence().AtMost(1));
+        }
+        private static IEnumerable<int> GetEmptySequence()
+        {
+            return LinqEnumerable.Empty<int>();
+        }
+
+		[Test]
```
sholland1: Formatting.

MoreLinq.Test/AtMostTest.cs
```diff
@@ -0,0 +1,176 @@
...
+        public void AtMostWithEmptySequenceHasAtMostOneElementForCollections()
+        {
+            Assert.IsTrue(GetEmptyArray().AtMost(1));
+        }
+        private static IEnumerable<int> GetEmptyArray()
+        {
+            return new int[] { };
+        }
+
+		[Test]
```
sholland1: Formatting.

MoreLinq.Test/AtMostTest.cs
```diff
@@ -0,0 +1,176 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
```
sholland1: The copyright notices should correspond to the author aka you. See 57295fff782ddc22d57c3883ed70dd77bd06e338.

MoreLinq/AtMost.cs
```diff
@@ -0,0 +1,64 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2015 sholland. All rights reserved.
```
sholland1: The copyright notices should correspond to the author. See 57295fff782ddc22d57c3883ed70dd77bd06e338.

jnyrup:
@sholland1 Thanks for the review!
They should all be fixed now and squashed into a single commit.

MoreLinq/AtMost.cs
```diff
@@ -0,0 +1,64 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Returns true when the number of elements in the given sequence is smaller than
+        /// or equal to the given integer.
+        /// This method throws an exception if the given integer is negative.
```
atifaziz: This doesn't belong in the summary and is already point out in line 40:

    /// <exception cref="ArgumentOutOfRangeException">count is negative</exception>

MoreLinq/AtMost.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// </summary>
+        /// <remarks>
+        /// The number of items streamed will be less than or equal to the given integer
```
atifaziz: This remark seems incorrect.

MoreLinq/AtMost.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <param name="source">The source sequence</param>
+        /// <param name="count">The maximum number of items a sequence must have for this
+        /// function to return true</param>
+        /// <exception cref="ArgumentNullException">source is null</exception>
+        /// <exception cref="ArgumentOutOfRangeException">count is negative</exception>
+        /// <returns><c>true</c> if the number of elements in the sequence is smaller than
+        /// or equal to the given integer or <c>false</c> otherwise.</returns>
+        /// <example>
+        /// <code>
+        /// var numbers = { 123, 456, 789 };
```
atifaziz: This doesn't seem to be valid C#. Did you mean the following?

    var numbers = new[] { 123, 456, 789 };

MoreLinq/AtMost.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// var result = numbers.AtMost(4);
+        /// </code>
+        /// The <c>result</c> variable will contain <c>true</c>.
+        /// </example>
+        public static bool AtMost<TSource>(this IEnumerable<TSource> source, int count)
```
atifaziz: This implementation has a lot of [duplication of logic with `AtLeast`](https://github.com/morelinq/MoreLINQ/blob/037fa04aa3bd42edb319ff7d9c7ec11b9f5df0b0/MoreLinq/AtLeast.cs#L51-L60) so it would be worthwhile to try and consolidate.

MoreLinq/AtMost.cs
```diff
@@ -0,0 +1,64 @@
...
+            if (source == null) throw new ArgumentNullException("source");
+            if (count < 0) throw new ArgumentOutOfRangeException("count", "The count must not be negative.");
+
+            var collection = source as ICollection<TSource>;
+            if (collection != null)
+            {
+                return collection.Count <= count;
+            }
+
+            return source.Take(count + 1).Count() <= count;
```
atifaziz: What's the reason for taking one more than `count`?

MoreLinq/AtMost.cs
jnyrup: At most `count + 1` elements are streamed. (The "plus one" is on line 33).

MoreLinq/AtMost.cs
jnyrup: To know that `source` has *no more* than `count` elements.
If e.g. `source` has 2 elements, and `count = 1` `source.Take(count).Count() <= count` yields `true` which is not the answer we want, as it has more than 1 element.

atifaziz: @jnyrup That was a stupid question on my end. I was confusing it with something else but thanks for clarifying nonetheless.

MoreLinq/AtMost.cs
jnyrup: It indeed has -- in fact, I started out by making a copy of [AtLeast.cs](https://github.com/morelinq/MoreLINQ/blob/037fa04aa3bd42edb319ff7d9c7ec11b9f5df0b0/MoreLinq/AtLeast.cs)
To reuse the logic from `AtLeast` it could be rewritten to
`public static bool AtMost<TSource>(this IEnumerable<TSource> source, int count) => !source.AtLeast(count + 1);`
if that is preferable.

MoreLinq/AtMost.cs
atifaziz: But the result is a `Boolean` so nothing is _streamed out_. I think you mean that it will _pull_ only on those many elements from the source (unless of course it's a materialised collection)? The wording is a little confusing & unusual.

jnyrup: You're right, it is only *pulling* the elements and not streaming them, as it does not pass them on.

atifaziz:
@jnyrup @sholland1 Having `AtLeast`, `AtMost` and `Exactly` seems a little excessive when they could be folded into just one method that checks the source length against an optional min and max boundary: `HavingCount(int? min, int? max)`. Then…

* `AtLeast`: count >= min; max = null
* `AtMost`: count <= max; min = null
* `Exactly`: count = min = max`
* NOP when min and max are both null

It would avoid a lot of duplication of code and tests. If someone finds prefers to see `AtLeast`, `AtMost` or `Exactly` in code then those would be just one line wrappers.

jnyrup:
@atifaziz 
Personally I find the separate methods much clearer - both in terms of their method signatures and implementations (In *really* try to avoid nulls if at all possible). 

If the goal is to minimize the code size, I would rather try to refactor their tests suites to share more code. The reasons I didn't do that for this PR was only because I saw that the other methods I looked at each had their own test class.

`HavingCount` would have to do a quite some checks if all checking should reside in the beginning of the method.

```
public static bool HavingCount<T>(this IEnumerable<T> source, int? min, int? max)
{
    if(source == null)
        throw new ArgumentNullException("source");

    if(min == null && max == null)
        throw new ArgumentException("Max and min cannot both be null", "min, max");

    if(min != null && min < 0)
        throw new ArgumentOutOfRangeException("min", "The min must not be negative.");

    if(max != null && max < 0)
        throw new ArgumentOutOfRangeException("max", "The max must not be negative.");

    if(min != null && max != null && max < min)
        throw new ArgumentException("Max cannot be smaller than min", "min, max");

    var collection = source as ICollection<T>;

    if(min == null)
    {
        // AtMost
        if(collection != null)
            return collection.Count <= max.Value;

        return source.Take(max.Value + 1).Count() <= max.Value;
    }

    if(max == null)
    {
        // AtLeast
        if(collection != null)
            return collection.Count >= min.Value;

        return source.Take(min.Value).Count() >= min.Value;
    }

    // Between
    if(collection != null)
        return collection.Count >= min.Value && collection.Count <= max.Value;

    var count = source.Take(max.Value + 1).Count();
    return count >= min.Value && count <= max.Value;
}
```

sholland1:
I think this is a good idea, but I think we need a better name.  I suggest `CountWithinRange` or `CountBetween`.

atifaziz:
> Personally I find the separate methods much clearer

I'd like to avoid *personal* views and strive, as much as possible, for an objective discussion.

> If the goal is to minimize the code size

The goal is rather to avoid a lot of duplication of code by reusing where it seems there may be opportunities to do so. Usually that translates into reduction of code size as well.

> In _really_ try to avoid nulls if at all possible

I would see that differently. The idea is not to avoid nulls but to have optionality. If you want to embed the optionality into types (making it clear in signatures) then here it comes in the form of nullable integers. And `null` is better than magical values like `-1` to mean undefined.
 
> I would rather try to refactor their tests suites to share more code.

The users of MoreLINQ will not benefit from tests sharing code, which may be a noble goal in itself. By virtual of implementing `HavingCount`, you'll end up having a single test fixture anyhow so why not try that first?

Your `HavingCount` implementation is a good start for discussion and we should see if we can arrive at something sensible that's more reusable & general yet not too convoluted.

atifaziz:
@jnyrup Here's what I had in mind with the duplication consolidated:

```c#
public static bool HavingCount<T>(this IEnumerable<T> source, int? min, int? max)
{
    if (source == null) throw new ArgumentNullException("source");
    if (min < 0) throw new ArgumentOutOfRangeException("min", "Count must not be negative.");
    if (max < 0) throw new ArgumentOutOfRangeException("max", "Count must not be negative.");
    if (max < min) throw new ArgumentException("Maximum count cannot be smaller than the minimum.");

    if (min == null && max == null)
        return false;
    
    var collection = source as ICollection<T>;
    var count = collection != null
              ? collection.Count
              : source.Take(Math.Max(min ?? 0, (max + 1) ?? 0)).Count();
    return (min == null || count >= min) && (max == null || count <= max);
}
```

Then the rest just become one-liners:

```c#
public static bool AtLeast<T>(this IEnumerable<T> source, int count) => source.HavingCount(count, null );
public static bool AtMost<T> (this IEnumerable<T> source, int count) => source.HavingCount(null , count);
public static bool Exactly<T>(this IEnumerable<T> source, int count) => source.HavingCount(count, count);
```

jnyrup:
@atifaziz
+1 for keeping the proper tone, when I started getting sharp.

You're without doubt right about `null` being better than `-1` to represent undefined.
The optionality that nullables provide here is, as I see it, only relevant for the implementations of `AtLeast`, `AtMost` and `Exactly` and is not for the user.
I would at least suggest to hide it and let the public methods take regular ints as arguments.

````
private static bool HavingCountImpl<T>(this IEnumerable<T> source, int? min, int? max){...}
public static bool AtLeast<T>(this IEnumerable<T> source, int count) => source.HavingCountImpl(count, null );
public static bool AtMost<T> (this IEnumerable<T> source, int count) => source.HavingCountImpl(null , count);
public static bool Exactly<T>(this IEnumerable<T> source, int count) => source.HavingCountImpl(count, count);
public static bool HavingCount<T>(this IEnumerable<T> source, int min, int max) => source.HavingCountImpl(min, max);
````

While fiddling with implementations I came to think of two things: 
1) [`Take`](http://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,aae92b018fa12b39,references) already checks for `source == null`
2) It isn't (completely) nonsense to ask for at least -1 elements. An empty list does have at least -1 elements and it is false that an empty list has exactly or at most -1 elements.

So the three functions could be implemented as:
````
public static bool AtLeast<T>(this IEnumerable<T> source, int count) => ((source as ICollection<T>)?.Count ?? source.Take(count).Count()) >= count;
public static bool AtMost<T> (this IEnumerable<T> source, int count) => ((source as ICollection<T>)?.Count ?? source.Take(count + 1).Count()) <= count;
public static bool Exactly<T>(this IEnumerable<T> source, int count) => ((source as ICollection<T>)?.Count ?? source.Take(count + 1).Count()) == count;
````

atifaziz:
@jnyrup Sorry for neglecting this. Meanwhile, @leandromoh addressed most points raised here, in PR #197, and which is now part of [v2.1.0](https://github.com/morelinq/MoreLINQ/releases/tag/v2.1.0). If you felt you were left out of the follow-up and conversation in #197 then that's my poor handling of the issues.

jnyrup:
No worries. 
I followed the discussion of the other PR and I really likes his implementation.

Issue #194, opened 2016-09-04
Port for netcore


Trojaner25:
Since it references mscorlib it can not be used with dotnet core.

Please make MoreLINQ.Core or something like that :+1: 

Thank you for your awesome work.

MiffyLiye:
I ported one at https://www.nuget.org/packages/MoreLinq.Core/
And here is the pull request at https://github.com/morelinq/MoreLINQ/pull/195

atifaziz:
@MiffyLiye So following on from [your comment](https://github.com/morelinq/MoreLINQ/pull/195#issuecomment-247547797) in #195:

> OK. When we have the plan of actions, I can create a new PR.

Here's an initial plan just off the top of my head:

- Retire the `MoreLinq.Portable` project and its test project
- The `MoreLinq` project should target .NET Framework 3.5 and .NET Standard 1.x ([trimmed](https://docs.microsoft.com/en-us/dotnet/articles/core/deploying/reducing-dependencies)) via one `project.json`
- Migrate [`<metadata>` in `morelinq.nuspec`](https://github.com/morelinq/MoreLINQ/blob/14d65bfde6c5a97a68ac42d505f62bfc94d327cf/pkg/morelinq.nuspec#L4-L17) to corresponding areas of `project.json` to the extent that `morelinq.nuspec` can be done away with.
- `MoreLinq.Test` project should test for the frameworks it targets.
- Optionally enable development on non-Windows platforms using VS Code by providing Bash scripts for building and packing. I say optional because this can be done as a separate after an initial migration.

The topic of `.csproj` versus `.xproj` is debatable depending on the pros and cons but based on experience with some other projects, seems like the command-line tooling is in a bit better shape than VS ones.

Let me know if you're willing to go the distance on this or it's turned out to be more than you're willing to bite at the moment.

MiffyLiye:
@atifaziz Most actions look fine.
The only problem may be the ```MoreLinq.Test```, because it depends on ```NUnit 2.x``` which does not support .NET Core. 
We may need to upgrade NUnit  to 3.x and migrate all affected tests (at least 204 cases) before they can run on all target platforms. 

I think upgrading UNnit and tests needs to be done as a separate task before or after the migration.

MiffyLiye:
Mostly done. Changes are
* Updated the ```MoreLinq``` subproject targeting .NET Framework 3.5, PCL, and .NET Standard Library
* Migrated metadata from nuspec to project.json
[...]

PR #195, opened 2016-09-12
Add project file to build MoreLinq to .NET Core

MiffyLiye:
To build MoreLinq with .NET Core, I added a .NET Core project file project.json.
It excluded files in ```["SequenceException.cs", "ToDataTable.cs", "GroupAdjacent.cs", "Trace.cs"] ```and successfully got a release build. 

After dotnet core environment is ready, I ran
```shell
dotnet restore MoreLinq/project.json
dotnet pack -c Release MoreLinq/project.json
```
and got the nuget package.

Now it is configured to only support .NETCoreApp 1.0 platform but we can add more later.

### DIFF
--- a/MoreLinq/project.json
+++ b/MoreLinq/project.json
@@ -1,0 +1,28 @@
+﻿{
+    "name": "MoreLinq.Core",
+    "version": "2.0.0-beta04",
+    "buildOptions": {
+        "debugType": "portable",
+        "emitEntryPoint": false,
+        "warningsAsErrors": true,
+        "compile": {
+            "exclude": [
+                "SequenceException.cs",
+                "ToDataTable.cs",
+                "GroupAdjacent.cs",
+                "Trace.cs"
+            ]
+        }
+    },
+    "frameworks": {
+        "netcoreapp1.0": {
+            "dependencies": {
+                "Microsoft.NETCore.App": {
+                    "type": "platform",
+                    "version": "1.0.0"
+                }
+            },
+            "imports": "dnxcore50"
+        }
+    }
+}


Trojaner25:
👍

atifaziz:
@MiffyLiye Ideally addressing #194 should not introduce yet a new package to maintain just for .NET Core. Sure, it's easy but it's not integral to the project & could confuse people. What would help instead is to retire the [MoreLinq.Portable](https://www.nuget.org/packages/MoreLinq.Portable/) and instead have just the [base MoreLINQ package](https://www.nuget.org/packages/morelinq/) host the right binaries per targeted framework.

MiffyLiye:
@atifaziz Do you mean building artefacts for different platforms and package them in one MoreLINQ package? 
I agree a single package targeting multiple platforms is much better than introducing a new package. 
```dotnet build -c Release MoreLinq/project.json``` will produce the .NET Core binaries. 
Add we can add the artefacts to the base package. 
I can delete MoreLinq.Core on nuget.org after the base MoreLINQ packages supports .NET Core.
Do you need help updating the build & package scripts?

atifaziz:
> Do you mean building artefacts for different platforms and package them in one MoreLINQ package? 

Yes.

> `dotnet build -c Release MoreLinq/project.json` will produce the .NET Core binaries.
> Add we can add the artefacts to the base package. 

Why not just move to `.xproj`? I know we'll [eventually have to revert back to `.csproj`](https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/) but I reckon that's still several months away.

> Do you need help updating the build & package scripts?

Yes, any help would be appreciated. Should we close this PR, come up with a plan of action in #194 and then go with a new PR?

MiffyLiye:
>Why not just move to .xproj? I know we'll eventually have to revert back to .csproj but I reckon that's still several months away.

OK. I can add ```.xproj / project.json```, and include them in ```*.sln```. 

>Yes, any help would be appreciated. Should we close this PR, come up with a plan of action in #194 and then go with a new PR?

OK. When we have the plan of actions, I can create a new PR.

CADbloke:
> Why not just move to .xproj? I know we'll eventually have to revert back to .csproj but I reckon that's still several months away.

Because those of us who were using the csproj now don't have one. 

Just an FYI, I can work around it. I first noticed it when it crashed https://github.com/CADbloke/CodeLinker which I use to strip the guts out of other CSPROJ files to then use that guts in projects with different build settings.

atifaziz:
> Because those of us who were using the csproj now don't have one.

@CADbloke I'm guessing `.xproj` is not an option, or is it just that you prefer to use `.csproj`. I can understand that not everyone can be on the latest & greatest so we can discuss bringing the `.csproj` back (perhaps as a [new issue](https://github.com/morelinq/MoreLINQ/issues/new)).

CADbloke:
@atifaziz Thanks for replying. Xproj *was* the latest and greatest but it seems to have been "Silverlighted" so no, I won't be adopting it. I haven't bothered adding xproj-support to Code Linker, mostly because this is the only project that I have that uses it and I seem to be the only person who uses Code Linker. 

Don't change direction simply on my case, I can work around it.

cheers

PR #196, opened 2016-09-18
Port for .NET Core

MiffyLiye:
Port as planned in https://github.com/morelinq/MoreLINQ/issues/194
* Retired the ```MoreLinq.Portable``` project and its test project
* The ```MoreLinq``` project targets .NET Framework 3.5, PCL and .NET Standard Library 1.6 via one ```project.json```
* The ```<metadata>``` in ```morelinq.nuspec``` are migrated to corresponding areas of ```project.json```
* ```MoreLinq.Test``` project can test for the .NET 3.5 (in .NET 4) and PCL (in .NET 4 Client) as there were before the port. .NET Standard Library 1.6 is not tested because NUnit 2 does not support .NET Core. 
* build, test, and pack scripts are provided (as ```powershell``` scripts).
* A CI config (```appveyor.yml``) is added to build, test, and pack on appveyor.com after each push. Project owner needs to turn on continuous integration of this project on https://www.appveyor.com to demonstrate the build status and to get packages for each build.

For development, 
* Build can be done with both Visual Studio 2015, and ```build.ps1```
* After build, tests can run with ```test.ps1``` (which uses ```.\tools\nuget.exe``` from  https://dist.nuget.org/win-x86-commandline/latest/nuget.exe to download test runners). If developers use other test runners, they may not need ```test.ps1``` and ```.\tools\nuget.exe```
* After  build, pack with ```pack.ps1```. The package named ```*.symbols.nupkg```  contains binaries, documents, and source code, the other ```*.nupkg``` does not contain source code.


build.ps1
```diff
@@ -0,0 +1,4 @@
+dotnet --info
```
atifaziz: Switching to PowerShell scripts was not part of the plan for this PR (can be discussed/raised as a separate issue) so kindly revert to using `build.cmd`. `build.ps1` is adding no value at the moment because in the end it's just calling `dotnet.exe` to do all the work and Windows batch scripts to do that more naturally. In fact, it worsens the situation right now because in the PowerShell version, all 4 steps will execute ignoring any errors in a previous step.

builddocs.cmd
```diff
@@ -1 +0,0 @@
-@"%~dp0build" docs
```
atifaziz: Documentation building needs to be restored as there was no agreement to dropping it as part of targeting .NET Core.

pack.ps1
```diff
@@ -0,0 +1 @@
+dotnet pack -c Release .\MoreLinq\project.json
```
atifaziz: Once more, switching to PowerShell scripts was not part of the plan for this PR (can be discussed/raised as a separate issue) so kindly revert to using `pack.cmd`. `pack.ps1` is adding no value at the moment (in fact, removes a lot) because in the end it's just calling `dotnet.exe` to do all the work.

pkg/MoreLinq.Source.MoreEnumerable.Acquire.nuspec
```diff
@@ -1,27 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
atifaziz: All source packages appear to have been and need to be restored because they are independent of issue #194.

test.ps1
```diff
@@ -0,0 +1,7 @@
+Remove-Item -Recurse .\tools\NUnit.Runners.*
+.\tools\nuget install NUnit.Runners -Version 2.6.4 -OutputDirectory tools
```
atifaziz: Not particularly keen about maintaining a 3MB executable in the repo just for the test script. It would be equally effective to use `Invoke-WebRequest` to download a specific version if it's not already there, i.e.:

    Invoke-WebRequest https://dist.nuget.org/win-x86-commandline/v3.4.4/NuGet.exe -OutFile .\tools\nuget.exe

Re-installing NUnit on each run should be avoided. If one has a repo cloned and working offline then it becomes impossible to run tests. Tests should be runnable offline if they have been run at least once while online (and while it would be ideal to have an offline setup right after cloning, this is still better than not being able to run them at all).

fsateler: I recommend against running arbitrary binaries from the internet. It becomes an attack vector.

Nuget is likely to be installed (or easily installed) on a developer machine, why not rely on it being existing and in PATH? I certainly don't want a nuget copy for each project I work on.

appveyor.yml
```diff
@@ -0,0 +1,23 @@
...
+  - ps: .\dotnet-install.ps1 -Version 1.0.0-preview2-003121
+  - ps: Remove-Item .\dotnet-install.ps1
+build_script:
+  - ps: dotnet --info
+  - ps: dotnet restore
+  - ps: dotnet build -c Debug .\*\project.json
+  - ps: dotnet build -c Release .\*\project.json
+before_test:
+  - ps: New-Item -ItemType Directory -Path tools -Force
+  - ps: Invoke-WebRequest "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe" -OutFile ".\tools\nuget.exe"
```
fsateler: It appears nuget is already available in appveyor, so no need to install? https://www.appveyor.com/docs/nuget/#restoring-nuget-packages-before-build

MiffyLiye: Yes, you are right. Updated appveyor.yml

test.cmd
```diff
@@ -0,0 +1,24 @@
+@echo off
+setlocal
+cd "%~dp0"
+chcp 1252 > nul
+if not exist tools\NUnit.Runners (
+    for %%i in (NuGet.exe) do set nuget=%%~dpnx$PATH:i
+    if "%nuget%"=="" goto :nonuget
```
atifaziz: This check here causes `test.cmd` to _always_ fail with the following output irrespective of it `NuGet.exe` is in the path or not:

```
NuGet executable not found in PATH
For more on NuGet, see http://nuget.codeplex.com
```

The reason is that the outer `if` along with its block is expanded _before_ being evaluated so `if "%nuget%"=="" goto :nonuget` becomes `if ""=="" goto :nonuget` at that point and will always be true. I would recommend using a simpler approach by refactoring the block as follows (so that each line is expanded as its evaluated):

```cmd
if exist tools\NUnit.Runners goto :test
for %%i in (NuGet.exe) do set nuget=%%~dpnx$PATH:i
if "%nuget%"=="" goto :nonuget
md tools
nuget.exe install NUnit.Runners -Version 2.6.4 -OutputDirectory tools
move .\tools\NUnit.Runners.2.6.4 .\tools\NUnit.Runners
if not %errorlevel%==0 exit /b %errorlevel%
:test
```

test.cmd
```diff
@@ -0,0 +1,24 @@
...
+    md tools
+    nuget.exe install NUnit.Runners -Version 2.6.4 -OutputDirectory tools
```
atifaziz: Missing check if `nuget install` failed or not. Perhaps use something like this?

```cmd
nuget.exe install NUnit.Runners -Version 2.6.4 -OutputDirectory tools || exit /b 1
```

test.cmd
```diff
@@ -0,0 +1,24 @@
...
+    move .\tools\NUnit.Runners.2.6.4 .\tools\NUnit.Runners
+)
+if not %errorlevel%==0 exit /b %errorlevel%
+
+call build /v:m
+.\tools\NUnit.Runners\tools\nunit-console.exe .\MoreLinq.Test\bin\Debug\net40\MoreLinq.Test.dll
```
atifaziz: If building fails in the previous step then tests shouldn't be run. Would be better to predicate each on the success of the previous using `&&`, as in:

```cmd
call build /v:m ^
  && .\tools\NUnit.Runners\tools\nunit-console.exe .\MoreLinq.Test\bin\Debug\net40\MoreLinq.Test.dll ^
  && .\tools\NUnit.Runners\tools\nunit-console.exe .\MoreLinq.Test\bin\Release\net40\MoreLinq.Test.dll ^
  && .\tools\NUnit.Runners\tools\nunit-console.exe .\MoreLinq.Test\bin\Debug\net40-client\MoreLinq.Test.dll ^
  && .\tools\NUnit.Runners\tools\nunit-console.exe .\MoreLinq.Test\bin\Release\net40-client\MoreLinq.Test.dll
```

test.cmd
MiffyLiye: Thanks for the detailed explanations. Updated ```test.cmd```.

atifaziz:
@MiffyLiye Thanks for working through all that feedback. I'm justing seeing something very bizarre now with the tests. While I haven't had the time to study the issue in detail I wanted to share my findings thus far and check if you're seeing the same and find it normal.

Here's the output from one of the test runs (the remaining being pretty much the same):

```
NUnit-Console version 2.6.4.14350
Copyright (C) 2002-2012 Charlie Poole.
Copyright (C) 2002-2004 James W. Newkirk, Michael C. Two, Alexei A. Vorontsov.
Copyright (C) 2000-2002 Philip Craig.
All Rights Reserved.

Runtime Environment -
   OS Version: Microsoft Windows NT 6.2.9200.0
  CLR Version: 2.0.50727.8689 ( Net 3.5 )

ProcessModel: Default    DomainUsage: Single
Execution Runtime: net-3.5
..........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................N.....N....................................................................................................................................................................................................................................................
Tests run: 865, Errors: 0, Failures: 0, Inconclusive: 0, Time: 4.09555907414573 seconds
  Not run: 2, Invalid: 0, Ignored: 2, Skipped: 0

Tests Not Run:
1) Ignored : MoreLinq.Test.PartialSortByTests.PartialSortByIsStable
   TODO
2) Ignored : MoreLinq.Test.PartialSortTests.PartialSortByIsStable
   TODO
```

The surprising bit is that the NUnit runner is compiled for and running on CLR 2 yet `MoreLinq.Test.dll` is compiled to run against CLR 4. The former shouldn't be able to load the latter.

Running [CorFlags](https://msdn.microsoft.com/en-us/library/ms164699(v=vs.110).aspx) on `nunit-console.exe` gives:

```
Microsoft (R) .NET Framework CorFlags Conversion Tool.  Version  4.6.1055.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Version   : v2.0.50727
CLR Header: 2.5
PE        : PE32
CorFlags  : 0x1
ILONLY    : 1
32BITREQ  : 0
32BITPREF : 0
Signed    : 0
```

And for `MoreLinq.Test.dll`:

```
Microsoft (R) .NET Framework CorFlags Conversion Tool.  Version  4.6.1055.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Version   : v4.0.30319
CLR Header: 2.5
PE        : PE32
CorFlags  : 0x1
ILONLY    : 1
32BITREQ  : 0
32BITPREF : 0
Signed    : 0
```

atifaziz:
Just quickly checked with running PowerShell on CLR 2 and 4 and trying to load `MoreLinq.Test.dll` and just as I expected, CLR 2 cannot load assemblies compiled for CLR 4. That is, doing:

    B:\>PowerShell -Version 2 -C "$PSVersionTable; Add-Type -Path B:\MoreLinq.Test.dll"

failed with:

```
Name                           Value                                                                                           
----                           -----                                                                                           
CLRVersion                     2.0.50727.8689                                                                                  
BuildVersion                   6.1.7600.16385                                                                                  
PSVersion                      2.0                                                                                             
WSManStackVersion              2.0                                                                                             
PSCompatibleVersions           {1.0, 2.0}                                                                                      
SerializationVersion           1.1.0.1                                                                                         
PSRemotingProtocolVersion      2.1                                                                                             
Add-Type : Could not load file or assembly 'file:///B:\MoreLinq.Test.dll' or one of its dependencies. This assembly is built by
 a runtime newer than the currently loaded runtime and cannot be loaded.
At line:1 char:26
+ $PSVersionTable; Add-Type <<<<  -Path B:\MoreLinq.Test.dll
    + CategoryInfo          : NotSpecified: (:) [Add-Type], BadImageFormatException
    + FullyQualifiedErrorId : System.BadImageFormatException,Microsoft.PowerShell.Commands.AddTypeCommand
```

And then:

    B:\>PowerShell -Version 4 -C "$PSVersionTable; Add-Type -Path B:\MoreLinq.Test.dll"

naturally succeeded with no error in the output:

```
Name                           Value                                                                                           
----                           -----                                                                                           
PSVersion                      5.0.10586.494                                                                                   
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                         
BuildVersion                   10.0.10586.494                                                                                  
CLRVersion                     4.0.30319.42000                                                                                 
WSManStackVersion              3.0                                                                                             
PSRemotingProtocolVersion      2.3                                                                                             
SerializationVersion           1.1.0.1                                                                                         


```

😖

MiffyLiye:
@atifaziz It seems that NUnit can auto select the runtime and run out of process if necessary (http://nunit.org/index.php?p=runtimeSelection&r=2.6.4)
Now it selects .NET 3.5 as the execution runtime, and maybe we can use ```nunit-console.exe MoreLinq.Test.dll /framework:net-4.0``` to set the preferred execution runtime.

atifaziz:
@MiffyLiye Yes, I just read up on NUnit runtime selection this morning so that does explain why it's working. I also attached the debugger to the `nunit-agent.exe` process and it was indeed using CLR 4. The `nunit-agent.exe.config` also has an interesting comment:

```xml
  <startup useLegacyV2RuntimeActivationPolicy="true">
    <!--
     Nunit-agent is normally run by the console or gui
     runners and not independently. In normal usage,
     the runner specifies which runtime should be used.

     Do NOT add any supportedRuntime elements here,
     since they may prevent the runner from controlling
     the runtime that is used!
    -->
  </startup>
```

However, what was throwing me off was what the runner was reporting on the console. I think that's probably misleading and [might be coming from `nunit-console.exe`](https://github.com/nunit/nunitv2/blob/2.6.4/src/ConsoleRunner/nunit-console/Runner.cs#L153-L157) that still runs under CLR 2 but launches `nunit-agent.exe` on CLR 4.

MiffyLiye:
OK, removed them in this PR.

MiffyLiye:
Fixed this incorrect  reference. It is strange that ```msbuild``` does not give  any warnings.
I checked them again,  and found ```Debug``` and ```Release``` configurations are missing, so also fixed.

atifaziz:
> …found Debug and Release configurations are missing, so also fixed.

Good catch though did it change anything for you? When you say “fixed”, is that for sake of completeness/explicitness? I'm not seeing any impact of the change in terms of the compiled outputs (meaning those configurations must be implicitly defined with the same options) but just wanted to double-check.

MiffyLiye:
I expect this commit is only for sake of explicitness. 
I tested the DEBUG and TRACE symbol with default Debug and Release build, found they are implicitly defined with the same options.
And one of my colleagues tested years ago that default Release build will optimize IL code while Debug build will not.

I haven't find any msdn documents describing the difference between default Debug and Release build. 
So it makes me feel safer to make them explicit.

MoreLinq/project.json
```diff
@@ -0,0 +1,112 @@
+{
+    "name": "MoreLinq",
```
atifaziz: This should be `morelinq`, all in lowercase.

MoreLinq/project.json
```diff
@@ -0,0 +1,112 @@
...
+        "licenseUrl": "http://www.apache.org/licenses/LICENSE-2.0",
+        "requireLicenseAcceptance": false,
+        "tags": [
+            "linq",
+            "extensions"
+        ],
+        "files": {
+            "include": [
+                "../COPYING.txt",
+                "../lic/corefx/LICENSE"
```
atifaziz: This one seems to be missing from the output package.

atifaziz:
@MiffyLiye Thanks for the explanations.

I ran a diff on the package produced by the `project.json` and the `nuspec` it replaces:

```diff
--- A:/MoreLINQ/tmp/morelinq.2.0.0-beta04.nuspec	Thu Sep 22 18:48:41 2016
+++ A:/MoreLINQ/main/dist/MoreLinq.2.0.0-beta04.nuspec	Thu Sep 22 18:50:18 2016
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
     <metadata>
-        <id>morelinq</id>
+        <id>MoreLinq</id>
         <version>2.0.0-beta04</version>
         <title>MoreLINQ</title>
         <authors>MoreLINQ Developers.</authors>
@@ -11,15 +11,33 @@
         <requireLicenseAcceptance>false</requireLicenseAcceptance>
         <description>This project enhances LINQ to Objects with the following methods: Acquire, Assert, AssertCount, Batch, Cartesian, Concat, Consume, DistinctBy, EquiZip, ExceptBy, Exclude, Fold, ForEach, Generate, GenerateByIndex, GroupAdjacent, Incremental, Index, Interleave, Lag, Lead, MaxBy, MinBy, NestedLoops, OrderBy, OrderedMerge, Pad, Pairwise, PartialSort, PartialSortBy, Partition, Permutations, Pipe, Prepend, PreScan, Random, RandomDouble, RandomSubset, Rank, RankBy, Repeat, RunLengthEncode, Scan, Segment, SingleOrFallback, SkipUntil, Slice, SortedMerge, Split, Subsets, TagFirstLast, TakeEvery, TakeLast, TakeUntil, ThenBy, ToDataTable, ToDelimitedString, ToHashSet, Trace, Windowed, ZipLongest, ZipShortest</description>
         <summary>This project enhances LINQ to Objects with extra methods, in a manner which keeps to the spirit of LINQ.</summary>
-        <releaseNotes>Adds new operators: Assert, Cartesian, Exclude, FallbackIfEmpty, Incremental, Interleave, Lag, Lead, NestedLoops, OrderBy, Partition, Permutations, Random, RandomDouble, RandomSubset, Rank, RankBy, Repeat, RunLengthEncode, Segment, Slice, SortedMerge, Subsets, TagFirstLast, ThenBy, Windowed, ZipShortest. See also https://github.com/morelinq/MoreLINQ/wiki/API-Changes.</releaseNotes>
+        <releaseNotes>Adds new operators: Assert, Cartesian, Exclude, FallbackIfEmpty, FullGroupJoin, Incremental, Interleave, Lag, Lead, NestedLoops, OrderBy, Partition, Permutations, Random, RandomDouble, RandomSubset, Rank, RankBy, Repeat, RunLengthEncode, Segment, Slice, SortedMerge, Subsets, TagFirstLast, ThenBy, Windowed, ZipShortest. See also https://github.com/morelinq/MoreLINQ/wiki/API-Changes.</releaseNotes>
         <copyright>Copyright © 2008 Jonathan Skeet. All rights reserved. Portions Copyright (c) Microsoft. All rights reserved.</copyright>
         <language>en-US</language>
-        <tags>linq extensions</tags>
+        <tags>linq,extensions</tags>
+        <dependencies>
+            <group targetFramework=".NETFramework3.5" />
+            <group targetFramework=".NETPortable4.0-Profile328" />
+            <group targetFramework=".NETStandard1.6">
+                <dependency id="NETStandard.Library" version="1.6.0" />
+            </group>
+        </dependencies>
+        <frameworkAssemblies>
+            <frameworkAssembly assemblyName="System.Data.DataSetExtensions" targetFramework=".NETFramework3.5" />
+            <frameworkAssembly assemblyName="System.Data" targetFramework=".NETFramework3.5" />
+            <frameworkAssembly assemblyName="System.Xml" targetFramework=".NETFramework3.5" />
+            <frameworkAssembly assemblyName="mscorlib" targetFramework=".NETPortable4.0-Profile328" />
+            <frameworkAssembly assemblyName="System" targetFramework=".NETPortable4.0-Profile328" />
+            <frameworkAssembly assemblyName="System.Core" targetFramework=".NETPortable4.0-Profile328" />
+        </frameworkAssemblies>
     </metadata>
     <files>
         <file src="lib\net35\MoreLinq.dll" target="lib\net35\MoreLinq.dll" />
         <file src="lib\net35\MoreLinq.xml" target="lib\net35\MoreLinq.xml" />
-        <file src="lic\corefx\LICENSE" target="lic\corefx\LICENSE" />
+        <file src="lib\netstandard1.6\MoreLinq.dll" target="lib\netstandard1.6\MoreLinq.dll" />
+        <file src="lib\netstandard1.6\MoreLinq.xml" target="lib\netstandard1.6\MoreLinq.xml" />
+        <file src="lib\portable40-net40+sl5+win8+wp8+wpa81\MoreLinq.dll" target="lib\portable40-net40+sl5+win8+wp8+wpa81\MoreLinq.dll" />
+        <file src="lib\portable40-net40+sl5+win8+wp8+wpa81\MoreLinq.xml" target="lib\portable40-net40+sl5+win8+wp8+wpa81\MoreLinq.xml" />
         <file src="COPYING.txt" target="COPYING.txt" />
     </files>
 </package>
\ No newline at end of file
```

Two things jumped out:

1. The package Id changed cases and [this could lead to much unneeded pain](http://adamralph.com/2015/03/04/nuget-gallery-casing-woes/).
1. `lic\corefx\LICENSE` didn't make it into the package.

Other changes are understandable (although seems odd `tags` is now comma-delimited).

MiffyLiye:
I restored the package id to lower case.
The second issue cannot be reproduced. 

I unzipped the two ```nupkg``` from https://ci.appveyor.com/project/MiffyLiye/morelinq/build/2.0.0-beta04-11/artifacts and they both have the ```lic/corefx/LICENSE```. 
Also, the ```nuspec``` in the ```nupkg``` does not contain the ```<files>``` section.

atifaziz:
@MiffyLiye Thanks for testing on your end. Looks like [NuGetPackageExplorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer) that I was using to re-generate the `.nuspec` from the `.nupkg` for comparison may be glitched. It doesn't show the `lic` folder…

capture

…but which is indeed there when list package content as a ZIP:

```
7-Zip [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21

Scanning the drive for archives:
1 file, 217846 bytes (213 KiB)

Listing archive: MoreLinq.2.0.0-beta04.nupkg

--
Path = MoreLinq.2.0.0-beta04.nupkg
Type = zip
Physical Size = 217846

   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
2016-09-20 15:16:48 .....          276          199  _rels\.rels
2016-09-20 15:16:48 .....         2949         1179  MoreLinq.nuspec
2016-09-20 15:16:48 .....       127488        52527  lib\net35\MoreLinq.dll
2016-09-20 15:16:48 .....       216576        19956  lib\net35\MoreLinq.xml
2016-09-20 15:16:48 .....       119808        49321  lib\portable40-net40+sl5+win8+wp8+wpa81\MoreLinq.dll
2016-09-20 15:16:48 .....       212123        19453  lib\portable40-net40+sl5+win8+wp8+wpa81\MoreLinq.xml
2016-09-20 15:16:48 .....       122368        49934  lib\netstandard1.6\MoreLinq.dll
2016-09-20 15:16:48 .....       212123        19453  lib\netstandard1.6\MoreLinq.xml
2016-09-20 15:16:48 .....        10174         3561  COPYING.txt
2016-09-20 15:16:48 .....         1082          631  lic\corefx\LICENSE
2016-09-20 15:16:48 .....          468          206  [Content_Types].xml
------------------- ----- ------------ ------------  ------------------------
2016-09-20 15:16:48            1025435       216420  11 files
```

atifaziz:
Closed by db8ebdf606980215e2665e24bff70c72b976ec65

@MiffyLiye Thanks for all your help & working through the reviews.

PR #197, opened 2016-09-24
Implemented quantity methods (AtLeast, AtMost, Exactly and CountBetween)

leandromoh:
Hello. I saw that the repository has one method `AtLeast`but no `AtMost`. I perceive also that the current `AtLeast`code use `return source.Take(count).Count() == count;`  that is O(count) + O(count) what is unnecessary if we want just count. 

From there I improved the code for just count (improving performance); and using a generic implementation method, I also created the `AtMost`method that was missing and the `Exactly`e `CountBetween`methods, that check if an IEnumerable has at most the given integer of elements, if the length of a IEnumerable is the given integer and if the length is between an certain range, respectively.


leandromoh:
Helo @atifaziz ! How are you? 
What changed in the project in the last week?
When I did the pull request it had no conflicts but now It has in the following files:

MoreLinq.Test/MoreLinq.Test.csproj
MoreLinq/MoreLinq.Portable.csproj
MoreLinq/MoreLinq.csproj

atifaziz:
@leandromoh Issue #194 got addressed is what happened. As part of that, the project system has been migrated from `.csproj` to `.xproj`. Is this posing a big issue?

leandromoh:
@atifaziz I dont know much about .net core projects, but I didn't use any windows plataform resource, so I think the migration isn't a problem here.

I searched the references of the MoreLinq methods (that were on csproj) through the project and I found no references. What I must do? Just delete the .csproj files?

atifaziz:
> What I must do? Just delete the .csproj files?

If you're rebasing your branch on top of master then you will probably run into a merge around the `.csproj` files. You can just delete them to resolve the conflict. The new `.xproj` will pick up your new sources auto-magically.

> I dont know much about .net core projects

You don't or shouldn't need to know much about .NET Core at this stage. We made the switch to `.xproj` to simplify targeting multiple runtimes instead of having to manage multiple `.csproj` files.

leandromoh:
@atifaziz thanks for the explanation. I will delete the .csproj files and commit again as soon as possible; But I still have two questions:

1- I saw that in the MoreLINQ/project.json file has the sections "copyright" and "releaseNotes", this last  mentioning some methods of the repository. I must add my methods as well copyright in this file or it will be done for someone in the next release?

2 - For upcoming pull requests, how can I test the new methods in the Test project?

atifaziz:
> 1- I saw that in the MoreLINQ/project.json file has the sections "copyright" and "releaseNotes", this last mentioning some methods of the repository. I must add my methods as well copyright in this file or it will be done for someone in the next release?

You must add it manually. Assuming it's your original work, add the initial year of publication for the copyright.

> 2 - For upcoming pull requests, how can I test the new methods in the Test project?

Run [`test.cmd`](https://github.com/morelinq/MoreLINQ/blob/21cabccba1cfcfb5d6a21abe05ebafa22a85fa6c/test.cmd).

leandromoh:
I understood, thanks for the explanation again. 

I was thinking about quantity methods implementation... Do you think they **must** be generic ? 

I believe the old implementation might use `IEnumarable<T>` because the LINQ method `Take `(that old implementation uses within), needs a `IEnumerable<T>`; 

But make sense to me a `non-generic IEnumerable` use `AtLeast`, `AtMost`, etc methods since all I want to know is about the `count`. What do you think?

leandromoh:
@atifaziz  all conflicts were solved.

MoreLinq/QuantityIterator.cs
```diff
@@ -0,0 +1,54 @@
...
+            if (col != null)
+            {
+                return predicate(col.Count);
+            }
+
+            var count = 0;
+
+            using (var e = source.GetEnumerator())
+            {
+                while (e.MoveNext())
```
fsateler: Why use this instead of `foreach` ?

fsateler:
I think having 3 separate files for 3 thin wrappers around a single function is a bit confusing. Maybe all the functions can be folded into a single file, similar to how eg, `GroupAdjacent` and `GroupAdjacentBy` are in the same file.

MoreLinq/QuantityIterator.cs
leandromoh: besides `MoveNext()`, `foreach` would also set the current element (in this case, something like 
`T element = (T)e.Current;` which is an unecessary operation since all we want to know is if there is a next element, not to read the value of this element and assign it to an object. 
That is, for an `IEnumerable<T>` with _n_ elements, we would perform _n_ unecessary operations. For long `IEnumerables`, I preferred to be on the safe side and did this way, to get maximum performance.

leandromoh:
I agree that folding the 5 functions (AtLeast, AtMost, Exactly, CountBetween and QuantityIterator) into a single file would be best to read, but if someone, in the future, rewrite a single function of these with a different logic (no more using the generic implementation), will not make more sense all in the same file... other case: how would be about copyright message in the header of the file? two copyright messages in the same file? It would became confusing.

atifaziz:
>  There is somethign wrong with the PR ? If not, when you plan to accept it ?

@leandromoh I haven't forgotten. Just a question of time. MoreLINQ is in lock down mode right now for releasing 2.0 (any day now). This means some waiting for new additions.

> I am planning to do others PR…However, since each PR modify the file `MoreLINQ/project.json` (release notes) I am planning to do they in sequence to avoid conflicts (besides it, some PR depends of others)

You should not wait and go ahead and make those PRs. If you want to avoid conflicts with the things you mentioned then those can always be done after reviewing the implementation together (rebasing and squashing). If one PR depends on another then that's fine too. It will only demonstrate the value of the initial PRs.

leandromoh:
> I haven't forgotten. Just a question of time. MoreLINQ is in lock down mode right now for releasing 2.0 (any day now). This means some waiting for new additions.

@atifaziz  I didnt know about releasing 2.0, thanks for the explanation.

> You should not wait and go ahead and make those PRs. If you want to avoid conflicts with the things you mentioned then those can always be done after reviewing the implementation together (rebasing and squashing). If one PR depends on another then that's fine too. It will only demonstrate the value of the initial PRs.

You are right, I will begin to code in my freed time and soon I will submit the news PR's.

Sorry for the ignorance, but I didn't understand the "enhancement" label.

atifaziz:
@leandromoh The _enhancement_ label means that this PR adds a new feature to the project. It's not, for example, a fix for a bug in an existing feature.

fsateler:
On 20 October 2016 at 23:05, Leandro Fernandes <notifications@github.com>
wrote:

> I agree that folding the 5 functions (AtLeast, AtMost, Exactly,
> CountBetween and QuantityIterator) into a single file would be best to
> read, but if someone, in the future, rewrite a single function of these
> with a different logic (no more using the generic implementation), will not
> make more sense all in the same file...
>

Then it can be split into a separate file.


> other case: how would be about copyright message in the header of the
> file? two copyright messages in the same file? It would became confusing.
>

I see no problem with that. See for example this file from the csound
project:
https://github.com/csound/csound/blob/8b6608d0c320d65fc80df3fee55aedbe030f45ca/Top/csound.c#L7-L9

>


-- 

Saludos,
Felipe Sateler

leandromoh:
@fsateler you convinced me, the methods were folded into a single file (CountMethods.cs) :)

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,169 @@
...
+        public static bool Exactly<T>(this IEnumerable<T> source, int length)
+        {
+            if (length < 0) throw new ArgumentOutOfRangeException("length", "length must not be negative.");
+
+            return QuantityIterator(source, length + 1, count => count == length);
+        }
+
+        /// <summary>
+        /// Returns true when the number of elements in the given sequence is between (inclusive)
+        /// to the minimum and maximum given integer.
```
atifaziz: Consider rewording the summary to:
> Determines whether or not the number of elements in the sequence is between an inclusive range of minimum and maximum integers.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// This method throws an exception if the minimum given integer is negative
+        /// or if the maximun given integer is lesser than the minimum integer.
+        /// </summary>
```
atifaziz: The summary need not talk about exceptional cases. Consider removing entirely.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// <remarks>
+        /// The number of items streamed will be greater than or equal to the given integer.
```
atifaziz: Which integer? There are two of them. And you mean the number of items that will be _consumed_ rather than being _streamed out_? Generally streaming implies passing through so it causes word confusion here. My hunch is that you want to indicate how much of the source is read or consumed by this method.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,169 @@
...
+        /// </remarks>
+        /// <typeparam name="T">Element type of sequence</typeparam>
+        /// <param name="source">The source sequence</param>
+        /// <param name="min">The minimum number of items a sequence must have for this
+        /// function to return true</param>
+        /// <param name="max">The maximun number of items a sequence must have for this
+        /// function to return true</param>
+        /// <exception cref="ArgumentNullException">source is null</exception>
+        /// <exception cref="ArgumentOutOfRangeException">min is negative</exception>
+        /// <exception cref="ArgumentOutOfRangeException">max is lesser than min</exception>
```
atifaziz: This should be folded into the above, as in:

```c#
        /// <exception cref="ArgumentOutOfRangeException">min is negative or max is less than min</exception>
```

Also, `min` and `max` should be `paramref` elements.

MoreLinq/CountMethods.cs
atifaziz: Most of my comments in this review also apply to `AtLeast`, `AtMost` and `Exactly` so would help to speed up things if you could integrate the feedback into those too.

atifaziz:
@leandromoh I'm doing some planning for the next release and I'd like to see this go out as part of [MoreLINQ 2.1][m21]. At the moment, I've scheduled 2.1 for end of Jan as a self-imposed deadline to target. Is that something you think your agenda permits? If so, I'll assign this issue the [milestone][m21].


[m21]: https://github.com/morelinq/MoreLINQ/milestone/1

leandromoh:
@atifaziz I did the last modifications you commented. So I think it's only missing to adjust the conflicts on project.json & README.md for this PR be include in the next release. 
If there is no more adjust pending, tell me please, so I will resolve the conflict with the current version of them.

MoreLinq.Test/AtMostTest.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Linq;
+using LinqEnumerable = System.Linq.Enumerable;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class AtMostTest
+    {
+        [Test]
+        [ExpectedException(typeof(ArgumentNullException))]
```
atifaziz: Use `Assert.Throws` instead of `[ExpectedException]` and ditto for all other tests testing for exceptions in this PR.

MoreLinq.Test/AtMostTest.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using LinqEnumerable = System.Linq.Enumerable;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class AtMostTest
+    {
+        [Test]
```
atifaziz: Have a single blank line _consistently_ between methods.

MoreLinq.Test/AtMostTest.cs
```diff
@@ -0,0 +1,94 @@
...
+        public void AtMostWithSingleElementHasAtMostOneElement()
+        {
+            Assert.IsTrue(GetSingleElementSequence().AtMost(1));
+        }
+        [Test]
+        public void AtMostWithSingleElementHasAtMostManyElements()
+        {
+            Assert.IsTrue(GetSingleElementSequence().AtMost(2));
+        }
+        private static IEnumerable<int> GetSingleElementSequence()
```
atifaziz: This should be replaced with direct used of `Enumerable.Repeat(42, 1)` or just `new[] { 42 }`. Having a method like `GetSingleElementSequence` is making the test code longer and difficult to read/review for no apparent benefit.

MoreLinq.Test/AtMostTest.cs
```diff
@@ -0,0 +1,94 @@
...
+            return GetSequence().Take(1);
+        }
+
+        [Test]
+        public void AtMostWithManyElementsHasAtMostOneElements()
+        {
+            Assert.IsFalse(GetManyElementSequence().AtMost(1));
+        }
+
+        private static IEnumerable<int> GetManyElementSequence()
```
atifaziz: There's only a single usage of this method so inline it because having a separate method is not making it more readable, and in fact, one has to jump around to build an understanding. Ditto for other places where the same strategy is being used.

MoreLinq.Test/AtMostTest.cs
```diff
@@ -0,0 +1,94 @@
...
+        public void AtMostWithEmptySequenceHasAtMostZeroElements()
+        {
+            Assert.IsTrue(GetEmptySequence().AtMost(0));
+        }
+        [Test]
+        public void AtMostWithEmptySequenceHasAtMostOneElement()
+        {
+            Assert.IsTrue(GetEmptySequence().AtMost(1));
+        }
+        private static IEnumerable<int> GetEmptySequence()
```
atifaziz: Inline use of `LinqEnumerable.Empty<int>()` directly and remove this method.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <returns><c>true</c> if the number of elements in the sequence is greater than
+        /// or equal to the given integer or <c>false</c> otherwise.</returns>
+        /// <example>
+        /// <code>
+        /// var numbers = { 123, 456, 789 };
+        /// var result = numbers.AtLeast(2);
+        /// </code>
+        /// The <c>result</c> variable will contain <c>true</c>.
+        /// </example>
+        public static bool AtLeast<T>(this IEnumerable<T> source, int min)
```
atifaziz: Rename `min` to simply `count`.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            if (min < 0) throw new ArgumentOutOfRangeException("min", "min must not be negative.");
```
atifaziz: Exception messages must be full English sentences. Instead of:

    min must not be negative.

Consider instead:

    Count cannot be negative.

Please revise other exception message in the same light.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <returns><c>true</c> if the number of elements in the sequence is lesser than
+        /// or equal to the given integer or <c>false</c> otherwise.</returns>
+        /// <example>
+        /// <code>
+        /// var numbers = { 123, 456, 789 };
+        /// var result = numbers.AtMost(2);
+        /// </code>
+        /// The <c>result</c> variable will contain <c>false</c>.
+        /// </example>
+        public static bool AtMost<T>(this IEnumerable<T> source, int max)
```
atifaziz: Rename `max` to simply `count`.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <returns><c>true</c> if the number of elements in the sequence is equals
+        /// to the given integer or <c>false</c> otherwise.</returns>
+        /// <example>
+        /// <code>
+        /// var numbers = { 123, 456, 789 };
+        /// var result = numbers.Exactly(3);
+        /// </code>
+        /// The <c>result</c> variable will contain <c>true</c>.
+        /// </example>
+        public static bool Exactly<T>(this IEnumerable<T> source, int length)
```
atifaziz: Rename `length` to `count`.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <example>
+        /// <code>
+        /// var numbers = { 123, 456, 789 };
+        /// var result = numbers.CountBetween(1, 2);
+        /// </code>
+        /// The <c>result</c> variable will contain <c>false</c>.
+        /// </example>
+        public static bool CountBetween<T>(this IEnumerable<T> source, int min, int max)
+        {
+            if (min < 0) throw new ArgumentOutOfRangeException("min", "min must not be negative.");
```
atifaziz: Replace:

    min must not be negative.

with:

    Minimum count cannot be negative.

MoreLinq/CountMethods.cs
```diff
@@ -0,0 +1,159 @@
...
+            if (max < min) throw new ArgumentOutOfRangeException("max", "max must be greater than or equals to min.");
```
atifaziz: Replace:

    max must be greater than or equals to min.

with:

    Maximum count must be greater than or equal to the minimum count.

README.md
```diff
@@ -19,12 +19,15 @@ Operator             | Summary
 `Assert`             | Asserts that all elements of a sequence meet a given condition otherwise throws an object. This method has 2 overloads.
 `AssertCount`        | Asserts that a source sequence contains a given count of elements. This method has 2 overloads.
 `AtLeast`            | Returns true when the number of elements in the given sequence is greater than or equal to the given integer. This method throws an exception if the given integer is negative
+`AtMost`             | Returns true when the number of elements in the given sequence is lesser than or equal to the given integer. This method throws an exception if the given integer is negative
```
atifaziz: Summaries here need to by synchronised with the revised ones.

MoreLinq/CountMethods.cs
leandromoh: done

MoreLinq/CountMethods.cs
leandromoh: done

MoreLinq/CountMethods.cs
leandromoh: even I dont understand this remarks now, so I remove it.

MoreLinq/CountMethods.cs
leandromoh: done

MoreLinq.Test/AtMostTest.cs
leandromoh: done

MoreLinq.Test/AtMostTest.cs
leandromoh: done

MoreLinq.Test/AtMostTest.cs
leandromoh: done

MoreLinq.Test/AtMostTest.cs
leandromoh: done

MoreLinq.Test/AtMostTest.cs
leandromoh: done

MoreLinq/CountMethods.cs
leandromoh: done

MoreLinq/CountMethods.cs
leandromoh: done

MoreLinq/CountMethods.cs
leandromoh: done

MoreLinq/CountMethods.cs
leandromoh: done

MoreLinq/CountMethods.cs
leandromoh: done

MoreLinq/CountMethods.cs
leandromoh: done

README.md
leandromoh: done

leandromoh:
@atifaziz all modifications were done

atifaziz:
Squashed and merge in d39873278b87d753dcb77e0a9e7d8036dfa108d8.

Thanks!

Issue #202, opened 2016-10-14
Sequence to array by index


atifaziz:
Turns a sequence to an array where each element of the sequence takes a position in the array based on the element's index (determined by a function of the element).

``` c#
static partial class MoreEnumerable
{
    public static T[] ToArrayByIndex<T>(this IEnumerable<T> source,
        Func<T, int> indexSelector) =>
        source.ToArrayByIndex(indexSelector, (e, _) => e);

    public static TResult[] ToArrayByIndex<T, TResult>(this IEnumerable<T> source,
        Func<T, int> indexSelector, Func<T, TResult> resultSelector) =>
        source.ToArrayByIndex(indexSelector, (e, _) => resultSelector(e));

    public static TResult[] ToArrayByIndex<T, TResult>(this IEnumerable<T> source,
        Func<T, int> indexSelector, Func<T, int, TResult> resultSelector)
    {
        var indexed = source.Select(e => new KeyValuePair<int, T>(indexSelector(e), e)).ToArray();
        var array = new TResult[indexed.Max(e => e.Key) + 1];
        foreach (var e in indexed)
            array[e.Key] = resultSelector(e.Value, e.Key);
        return array;
    }

    public static T[] ToArrayByIndex<T>(this IEnumerable<T> source, int length,
        Func<T, int> indexSelector) =>
        source.ToArrayByIndex(length, indexSelector, (e, _) => e);

    public static TResult[] ToArrayByIndex<T, TResult>(this IEnumerable<T> source, int length,
        Func<T, int> indexSelector, Func<T, TResult> resultSelector) =>
        source.ToArrayByIndex(length, indexSelector, (e, _) => resultSelector(e));

    public static TResult[] ToArrayByIndex<T, TResult>(this IEnumerable<T> source, int length,
        Func<T, int> indexSelector, Func<T, int, TResult> resultSelector)
    {
        var array = new TResult[length];
        foreach (var e in source)
        {
            var i = indexSelector(e);
            if (i < array.Length)
                array[i] = resultSelector(e, i);
        }
        return array;
    }
}
```

Suggestions for a better name welcome. 😃

fsateler:
What is the use case of this? Why isn't a `Dictionary<int, T>` sufficient? If order is important, a `SortedList<int, T>` can easily wrap the dictionary.

As to name, so far all functions end in `By` when they take a similar parameter, so maybe change to `ToArrayIndexedBy`?

atifaziz:
[...]

PR #204, opened 2016-10-20
ToArrayByIndex creates an array by mapping each element to an index

atifaziz:
PR for #202. Tests pending.


atifaziz:
The remarks sections of the methods say:

> If more than one element maps to the same index then the latter element overwrites the former in the resulting array.

I'm tempted to add overloads that help in determining what to do when two elements map to the same index. It would amount to an aggregate function per index. This would also make the function integrate grouping and aggregation functionality. This could be held of until needed.

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+        /// </remarks>
+
+        public static TResult[] ToArrayByIndex<T, TResult>(this IEnumerable<T> source,
+            Func<T, int> indexSelector, Func<T, int, TResult> resultSelector)
+        {
+            if (source == null) throw new ArgumentNullException("source");
+            if (indexSelector == null) throw new ArgumentNullException("indexSelector");
+            if (resultSelector == null) throw new ArgumentNullException("resultSelector");
+
+            var indexed = source.Select(e => new KeyValuePair<int, T>(indexSelector(e), e)).ToArray();
```
fsateler: I think `ToList` should be preferred here, as `ToArray` may perform extra allocations to resize the buffer into the final array.

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+            if (indexSelector == null) throw new ArgumentNullException("indexSelector");
+            if (resultSelector == null) throw new ArgumentNullException("resultSelector");
+
+            var array = new TResult[length];
+            if (length > 0)
+            {
+                foreach (var e in source)
+                {
+                    var i = indexSelector(e);
+                    if (i < array.Length)
```
fsateler: I think this method should throw if the size is not enough. The source list can be filtered by the user before if it may contain values larger than expected, but this method does not know if it is a fatal condition.

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+        public static TResult[] ToArrayByIndex<T, TResult>(this IEnumerable<T> source, int length,
+            Func<T, int> indexSelector, Func<T, int, TResult> resultSelector)
+        {
+            if (source == null) throw new ArgumentNullException("source");
+            if (length < 0) throw new ArgumentOutOfRangeException("length");
+            if (indexSelector == null) throw new ArgumentNullException("indexSelector");
+            if (resultSelector == null) throw new ArgumentNullException("resultSelector");
+
+            var array = new TResult[length];
+            if (length > 0)
```
fsateler: Why allow `length == 0`? Seems like a not-very-useful corner case, and not allowing it reduces indentation by one level.

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        /// <summary>
+        /// Creates an array of user-specified length from an
+        /// <see cref="IEnumerable{T}"/> where a function is used to determine
+        /// the index at which an element will be placed in the array. The
+        /// elements are projected into the array via an additional function.
+        /// </summary>
+        /// <param name="source">The source sequence for the array.</param>
+        /// <param name="length">The length of the resulting array.</param>
```
fsateler: Should probably describe it must be nonnegative?

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+        /// <param name="indexSelector">
+        /// A function that maps an element to its index.</param>
+        /// <param name="resultSelector">
+        /// A function to project a source element into an element of the
+        /// resulting array.</param>
+        /// <typeparam name="T">
+        /// The type of the element in<paramref name="source"/>.</typeparam>
+        /// <typeparam name="TResult">
+        /// The type of the element in the resulting arrary.</typeparam>
```
fsateler: s/arrary/array/

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+        /// <returns>
+        /// An array that contains the elements from the input sequence. The
+        /// size of the array will be as large as the highest index returned
+        /// by the <paramref name="indexSelector"/> plus 1.
+        /// </returns>
+        /// <remarks>
+        /// This method forces immediate query evaluation. It should not be
+        /// used on infinite sequences. If more than one element maps to the
+        /// same index then the latter element overwrites the former in the
+        /// resulting array.
```
fsateler: I find this wording a bit difficult to follow as a non-native speaker. Maybe something like the following is a bit clearer:

> If more than one element maps to the same index then the last element overwrites all previous elements in the resulting array.

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+        /// A function that maps an element to its index.</param>
+        /// <typeparam name="T">
+        /// The type of the element in<paramref name="source"/>.</typeparam>
+        /// <returns>
+        /// An array that contains the elements from the input sequence. The
+        /// size of the array will be as large as the highest index returned
+        /// by the <paramref name="indexSelector"/> plus 1.
+        /// </returns>
+        /// <remarks>
+        /// This method forces immediate query evaluation. It should not be
```
fsateler: s/should/can/ ?

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+        /// <summary>
+        /// Creates an array from an <see cref="IEnumerable{T}"/> where a
+        /// function is used to determine the index at which an element will
+        /// be placed in the array.
+        /// </summary>
+        /// <param name="source">The source sequence for the array.</param>
+        /// <param name="indexSelector">
+        /// A function that maps an element to its index.</param>
+        /// <typeparam name="T">
+        /// The type of the element in<paramref name="source"/>.</typeparam>
```
fsateler: missing space before `<paramref`

MoreLinq/ToArrayByIndex.cs
fsateler: Also, maybe checking for `< 0` and throwing an exception with a better error message would be more user-friendly. Currently generating a negative number yields a generic "Index was outside the bounds of the array." which may be confusing.

```csharp
if (i < 0)
    throw new IndexOutOfRangeException("A negative index was generated by indexSelector");
```

A similar check could be added to the size-less overload.

fsateler:
> I'm tempted to add overloads that help in determining what to do when two elements map to the same index. It would amount to an aggregate function per index. This would also make the function integrate grouping and aggregation functionality. This could be held of until needed.

I'd say it's trivial to perform a `GroupBy` before:

```csharp
src.GroupBy(i => i.Index).ToArrayByIndex(g => g.Key, g => g.Sum(i => i.Value))
```

fsateler:
Also, I take it you didn't like `ToArrayIndexedBy` ?

MoreLinq/ToArrayByIndex.cs
atifaziz: Docs on LINQ methods are usually silent on behaviour with respect to infinite lists. I just wanted to mention it here in passing. Perhaps it's not even needed because immediate evaluation will always cause issues with infinite sequences.

On _cannot_ vs _should not_. I think _cannot_ implies technical inability. You certainly can call the method, it will compile and it will run. It just won't ever produce a result. So while you technically _can_ you _shouldn't_.

MoreLinq/ToArrayByIndex.cs
atifaziz: > If more than one element maps to the same index then the last element overwrites all previous elements in the resulting array.

This actually reads even more confusing. It says that if one element maps to the same index (as what?) then `a[N]` (_last_ element) will replace _all previous_ elements from `a[0]` to `a[N-1]` where N is the _last_ element. I think speaking positionally is confusing when you have an array so I chose to use _latter_ and _former_ for the index in question.

Does this sounds better?

> If more than one element maps to the same index then the newer elements (in sequential order) overwrites older ones at that index in the resulting array.

MoreLinq/ToArrayByIndex.cs
fsateler: Indeed, neither `ToArray` nor `ToList` mention this, so it's OK as it is.

MoreLinq/ToArrayByIndex.cs
fsateler: That sounds a bit too complicated as well. Maybe the following:

> If more than one element maps to the same index then the value used will be the last one seen in the source sequence.

If you don't find this better than the original, I say let's keep the original.

atifaziz: Unfortunately when I read it back, it says something else. It says the _last in the whole sequence_ and not _last in the sequence that mapped to the same index_. To avoid ambiguity about how to interpret _last_, it's important to re-iterate that it has to be one mapping to the same index.

I think let's go with the original for now. If someone comes back confused then we can revisit.

MoreLinq/ToArrayByIndex.cs
atifaziz: Hmm…I wonder why [`Enumerable.Take`][take] even allows count to be negative.

> If count is less than or equal to zero, source is not enumerated and an empty `IEnumerable<T>` is returned.

Wonder if same spirit should be maintained here although I am tempted to validate the argument.

[take]: https://msdn.microsoft.com/en-us/library/bb503062.aspx

MoreLinq/ToArrayByIndex.cs
atifaziz: I'd ask why not allow it? The length could be dynamically calculated and could perfectly end up being zero. You don't want the caller to cater for that.

fsateler: Well, I suppose it is OK, *if* the source sequence ends up being empty anyway (and thus, this `if` is removed anyway). In general, I do not like APIs that try to guess what the programmer meant. Why does this function assume that I prefer an empty array be returned if I passed a zero length, but a non-empty sequence? It is just as likely that I made I mistake in the previous code, but this method wouldn't tell me.

atifaziz: > Why does this function assume that I prefer an empty array be returned if I passed a zero length, but a non-empty sequence?

I think that goes with the function's policy that you've also criticized, that an out of bounds index just silently drops the element. If we revise that then this `if` can go.

atifaziz:
> Also, I take it you didn't like `ToArrayIndexedBy` ?

That sounds like one is indexing the array by some key whereas `ToArrayByIndex`, I feel, states more clearly that the output will be conversion to an array (`ToArray...`) and this is going to have _by a key that is an index_ (`...ByIndex`).

atifaziz:
@fsateler LGTY?

MoreLinq/ToArrayByIndex.cs
fsateler: This is still missing. Either the check should be removed, or the documentation updated (otherwise one needs to look at the source to know this exception can happen). Alternatively`<exception>` sections could be added and would be nice anyway.

MoreLinq/ToArrayByIndex.cs
```diff
@@ -0,0 +1,253 @@
...
+        public static TResult[] ToArrayByIndex<T, TResult>(this IEnumerable<T> source,
+            Func<T, int> indexSelector, Func<T, int, TResult> resultSelector)
+        {
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (indexSelector == null) throw new ArgumentNullException(nameof(indexSelector));
+            if (resultSelector == null) throw new ArgumentNullException(nameof(resultSelector));
+
+            var lastIndex = -1;
+            var indexed = (List<KeyValuePair<int, T>>) null;
+            List<KeyValuePair<int, T>> Indexed() => indexed ?? (indexed = new List<KeyValuePair<int, T>>());
```
atifaziz: I'd stick with `Indexed` because the accessibility of a method in .NET does not conventionally change its casing (i.e. Pascal) and `Indexed` here is certainly not a variable like `indexed` in the line immediately above. It seems simplest to me to just continue to apply the convention for methods to local functions (which are not functions really but methods as that is how they are hoisted).

> It took me a while to find this Indexed function ;)

I can imagine, especially if you're reading code online and not in an IDE where you can use a go-to-definition-like feature.

atifaziz:
@fsateler We done with this?

fsateler:
Yes, sorry for the delay. The remaining issue is one of style, and as long as there it is consistent I'm OK.

PR #205, opened 2016-10-27
Await: asynchronous evaluation of sequence elements

atifaziz:
This PR is just an idea, an RFC to solicit feedback. I'm not even sure this belongs in MoreLINQ as it may expand the scope of the library to cover async scenarios.

``` c#
var beatles = new[]
{
    "John Lennon",
    "Paul McCartney",
    "George Harrison",
    "Ringo Starr",
};

var http = new HttpClient();
var results = await beatles.SelectAsync(
    async e => await http.SendAsync(new HttpRequestMessage(HttpMethod.Get, "https://en.wikipedia.org/wiki/" + e.Replace(' ', '_'))),
    (e, rsp) => new
    {
        Topic = e,
        rsp.StatusCode,
        rsp.Content.Headers.ContentLength,
        rsp.Content.Headers.ContentType.MediaType,
    });

foreach (var e in results)
    Console.WriteLine(e.ToString());
```

Output is:

```
{ Topic = John Lennon, StatusCode = OK, ContentLength = 516857, MediaType = text/html }
{ Topic = Paul McCartney, StatusCode = OK, ContentLength = 754121, MediaType = text/html }
{ Topic = George Harrison, StatusCode = OK, ContentLength = 591170, MediaType = text/html }
{ Topic = Ringo Starr, StatusCode = OK, ContentLength = 476451, MediaType = text/html }
```


fsateler:
Interesting. Some thoughts, in no particular order

- This method consumes (and executes the tasks of) the entire source sequence. More general scenarios will likely require a maximum number of tasks to be executed at the same time.
- If it is named `SelectAsync`, it should probably mimic `Select` and return just the `TAsync` instead of a `KeyValuePair`.
- The method is not streaming, as it will wait for all tasks to finish before yielding the first result. This is another difference from plain `Select`.

While this is very useful, I'm unsure this is a good fit for MoreLINQ and it should instead be put in a separate library. Processing of async sequences is likely to require a lot more functions and functionality, and I don't know if this will end up polluting the MoreLINQ library with either more dependencies, or reimplementation of functionality available elsewhere[1]. Perhaps putting this into a separate dll (`MoreAsyncLINQ.dll` ?), within this same repository makes sense.

[1] For example, I made my own implementation of `SelectAsync` by building on top of `Rx`:

```csharp
public static IEnumerable<TResult> SelectAsync<TSource, TResult>(this IEnumerable<TSource> src, Func<TSource, Task<TResult>> asyncSelector, int maxConcurrency) {
    return src.Select(s => Observable.FromAsync(() => asyncSelector(s))).Merge(maxConcurrency).ToEnumerable();
}
```

atifaziz:
@fsateler Thanks for all that feedback! The initial implementation was a very poor crack at the problem. The latter commits address all of your points & `SelectAsync` now pretty much looks (signature-wise) and behaves (streams) like `Select`.

The updated usage now looks like this:

```c#
var urls = new[]
{
    "John Lennon",
    "Paul McCartney",
    "George Harrison",
    "Ringo Starr",
};
    
var http = new HttpClient();
var results =
    from e in urls.SelectAsync(async e => new
    { 
        Topic = e,
        Response = await http.SendAsync(new HttpRequestMessage(HttpMethod.Get, "https://en.wikipedia.org/wiki/" + e.Replace(' ', '_')))
    })
    let rsp = e.Response
    select new
    {
        e.Topic,
        rsp.StatusCode,
        rsp.Content.Headers.ContentLength,
        rsp.Content.Headers.ContentType.MediaType,
    };

foreach (var e in results)
    Console.WriteLine(e.ToString());
```

> Processing of async sequences is likely to require a lot more functions and functionality, 

Not sure. Take a look at the code now.

> …will end up polluting the MoreLINQ library with either more dependencies, or reimplementation of functionality available elsewhere. 

That's what I'm afraid of too thus why I consider this PR as an RFC (should have a label for that?) for now.

> Perhaps putting this into a separate dll (MoreAsyncLINQ.dll ?), within this same repository makes sense.

I don't want to create & maintain another whole library for just one method in the _async_ category. Perhaps a `MoreLinq.Sandbox` would be better for all ideas that are cooking and unsupported?

> I made my own implementation of `SelectAsync` by building on top of `Rx`

Thanks for sharing that and I had done something very similar in the past in projects where I already had a dependency on Rx. However, for projects that don't, I felt it may be possible to implement it simply using existing infrastructure from the framework.

fsateler:
> I consider this PR as an RFC (should have a label for that?)

There is already a `discussion` label. Maybe this tag can be used for this too?

> I don't want to create & maintain another whole library for just one method in the async category. Perhaps a `MoreLinq.Sandbox` would be better for all ideas that are cooking and unsupported?

This sounds like a good idea, but orthogonal to what I proposed.

>> I made my own implementation of SelectAsync by building on top of Rx
>
>Thanks for sharing that and I had done something very similar in the past in projects where I already had a dependency on Rx. However, for projects that don't, I felt it may be possible to implement it simply using existing infrastructure from the framework.

It is certainly possible. But, then we get back to the most important issue:

>> …will end up polluting the MoreLINQ library with either more dependencies, or reimplementation of functionality available elsewhere.
>
> That's what I'm afraid of too

.

>>Processing of async sequences is likely to require a lot more functions and functionality,
>
> Not sure. Take a look at the code now.

What I mean is that mixing async code and sequences is usually more complex than just having a `SelectAsync`. Why not implement `WhereAsync`, `SelectManyAsync`, `JoinAsync`, etc too? Another example: the code now no longer preserves order of the source sequence. Should there be a new overload that does?
If there is potential for adding more async-related methods, I suspect the amount of reimplementation of stuff already present in other libraries will increase..

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,128 @@
...
+            Func<T, Task<TResult>> selector)
+        {
+            if (source == null) throw new ArgumentNullException("source");
+            if (maxConcurrency <= 0) throw new ArgumentOutOfRangeException(nameof(maxConcurrency));
+            if (selector == null) throw new ArgumentNullException("selector");
+
+            var queue = new BlockingCollection<object>();
+            using (var _ = source.GetEnumerator())
+            {
+                var item = _;
```
fsateler: Why this indirection?

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,128 @@
...
+            this IEnumerable<T> source,
+            int maxConcurrency,
+            TaskScheduler scheduler,
+            Func<T, Task<TResult>> selector)
+        {
+            if (source == null) throw new ArgumentNullException("source");
+            if (maxConcurrency <= 0) throw new ArgumentOutOfRangeException(nameof(maxConcurrency));
+            if (selector == null) throw new ArgumentNullException("selector");
+
+            var queue = new BlockingCollection<object>();
```
fsateler: Why not use `Tuple<ExceptionDispatchInfo, Task<TResult>>` instead of object? Then, use an approach like the one in Go or Javascript: The first item is null if no error ocurred, the second item is null if an error occurred.

fsateler:
> What I mean is that mixing async code and sequences is usually more complex than just having a SelectAsync. Why not implement WhereAsync, SelectManyAsync, JoinAsync, etc too? Another example: the code now no longer preserves order of the source sequence. Should there be a new overload that does?

For avoidance of doubt, I'm not suggesting you should provide these methods for `SelectAsync` to be useful. The questions are mostly thinking out loud trying to discover if the async-related methods will grow to a sizable number.

MoreLinq/SelectAsync.cs
atifaziz: To stop ReSharper from complaining about potentially accessing a disposed closure. I found this was good enough for now instead of polluting the code with suppression comments.

atifaziz:
> Another example: the code now no longer preserves order of the source sequence. Should there be a new overload that does?

`Select` in LINQ to SQL and [PLINQ does not preserve order](https://msdn.microsoft.com/en-us/library/dd460677(v=vs.110).aspx) either:

> In PLINQ, the goal is to maximize performance while maintaining correctness. A query should run as fast as possible but still produce the correct results. In some cases, correctness requires the order of the source sequence to be preserved; however, ordering can be computationally expensive. Therefore, by default, PLINQ does not preserve the order of the source sequence. In this regard, PLINQ resembles LINQ to SQL, but is unlike LINQ to Objects, which does preserve ordering.

I think it's understood that there needs to be a balancing act here. If streaming is the greater benefit then order can't guaranteed and if order is important then you need to follow-up with an `OrderBy`.

I have the same concerns as you about the proliferation of other async-enabled operators, which is why I continue to be in two minds about adding this to MoreLINQ. I reckon though that `SelectAsync` will be the most popular (the primary motivation being use of async lambdas in existing LINQ queries) & the rest may just fall under YAGNI. If anyone is doing anything more sophisticated then they'll probably be using Rx in the first place. Instead of guessing & worrying, I think the right answer to put all these fears to rest may be to have another library for hosting experiments, be that `MoreLinq.Sandbox` or `MoreLinq.Labs`. And it doesn't have to have tests or rigorous review.

atifaziz:
Some open points:

- What happens to tasks that are in-flight when the iterator is closed or an error is thrown by one of the asynchronous operations?
- How would one support cancellation for the entire enumeration? The cancellation token would have to be re-created on each enumeration for the query to have repeatable semantics. Would cancellation be ever needed?

atifaziz:
The [`IDisposable.Dispose`](https://msdn.microsoft.com/en-us/library/system.idisposable.dispose.aspx) docs clearly state that subsequent calls should be ignored:

> If an object's Dispose method is called more than once, the object must ignore all calls after the first one. 

So it is safe (or not incorrect) to assume so.

atifaziz:
> What happens to tasks that are in-flight when the iterator is closed or an error is thrown by one of the asynchronous operations?

There are now new overload where the projection function receives an additional argument that's a `CancellationToken`. The cancellation token is signaled when either the iteration is stopped early (e.g. `SelectAsync` is combined with `Take` or `TakeWhile`) or the project of an element throws an error, thus allowing in-flight tasks to cancel.

Below is an example using an overload that send a `CancellationToken` to the projection function.

```c#
var beatles = new[]
{
    "John Lennon",
    "Paul McCartney",
    "George Harrison",
    "Ringo Starr",
};
    
var http = new HttpClient();
var results =
    from e in beatles.SelectAsync(async (e, ct) =>
    {
        try
        {
            var url = "https://en.wikipedia.org/wiki/" + e.Replace(' ', '_');
            return new
            { 
                Topic = e,
                Response = await http.SendAsync(new HttpRequestMessage(HttpMethod.Get, url), ct),
            };
        }
        catch (OperationCanceledException)
        {
            Console.WriteLine($"ABORT! {e}");
            throw;
        }
    })
    let rsp = e.Response
    select new
    {
        e.Topic,
        rsp.StatusCode,
        rsp.Content.Headers.ContentLength,
        rsp.Content.Headers.ContentType.MediaType,
    };

foreach (var e in results.Take(2))
    Console.WriteLine(e.ToString());
```

Because we only care about 2 results (note the `Take(2)` in the `foreach` loop at the end), one possible output is:

```
{ Topic = John Lennon, StatusCode = OK, ContentLength = 518001, MediaType = text/html }
{ Topic = George Harrison, StatusCode = OK, ContentLength = 591180, MediaType = text/html }
ABORT! Ringo Starr
ABORT! Paul McCartney
```

Since `SelectAsync` does not respect the source order, the output above will vary with each run.

atifaziz:
> Another example: the code now no longer preserves order of the source sequence. Should there be a new overload that does?

There is now an option to preserve the source sequence order, which is off by default. Instead of adding loads of overloads with all combinations of parameters (ordering of which is another cognitive burden), I've refactored the API into just two simple and clean overloads focused on projection; all other options are expressed as further chains. Here's an example that demonstrates the new shape and usage of the API:

```c#
Enumerable
    .Range(0, 10)
    .RandomSubset(10)
    .Pipe(x => Console.WriteLine($"{DateTime.Now.TimeOfDay} => {x}"))
    .SelectAsync(async n => { await Task.Delay(TimeSpan.FromSeconds(n)); return n; })
     // .MaxConcurrency(3)
        .PreserveOrder()
    .ForEach(x => Console.WriteLine($"{DateTime.Now.TimeOfDay} <= {x}"));
```

Notice how `MaxConcurrency`, `PreserveOrder` and `Scheduler` just compose as options. This does introduce `ISelectAsyncEnumerable` as a new abstraction, but hey, this is experimental work after all. 😇 

Here's the output from one run:

```
18:22:00.2042515 => 0
18:22:00.2042515 => 4
18:22:00.2042515 => 8
18:22:00.2042515 => 3
18:22:00.2042515 => 5
18:22:00.2042515 => 9
18:22:00.2042515 => 6
18:22:00.2042515 => 2
18:22:00.2042515 => 7
18:22:00.2042515 => 1
18:22:00.2042515 <= 0
18:22:04.2117466 <= 4
18:22:08.2131816 <= 8
18:22:08.2131816 <= 3
18:22:09.2126904 <= 5
18:22:09.2131736 <= 9
18:22:09.2131736 <= 6
18:22:09.2131736 <= 2
18:22:09.2131736 <= 7
18:22:09.2131736 <= 1
```

atifaziz:
While `SelectAsync` is a natural name, [it is also popular](https://github.com/search?l=C%23&q=linq+SelectAsync&type=Code&utf8=%E2%9C%93) and could lead to conflicts easily. I am inclined to change it to `MapAsync`. 🤔

atifaziz:
:memo: `WhereAsync` can be built using or on top of `SelectAsync`:

```c#
Enumerable
    .Range(0, 10)
    .SelectAsync(async n => (Item: n, Matched: await Task.FromResult(n % 2 == 0)))
    .AsOrdered()
    .Where(e => e.Matched)
    .Select(e => e.Item)
    .ForEach(x => Console.WriteLine(x));
```

MoreLinq/Experimental/SelectAsync.cs
```diff
@@ -0,0 +1,519 @@
...
+                                   TaskScheduler.Default,
+                                   preserveOrder: false);
+
+        /// <summary>
+        /// Gets a positive (non-zero) integer that specifies the maximum
+        /// projections to run concurrenctly or <c>null</c> to mean unlimited
+        /// concurrency.
+        /// </summary>
+
+        public int? MaxConcurrency { get; }
```
fsateler: Why not non-nullable int? 0 is not allowed anyway.

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,439 @@
...
+                                   TaskScheduler.Default,
+                                   preserveOrder: false);
+
+        /// <summary>
+        /// Gets a positive (non-zero) integer that specifies the maximum
+        /// projections to run concurrenctly or <c>null</c> to mean unlimited
+        /// concurrency.
+        /// </summary>
+
+        public int? MaxConcurrency { get; }
```
fsateler: Why not `int`? Zero is not allowed anyway.

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,439 @@
...
+    using System.Linq;
+    using System.Runtime.ExceptionServices;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Represents options for an asynchronous projection operation.
+    /// </summary>
+
+    public sealed class SelectAsyncOptions
```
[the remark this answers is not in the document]
atifaziz: If equality is implemented, does the choice matter then? For non-changing options (each `With*` method returns `this` if nothing is changing), the reference optimises away copying that comes with value types.

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,439 @@
...
+            {
+                if (options == null) throw new ArgumentNullException(nameof(options));
+                return Options == options ? this : new SelectAsyncEnumerable<T>(_impl, options);
+            }
+
+            public IEnumerator<T> GetEnumerator() => _impl(Options).GetEnumerator();
+            IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+        }
+
+        sealed class DelegatingComparer<T> : IComparer<T>
```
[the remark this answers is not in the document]
atifaziz: I must have missed the news that they added it in .NET Framework 4.5!

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,439 @@
...
+                var completed = false;
+
+                async Task<KeyValuePair<int, TResult>> Select(KeyValuePair<int, T> input, CancellationToken cancellationToken) =>
+                    new KeyValuePair<int, TResult>(input.Key, await selector(input.Value, cancellationToken).ConfigureAwait(false));
+
+                var item = source.Index().GetEnumerator();
+                IDisposable disposable = item; // disables AccessToDisposedClosure warnings
+                try
+                {
+                    Task.Factory.StartNew(async () =>
```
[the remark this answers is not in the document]
atifaziz: Fair point.

Local methods can break the flow when reading top-down, forcing more on the cognitive stack. Moving into its own method means moving all (and the many) free variables into parameters, which when taken out of context may not read easily. The method has become large by supporting ordered results. I can give it a shot and see what it gives but I reckon that different tactics may yields better results.

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,439 @@
...
+                        else
+                        {
+                            // Received a result out of order when order must be
+                            // preserved, so withhold the result by finding out
+                            // where it belongs in the order of results withheld
+                            // so far and insert it in the list.
+
+                            var i = holds.BinarySearch(r, KeyValueComparer<int, TResult>.Default);
+                            Debug.Assert(i < 0);
+                            holds.Insert(~i, r);
```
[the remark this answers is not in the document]
atifaziz: > Lists are pretty bad at insertion in the middle.

Yeah, I'm well aware but this is a first cut to focus on getting the API shape and runtime behaviour right.

> This could be postponed until List has been shown to be a problem

Absolutely! There should be room left for improvement. 😇

MoreLinq/Experimental/SelectAsync.cs
atifaziz: I wanted to start with a nullable integer and see how it looks and feel. I agree that it's not any clearer to anyone what `null` would mean here besides undefined and that undefined means unbounded. Both 0 and `null` would need that to be spelled out in the documentation.

MoreLinq/SelectAsync.cs
atifaziz: Because it can be undefined or unspecified, which means unbounded concurrency.

atifaziz:
> but I didn't want to be silent on this for so long. 

Hey, I appreciate any slice of time you can offer and glad you dropped by. I didn't want to pester you with more review items if you didn't have the time. Plus, since this is experimental, I expect some oversight to slip by. The earlier we can catch them, though, the better!

> I haven't fully understood the meat of the implementation

Happy to explain any mysteries though if it's mysterious then I've done a poor job.

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,439 @@
...
+                                        // Cancellation during the wait means
+                                        // the enumeration has been stopped by
+                                        // the user so the results of the
+                                        // remaining tasks are no longer needed.
+                                        // Those tasks should cancel as a result
+                                        // of sharing the same cancellation
+                                        // token and provided that they passed
+                                        // it on to any downstream asynchronous
+                                        // operations. Either way, this loop
+                                        // is done so exit hard here.
```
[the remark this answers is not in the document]
atifaziz: The user can stop the enumeration by disposing the enumerator over the enumerable we return. The cancellation is caused indirectly. [See my earlier comment about it and the example with `Take`](https://github.com/morelinq/MoreLINQ/pull/205#issuecomment-256885850).

MoreLinq/SelectAsync.cs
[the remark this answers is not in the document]
atifaziz: > This class appears to match value semantics

Do you say that on the sole basis that it is immutable? Building value semantics such as equality, while possible, may be a bit meaningless when `Scheduler` is a reference. In fact, I think `Scheduler` should even be a scheduler factory so that something like [`TaskScheduler.FromCurrentSynchronizationContext()`](https://docs.microsoft.com/en-gb/dotnet/api/system.threading.tasks.taskscheduler.fromcurrentsynchronizationcontext) is evaluated at this time of evaluation rather than at the time the query/enumerable/computation is setup.

MoreLinq/SelectAsync.cs
```diff
@@ -0,0 +1,439 @@
...
+        /// <para>
+        /// This method starts a new task where the asynchronous projections
+        /// are started and awaited.</para>
+        /// <para>
+        /// The <paramref name="selector"/> function should be designed to be
+        /// thread-agnostic.</para>
+        /// </remarks>
+
+        public static ISelectAsyncEnumerable<TResult> SelectAsync<T, TResult>(
+            this IEnumerable<T> source, Func<T, CancellationToken, Task<TResult>> selector)
```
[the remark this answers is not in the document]
atifaziz: We have to be careful about passing “live” items into a query. The query is a declaration of a computation that is then evaluated, possible at some point in the future. The `CancellationToken` can introduce side-effects that breaks subsequent evaluations.

`SelectAsync`, which I am thinking about renaming, is very misleading as a name. It's not async because it returns a sequence. It's the projection that's async. Someone who is async-heavy won't be making use of it because it can cause the thread calling `SelectAsync` to block. `SelectAsync` is good in synchronous code that requires I/O or asynchronous operation for the projections.

atifaziz:
I want to call out [my earlier comment](https://github.com/morelinq/MoreLINQ/pull/205#discussion_r171705317) here into the main thread:

> `SelectAsync`, which I am thinking about renaming, is very misleading as a name. It's not async because it returns a sequence. It's the projection that's async. Someone who is async-heavy won't be making use of it because it can cause the thread calling `SelectAsync` to block. `SelectAsync` is good in synchronous code that requires I/O or asynchronous operation for the projections.

A true `SelectAsync` would return a `Task<IEnumerable<T>>`, so the name is misleading because the return type is still just a sequence: `IEnumerable<TResult>`. I am tempted to call it `Fetch`, instead, to convey that it is an operator for merely _fetching_ results, albeit asynchronously.

Thoughts?

MoreLinq/SelectAsync.cs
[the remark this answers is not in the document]
atifaziz: See how you feel about 33e437aa52582c11c497d041f1217332ed43401d. One bit of bad news is boxing, which results from being generic about the extract method and `BlockingCollection<>` generic type argument being `object`. I could offset that by introducing a discriminated union notification `struct` ([à la F#](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/discriminated-unions#struct-discriminated-unions)) but it seems like an overkill at this stage? Improvement for the future? The clarity (brought by the extraction) has come at some cost. Either we revert and improve by refactoring later or go forward and improve by optimising the boxing later. Both ways, speaking objectively, we end up at the same place. This is about internals, so like use of `List<>` not being the best data structure for insertion, this is something we should revisit later and focus instead on the shape of the API and the runtime behaviour for a first (experimental) release. What say?

MoreLinq/Experimental/SelectAsync.cs
atifaziz: > I agree that it's not any clearer to anyone what `null` would mean here

On second thoughts, `MaxConcurrency(0)` reads just worse than `MaxConcurrency(null)`. With the former, it's an invalid value overloaded to be [magical](https://en.wikipedia.org/wiki/Magic_number_(programming)) and one can only take a wild guess without looking up the documentation. With `null`, an undefined maximum, could really only mean one thing for concurrency: no maximum, therefore unbounded. The null here means an _optional maximum_ and type of `int?` communicates that better. I'm inclined to stay with this design.

MoreLinq/SelectAsync.cs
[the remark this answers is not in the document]
atifaziz: In 90acfbeff8f42729894da7c84a85447a96a76a88, I'm [using a tuple like you suggested some time ago](https://github.com/morelinq/MoreLINQ/pull/205#discussion_r85404147), except it's a value tuple so no heap allocation. The boxing concern is now also mute and there are fewer chances to make errors with casting decision (since we don't have tests) as the storage class for notifications is no longer `object`. I'm happy with the all the positive impacts resulting from the method extraction. Thanks for the challenge :+1: and hope you as well like where we've ended.

MoreLinq/Experimental/SelectAsync.cs
```diff
@@ -0,0 +1,548 @@
...
+
+                    var nextKey = 0;
+                    var holds = ordered ? new List<(int, TResult)>() : null;
+
+                    foreach (var (kind, result, error) in notices.GetConsumingEnumerable())
+                    {
+                        if (kind == Notice.Error)
+                            error.Throw();
+
+                        if (kind == Notice.End)
```
fsateler: What is the purpose of this `Notice.End` enum value? AFAICT, if this was missing, we would just the loop one iteration earlier. I feel like I'm missing something

MoreLinq/SelectAsync.cs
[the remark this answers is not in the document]
fsateler: Why not use [SortedList&lt;T>](https://msdn.microsoft.com/en-us/library/ms132319.aspx) ? It seems to fit the purposes quite well, and it should allow removing some book-keeping code.

MoreLinq/Experimental/SelectAsync.cs
fsateler: Given there is already a `AsSequential`, maybe there should be `MaxConcurrency(int)` to set a limit, and `UnboundedConcurrency` to set it to null.

MoreLinq/Experimental/SelectAsync.cs
```diff
@@ -0,0 +1,548 @@
...
+                    if (more && (more = e.MoveNext()))
+                        tasks.Add(taskSelector(e.Current).Select(r => (e.Current, r)));
+                }
+
+                collection.Add(endNotice);
+            }
+            catch (Exception ex)
+            {
+                cancellationTokenSource.Cancel();
+                collection.Add(errorNoticeSelector(ex));
```
fsateler: We need to dispose `e` here.

MoreLinq/Experimental/SelectAsync.cs
```diff
@@ -0,0 +1,548 @@
...
+
+                        return;
+                    }
+
+                    var task = (Task<(T Input, TResult Result)>) completedTask;
+                    tasks.Remove(task);
+                    collection.Add(resultNoticeSelector(task.Result.Input, task.Result.Result));
+
+                    if (more && (more = e.MoveNext()))
+                        tasks.Add(taskSelector(e.Current).Select(r => (e.Current, r)));
```
fsateler: Disposing `e` as soon as `more == false` would be nice.

MoreLinq/Experimental/SelectAsync.cs
```diff
@@ -0,0 +1,548 @@
...
+                return Options == options ? this : new SelectAsyncEnumerable<T>(_impl, options);
+            }
+
+            public IEnumerator<T> GetEnumerator() => _impl(Options).GetEnumerator();
+            IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+        }
+
+        static class TupleComparer<T1, T2>
+        {
+            public static readonly IComparer<(T1, T2)> Key1 =
```
fsateler: I would name these `Item1` and `Item2` since `ValueTuple` does not have keys.

MoreLinq/Experimental/SelectAsync.cs
```diff
@@ -0,0 +1,548 @@
...
+
+        public static readonly SelectAsyncOptions Default =
+            new SelectAsyncOptions(null /* = unbounded concurrency */,
+                                   TaskScheduler.Default,
+                                   preserveOrder: false);
+
+        /// <summary>
+        /// Gets a positive (non-zero) integer that specifies the maximum
+        /// projections to run concurrenctly or <c>null</c> to mean unlimited
+        /// concurrency.
```
fsateler: This is not strictly true. `null` means `int.MaxValue` concurrency. I suppose one could treat `int.MaxValue` as an approximation of unlimited though 😉

MoreLinq/Experimental/SelectAsync.cs
```diff
@@ -0,0 +1,548 @@
...
+                        cancellationTokenSource.Cancel();
+                    disposable.Dispose();
+                }
+            }
+        }
+
+        enum Notice { Result, Error, End }
+
+        static async Task<TResult> Select<T, TResult>(this Task<T> task, Func<T, TResult> selector) =>
+            selector(await task);
```
fsateler: I'm not an expert in async code, but I think it is best practice to always use `ConfigureAwait(false)` in library code.

MoreLinq/Experimental/SelectAsync.cs
atifaziz: This is designed for sorting, which is why I called them `Key1` and `Key2` but I guess if you found it confusing then perhaps best to keep parity. I'll rename.

MoreLinq/Experimental/SelectAsync.cs
atifaziz: Admittedly, the documentation lies. 😱 The implementation takes a shortcut that can be revised with time if someone hits the limit of 2 billion outstanding tasks. 🚀

atifaziz:
> but I haven't been able to come up with better alternatives. `CollectTasks`, `Wait`, `Await` don't sound any better to me.

`SelectTask` or `SelectAwaitable` also come to mind but I fear they won't bring much more clarity. Is `Fetch` so bad? Do we need to really make anything about tasks or awaiting obvious in the name?

> BTW, since I never explicitly said this before: I like the overall shape, and I think this can be a useful addition to MoreLINQ.

Cool and I'd like to work towards getting moving it out of the experimental space. If we can settle on a name and release it as experimental then there'll be time to work out unit tests and iron out any kinks from battle-testing.

atifaziz:
> `CollectTasks`, `Wait`, `Await` don't sound any better to me.

I realised that it is a bit useless to have a projection function in the first overload. The method can be reduced to simply being an extension of a sequence of tasks (`IEnumerable<Task<T>>`). See what it gives in 3819d3d44546eb8f217c3044f03ce8fa3629483f. What's also interesting is that it then made sense to simply call the method `Await`, as one of your suggestions! This is either a good thing or it means that the first overload should be removed because no one in their right mind will use it; it'll waste work if the sequence is not full consumed. The second overload makes you think about cancellation.

atifaziz:
With 09497dc85854569ad9f510ec3bdda5761741ace2, I'm proposing to rename `SelectAsync` to `Await` entirely. With some careful re-wording, I think it brings about more clarity. That is, `SelectAsync` was possibly the wrong name all along because the function in following overload _felt_ like a projection when it's really an _evaluation_.

```c#
public static IAwaitQuery<TResult> Await<T, TResult>(
    this IEnumerable<T> source,
    Func<T, CancellationToken, Task<TResult>> evaluator)
```

It's purpose is not as much to project (_T_ &rarr; _TResult_) as it is to simply supply or _inject_ the `CancellationToken` into the evaluation in order to abort it if the iteration is terminated prematurely.

@fsateler What do you think? Am I twisting things to force retrofitting a definition that favours `Await` as a name or does it also make sense to you, and clarify/scope things better?

atifaziz:
@fsateler Did I miss anything from the [changes you requested in your last review](https://github.com/morelinq/MoreLINQ/pull/205#pullrequestreview-101557101)? Wondering what's keeping to ship this as an experiment?

Issue #104, opened 2015-08-21
F# countBy analogue


atifaziz:
Similar to the group by, but in the result values are integers, instead of sequences

https://msdn.microsoft.com/en-us/library/ee340317.aspx

---

Originally reported on Google Code with ID 104

Reported by @marhoily on 2015-06-23 14:21:00

R2-Miao:
Hello @atifaziz, I would like to try my hand at implementing this.

atifaziz:
@R2-Miao By all means. Thanks for volunteering. Please fork, branch and send a PR. We'll take it from there.

atifaziz:
X-ref PR #146

PR #207, opened 2016-10-30
Implemented CountBy (Issue 104)

leandromoh:
Implemented the F# countBy analogue (#104)

Currently missing to add method reference in "project.json" and "README.md" files.
I pretend to do it latter to avoid conflicts.

### DIFF
--- a/MoreLinq.Test/CountByTest.cs
+++ b/MoreLinq.Test/CountByTest.cs
@@ -88,5 +88,16 @@ namespace MoreLinq.Test
 
             result.AssertSequenceEqual(expectations);
         }
+        
+        [Test]
+        public void CountByHasKeysOrderedLikeGroupBy()
+        {
+            var randomSequence = MoreEnumerable.Random(0, 100).Take(100).ToArray();
+
+            var countByKeys = randomSequence.CountBy(x => x).Select(x => x.Key);
+            var groupByKeys = randomSequence.GroupBy(x => x).Select(x => x.Key);
+
+            countByKeys.AssertSequenceEqual(groupByKeys);
+        }
     }
 }
--- a/MoreLinq.Test/CountByTest.cs
+++ b/MoreLinq.Test/CountByTest.cs
@@ -1,4 +1,4 @@
-﻿#region License and Terms
+#region License and Terms
 // MoreLINQ - Extensions to LINQ to Objects
 // Copyright (c) 2016 Leandro F. Vieira (leandromoh). All rights reserved.
 // 
@@ -45,31 +45,48 @@ namespace MoreLinq.Test
         [Test]
         public void CountBySimpleTest()
         {
-            IEnumerable<KeyValuePair<int, int>> result = new[] { 1, 2, 3, 4, 5, 6, 1, 2, 3, 1, 1, 2 }.CountBy(c => c);
+            var result = new[] { 1, 2, 3, 4, 5, 6, 1, 2, 3, 1, 1, 2 }.CountBy(c => c);
 
-            IEnumerable<KeyValuePair<int, int>> expecteds = new Dictionary<int, int>() { { 1, 4 }, { 2, 3 }, { 3, 2 }, { 4, 1 }, { 5, 1 }, { 6, 1 } };
+            var expectations = new List<KeyValuePair<int, int>>()
+            {
+                { 1, 4 },
+                { 2, 3 },
+                { 3, 2 },
+                { 4, 1 },
+                { 5, 1 },
+                { 6, 1 },
+            };
 
-            result.AssertSequenceEqual(expecteds);
+            result.AssertSequenceEqual(expectations);
         }
 
         [Test]
         public void CountByEvenOddTest()
         {
-            IEnumerable<KeyValuePair<int, int>> result = Enumerable.Range(1, 100).CountBy(c => c % 2);
+            var result = Enumerable.Range(1, 100).CountBy(c => c % 2);
 
-            IEnumerable<KeyValuePair<int, int>> expecteds = new Dictionary<int, int>() { { 1, 50 }, { 0, 50 } };
+            var expectations = new List<KeyValuePair<int, int>>()
+            {
+                { 1, 50 },
+                { 0, 50 },
+            };
 
-            result.AssertSequenceEqual(expecteds);
+            result.AssertSequenceEqual(expectations);
         }
 
         [Test]
         public void CountByWithEqualityComparer()
         {
-            IEnumerable<KeyValuePair<string, int>> result = new[] { "a", "B", "c", "A", "b", "A" }.CountBy(c => c, StringComparer.OrdinalIgnoreCase);
+            var result = new[] { "a", "B", "c", "A", "b", "A" }.CountBy(c => c, StringComparer.OrdinalIgnoreCase);
 
-            IEnumerable<KeyValuePair<string, int>> expecteds = new Dictionary<string, int>() { { "a", 3 }, { "B", 2 }, { "c", 1 } };
+            var expectations = new List<KeyValuePair<string, int>>()
+            {
+                { "a", 3 },
+                { "B", 2 },
+                { "c", 1 },
+            };
 
-            result.AssertSequenceEqual(expecteds);
+            result.AssertSequenceEqual(expectations);
         }
     }
 }
--- a/MoreLinq.Test/CountByTest.cs
+++ b/MoreLinq.Test/CountByTest.cs
@@ -1,0 +1,75 @@
+﻿#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2016 Leandro F. Vieira (leandromoh). All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using LinqEnumerable = System.Linq.Enumerable;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class CountByTest
+    {
+        [Test]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void CountByWithNullSequence()
+        {
+            IEnumerable<int> sequence = null;
+            sequence.CountBy(x => x % 2 == 0);
+        }
+
+        [Test]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void CountByWithNullProjection()
+        {
+            Func<int, bool> projection = null;
+            Enumerable.Range(1, 10).CountBy(projection);
+        }
+
+        [Test]
+        public void CountBySimpleTest()
+        {
+            IEnumerable<KeyValuePair<int, int>> result = new[] { 1, 2, 3, 4, 5, 6, 1, 2, 3, 1, 1, 2 }.CountBy(c => c);
+
+            IEnumerable<KeyValuePair<int, int>> expecteds = new Dictionary<int, int>() { { 1, 4 }, { 2, 3 }, { 3, 2 }, { 4, 1 }, { 5, 1 }, { 6, 1 } };
+
+            result.AssertSequenceEqual(expecteds);
+        }
+
+        [Test]
+        public void CountByEvenOddTest()
+        {
+            IEnumerable<KeyValuePair<int, int>> result = Enumerable.Range(1, 100).CountBy(c => c % 2);
+
+            IEnumerable<KeyValuePair<int, int>> expecteds = new Dictionary<int, int>() { { 1, 50 }, { 0, 50 } };
+
+            result.AssertSequenceEqual(expecteds);
+        }
+
+        [Test]
+        public void CountByWithEqualityComparer()
+        {
+            IEnumerable<KeyValuePair<string, int>> result = new[] { "a", "B", "c", "A", "b", "A" }.CountBy(c => c, StringComparer.OrdinalIgnoreCase);
+
+            IEnumerable<KeyValuePair<string, int>> expecteds = new Dictionary<string, int>() { { "a", 3 }, { "B", 2 }, { "c", 1 } };
+
+            result.AssertSequenceEqual(expecteds);
+        }
+    }
+}
--- a/MoreLinq.Test/CountByTest.cs
+++ b/MoreLinq.Test/CountByTest.cs
@@ -99,5 +99,11 @@ namespace MoreLinq.Test
 
             countByKeys.AssertSequenceEqual(groupByKeys);
         }
+        
+        [Test]
+        public void CountByIsLazy()
+        {
+            new BreakingSequence<string>().CountBy(x => x.Length);
+        }
     }
 }
--- a/MoreLinq.Test/TestExtensions.cs
+++ b/MoreLinq.Test/TestExtensions.cs
@@ -73,5 +73,9 @@ namespace MoreLinq.Test
                 yield return split;
         }
 
+        internal static void Add<TKey, TValue>(this List<KeyValuePair<TKey, TValue>> list, TKey key, TValue value)
+        {
+            list.Add(new KeyValuePair<TKey, TValue>(key, value));
+        }
     }
 }
--- a/MoreLinq.Test/TestExtensions.cs
+++ b/MoreLinq.Test/TestExtensions.cs
@@ -73,7 +73,7 @@ namespace MoreLinq.Test
                 yield return split;
         }
 
-        internal static void Add<TKey, TValue>(this List<KeyValuePair<TKey, TValue>> list, TKey key, TValue value)
+        internal static void Add<TKey, TValue>(this IList<KeyValuePair<TKey, TValue>> list, TKey key, TValue value)
         {
             list.Add(new KeyValuePair<TKey, TValue>(key, value));
         }
--- a/MoreLinq/CountBy.cs
+++ b/MoreLinq/CountBy.cs
@@ -58,11 +58,37 @@ namespace MoreLinq
             return CountByImpl(source, keySelector, comparer);
         }
 
-        private static IEnumerable<KeyValuePair<TKey, int>> CountByImpl<TSource, TKey>(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey> comparer)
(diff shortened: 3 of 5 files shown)


MoreLinq/CountBy.cs
```diff
@@ -0,0 +1,79 @@
...
+    using System;
+    using System.Collections.Generic;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Applies a key-generating function to each element of a sequence and returns a sequence of 
+        /// unique keys and their number of occurrences in the original sequence.
+        /// </summary>
+        /// <typeparam name="TSource">Type of the source sequence</typeparam>
```
atifaziz: It's not the type of the source sequence but the type of _the elements of the source sequence_. Ditto for the other overload.

Also, all XML summary comment are full sentences so consider terminating them with a period or full stop (`.`).

MoreLinq/CountBy.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <param name="keySelector">Function that transforms each item of source sequence into a key to be compared against the others</param>
+        /// <returns>A sequence of unique keys and their number of occurrences in the original sequence</returns>
+        public static IEnumerable<KeyValuePair<TKey, int>> CountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector)
+        {
+            return source.CountBy(keySelector, null);
+        }
+
+        /// <summary>
+        /// Applies a key-generating function to each element of a sequence and returns a sequence of 
+        /// unique keys and their number of occurrences in the original sequence.
```
atifaziz: This does not mention how this overload is any different that the other & therefore could look odd on a [page of summaries](http://morelinq.github.io/1.x/ref/api/html/T_MoreLinq_MoreEnumerable.htm). It should mention the additional argument. Considering adding:

> An additional argument specifies a comparer to use for testing equivalence of keys.

MoreLinq/CountBy.cs
```diff
@@ -0,0 +1,79 @@
...
+            return CountByImpl(source, keySelector, comparer);
+        }
+
+        private static IEnumerable<KeyValuePair<TKey, int>> CountByImpl<TSource, TKey>(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey> comparer)
+        {
+            var dic = new Dictionary<TKey, int>(comparer);
+
+            foreach (var item in source)
+            {
+                TKey key = keySelector(item);
```
atifaziz: Use `var` instead `TKey`.

MoreLinq/CountBy.cs
```diff
@@ -0,0 +1,79 @@
...
+
+                if (dic.ContainsKey(key))
```
atifaziz: Consider collapsing this `if` into the simpler form:

```c#
dic[key] = dic.Contains(key) ? dic[key] + 1 : 1;
```

MoreLinq.Test/CountByTest.cs
```diff
@@ -0,0 +1,75 @@
...
+        public void CountByWithNullProjection()
+        {
+            Func<int, bool> projection = null;
+            Enumerable.Range(1, 10).CountBy(projection);
+        }
+
+        [Test]
+        public void CountBySimpleTest()
+        {
+            IEnumerable<KeyValuePair<int, int>> result = new[] { 1, 2, 3, 4, 5, 6, 1, 2, 3, 1, 1, 2 }.CountBy(c => c);
```
atifaziz: Use `var` here and elsewhere instead of repeating the full type.

MoreLinq.Test/CountByTest.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            IEnumerable<KeyValuePair<int, int>> expecteds = new Dictionary<int, int>() { { 1, 4 }, { 2, 3 }, { 3, 2 }, { 4, 1 }, { 5, 1 }, { 6, 1 } };
```
atifaziz: Replace `expecteds` with `expectations` because I think that's what you mean. Also, I would consider laying out the dictionary vertically for readability:

```c#
var expecations = new Dictionary<int, int>
{
    [1] = 4,
    [2] = 3,
    [3] = 2,
    [4] = 1,
    [5] = 1,
    [6] = 1,
};
```

MoreLinq.Test/CountByTest.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            result.AssertSequenceEqual(expecteds);
```
atifaziz: Using `AssertSequenceEqual` with dictionaries is dangerous. `AssertSequenceEqual` tests sequences & dictionaries do not guarantee enumeration in any particular or logical order. Here you are comparing a sequence (as _logically_ returned by `CountBy`) with a dictionary. If the test is passing, it's really by fluke & therefore a false positive. If you want to use dictionaries, you must sort entries by key before using `AssertSequenceEqual`, e.g.:

```c#
results.OrderBy(e => e.Key).AssertSequenceEqual(expecteds).OrderBy(e => e.Key);
```

If you do the above, a comment in the code may be worthwhile.

atifaziz:
It would be nice to attempt a version that, like `GroupBy`, maintains the order of the keys with respect to the source sequence. If it turns out too complicated then we can revert but it seems worth a try.

leandromoh:
@atifaziz Thanks for the review! All adjusts were done.

MoreLinq/CountBy.cs
```diff
@@ -0,0 +1,83 @@
...
+                    dic[key]++;
+                }
+                else
+                {
+                    dic[key] = 1;
+                    keys.Add(key);
+                }
+            }
+
+            return keys.Select(k => new KeyValuePair<TKey, int>(k, dic[k]));
```
atifaziz: Can we add a test case to specifically test that keys are indeed returned in the order they were seen in the source? If we return `dic` directly, we should be able to see the test fail.

atifaziz: Can we add a test case to specifically test that keys are indeed returned in the order they were seen in the source? If we return `dic` directly, we should be able to see the test fail.

MoreLinq.Test/TestExtensions.cs
```diff
@@ -73,5 +73,9 @@ internal static IEnumerable<string> GenerateSplits(this string str, params char[
                 yield return split;
         }
 
+        internal static void Add<TKey, TValue>(this List<KeyValuePair<TKey, TValue>> list, TKey key, TValue value)
```
atifaziz: If we are going to add this, might as well extend `IList<KeyValuePair<TKey, TValue>>` (_interface_) as opposed to `List<KeyValuePair<TKey, TValue>>` (_implementation_).

atifaziz:
> Could you please rebase this over master? It looks like you added this on top of the [2.0 beta 2](https://github.com/morelinq/MoreLINQ/releases/tag/2.0-beta-2), which has a different project structure.

Ignore this part. I was looking at a stale branch that was probably rewritten. Sorry.

MoreLinq.Test/CountByTest.cs
```diff
@@ -0,0 +1,103 @@
...
+                { "a", 3 },
+                { "B", 2 },
+                { "c", 1 },
+            };
+
+            result.AssertSequenceEqual(expectations);
+        }
+        
+        [Test]
+        public void CountByHasKeysOrderedLikeGroupBy()
```
atifaziz: If I change the last line of `CountByImpl`:

```c#
             return keys.Select(k => new KeyValuePair<TKey, int>(k, dic[k]));
```

to simply return `dic` instead:

```c#
             return dic;
```

then this test (`CountByHasKeysOrderedLikeGroupBy`) doesn't fail.

leandromoh: > If I change the last line of CountByImpl to simply return dic instead then this test (CountByHasKeysOrderedLikeGroupBy) doesn't fail.

In pratice, yes, it doesnt fail (because, currently, dictionary brings the results in order that keys were inserted), but as you once said yourself 

> dictionaries do not guarantee enumeration in any particular or logical order

So order in enumeration of dictionary can change anyday with a new implementation. To ensure CountBy always return in the order that keys were found the current last line is necessary.

leandromoh: > There is a test missing to check that CountBy is lazy.

What `CountByIsLazy ` test should exactly do? If I understood well, the implementation isn't lazy since I _must_ iterate over all the IEnumerable before return any KeyValuePair.

atifaziz: `CountBy` isn't lazy but it can and should be (like `GroupBy`). The iterator doesn't need to run until someone _iterates_ the results! You can make it lazy very easily by using `yield` to return the results. So instead of the following as the last line of `CountByImpl`:

```c#
            return keys.Select(k => new KeyValuePair<TKey, int>(k, dic[k]));
```

Do instead:

```c#
            foreach (var key in keys)
                yield return new KeyValuePair<TKey, int>(key, dic[key]);
```

Now the compiler will re-write the code to run _during iteration_, rendering it lazy!

leandromoh: how could I test if the return of CountBy is lazy (for CountByIsLazy) ?

atifaziz: > how could I write a test that check if the return of CountBy is lazy (for CountByIsLazy) ?

[See some of the existing tests that test for laziness](https://github.com/morelinq/MoreLINQ/search?utf8=%E2%9C%93&q=testfixture+lazy&type=Code) for inspiration.

atifaziz: > But It isn't the same thing since Select is lazy (do exactly the `foreach yield`) ?

Yes, `Select` is lazy except you compute the result imperatively _before_ you send them back via `Select`. That computation is happening when `CountBy` is called when it should happen when the enumerable returned by your method is enumerated by the caller (and which could be some time after).

leandromoh: I see, thanks for the explanation.  `CountByIsLazy` was implemented.

MoreLinq/CountBy.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <returns>A sequence of unique keys and their number of occurrences in the original sequence.</returns>
+        public static IEnumerable<KeyValuePair<TKey, int>> CountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey> comparer)
+        {
+            if (source == null) throw new ArgumentNullException("source");
+            if (keySelector == null) throw new ArgumentNullException("keySelector");
+
+            return CountByImpl(source, keySelector, comparer);
+        }
+
+        private static IEnumerable<KeyValuePair<TKey, int>> CountByImpl<TSource, TKey>(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey> comparer)
```
atifaziz: Now that we have all the tests in place, it would be good to work in some optimisations. The current implementation does too many lookups:

1. Key existence check: `dic.ContainsKey(key)`
2. Counter increment: `dic[key]++`
3. At the end when yielding back the results

Another optimisation could be is that as long as the current key is the same as the last, just increase the counter. This way, adjacent keys don't require containment tests on the dictionary.

atifaziz:
I've take a crack at implementing the [optimizations I suggested](https://github.com/morelinq/MoreLINQ/pull/207#discussion_r86244247). The tests are passing but feel free to review & let me know if you think we may need to add some tests to for corner cases where the optimization may fail.

## Benchmarks

I measured the overall benefit of the optimizations. Here's the benchmark code (using [BenchmarkDotNet](https://www.nuget.org/packages/BenchmarkDotNet/0.9.9)) that runs a `CountBy` on an ordered and unordered sequence of 10K integers (with duplicates) as keys:

```
using System;
using System.Linq;
using BenchmarkDotNet.Attributes;
using MoreLinq;

public class Benchmark
{
    static readonly int[] UnorderedData;
    static readonly int[] OrderedData;

    static Benchmark()
    {
        var xs = Enumerable.Range(1, 1000)
                           .SelectMany(x => Enumerable.Repeat(x, 10))
                           .ToArray();
        OrderedData = xs;
        UnorderedData = xs.OrderBy(_ => Guid.NewGuid()).ToArray();
    }

    [Benchmark]
    public static void CountByOnOrderedKeys() =>
        OrderedData.CountBy(e => e).Consume();

    [Benchmark]
    public static void CountByOnUnorderedKeys() =>
        UnorderedData.CountBy(e => e).Consume();
}
```

### Benchmark Environment

```
Host Process Environment Information:
BenchmarkDotNet.Core=v0.9.9.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-6600U CPU 2.60GHz, ProcessorCount=4
Frequency=2742185 ticks, Resolution=364.6727 ns, Timer=TSC
CLR=MS.NET 4.0.30319.42000, Arch=32-bit RELEASE
GC=Concurrent Workstation
JitModules=clrjit-v4.6.1080.0

Type=Benchmark  Mode=Throughput  
```

### Results without Optimizations

                 Method |      Median |     StdDev |
----------------------- |------------ |----------- |
   CountByOnOrderedKeys | 478.9549 us | 45.6884 us |
 CountByOnUnorderedKeys | 481.1692 us | 29.9369 us |

### Results with Optimizations

Method |      Median |     StdDev |
----------------------- |------------ |----------- |
   CountByOnOrderedKeys | 210.2153 us |  3.8459 us |
 CountByOnUnorderedKeys | 334.2708 us | 17.2886 us |

### Conclusion

The optimizations have improved the speed of `CountBy`! 🙌 In the unordered case (`CountByOnUnorderedKeys`), it runs 30% faster due to fewer dictionary look-ups. In the ordered case (`CountByOnOrderedKeys`), the improvement is 2 folds by avoiding look-ups as long as the key is not changing.

MoreLinq/CountBy.cs
```diff
@@ -0,0 +1,115 @@
...
+            // loop over the yielded pairs could span GC cycles. However,
+            // instead of doing simply the following:
+            //
+            // dic = null;
+            //
+            // the reference is nulled through a method that the JIT compiler
+            // is told not to try and inline; done so assuming that the above
+            // method could have been turned into a NOP (in theory).
+
+            Null(ref dic); // dic = null;
```
fsateler: I think it would be less magic and more readable to simply split the counting loop to a separate function.

For other readers (and myself, I had to google this), setting `dic = null` is usually not useful, because the CLR knows that you are not using the variable anymore, and thus marks the referenced object as eligible for collection. However, the CLR cannot be sure when a generator or async method is used (I only found a passing reference by [Stephen Cleary](http://blog.stephencleary.com/2010/02/q-should-i-set-variables-to-null-to.html), is there a more detailed source?), and thus it waits until the method ends. I *think* this is because the generator method is compiled to a class, with all local parameters as fields on that class. If the instance of this generated class outlives the use of the fields in that class, the CLR cannot know this. However, this same analysis would suggest that it would be a bug in either the C# compiler or the JIT to turn `dict = null` into a NOP.

fsateler: I was too terse I think: by split the counting loop into a separate function, I mean that that function runs eagerly, not lazily, thus avoiding this problem.

atifaziz: > I think it would be less magic and more readable to simply split the counting loop to a separate function.

That's an excellent suggestion! 👍 It doesn't take away the magic one would need to understand that is the _raison d'être_ for split. Someone in the future could say, “This is stupid. Let me inline this for simplicity's sake!”

> analysis would suggest that it would be a bug in either the C# compiler or the JIT to turn `dict = null` into a NOP.

Yeah it would be, especially if it's been lifted into a field but it depends on how clever the compiler can get about tying the local variable lifetimes to various stages of the state machine it generates. Technically, the `dic` would only need to exist for the first state of the machine and therefore could be local to the `case` of the `switch` generated for the `MoveNext` machine. Only `key`, `counts` and `i` would need to extend over the rest of the states. However, I neither checked out the generated code (by the C# or JIT compiler) nor did I want to depend on it (today it's a `switch`/`case`, tomorrow it could be done in other ways). I guess I was trying to err on the safe side but scoping to another function altogether is cleaner.

leandromoh:
@atifaziz you've done a great job with the optimization! I would begin it today but you were faster lol 
I have only one doubt: why did you use 

```
dic.Comparer.GetHashCode(prevKey) == dic.Comparer.GetHashCode(key) 
&& dic.Comparer.Equals(prevKey, key)
```
instead of just `dic.Comparer.Equals(prevKey, key)` ?

atifaziz:
@leandromoh It's not worth calling `Equals` for performance reasons if the hash code is already different because that's what a dictionary would do internally too. Comparing hash codes can be faster in many cases. For example, `Equals` for a string will check for every character but a hash code comparison is just an integer equality check and therefore much faster (same goes for structural value types that are compounds of primitives). If hash codes match then a full equality comparison of the actual content is needed because the hash is just a digest & can be the same for several values of a type.

Since you're happy with the optimisations then I can merge this PR if you can just fix [the type parameter reference doc issue](https://github.com/morelinq/MoreLINQ/pull/207#discussion_r86241035).

MoreLinq/CountBy.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// Applies a key-generating function to each element of a sequence and returns a sequence of 
+        /// unique keys and their number of occurrences in the original sequence.
+        /// An additional argument specifies a comparer to use for testing equivalence of keys.
+        /// </summary>
+        /// <typeparam name="TSource">Type of the elements of the source sequence.</typeparam>
+        /// <typeparam name="TKey">Type of the projected element.</typeparam>
+        /// <param name="source">Source sequence.</param>
+        /// <param name="keySelector">Function that transforms each item of source sequence into a key to be compared against the others.</param>
+        /// <param name="comparer">The equality comparer to use to determine whether or not keys are equal.
+        /// If null, the default equality comparer for <c>TSource</c> is used.</param>
```
leandromoh: Done.

leandromoh:
`Can we address this now as the last item? Should be done then. Thanks!`

Done!

atifaziz:
Merged into `master` by a02b8f97af48670c2ff628e0d31bc6b878343064.

Thanks for adding this and working through all the reviews.

---
Note that I squashed your commits in the final merge except the optimizations because I felt it would be good to keep a change history of those in case anything needs to be reverted or reviewed if ever a bug is reported.

atifaziz:
I know [I said MoreLINQ is in lock down for 2.0 release](https://github.com/morelinq/MoreLINQ/pull/197#issuecomment-255585332) but I'm being bad here and letting this one slip… 😇 

`CountBy` is now published in [2.0 beta 6](https://www.nuget.org/packages/morelinq/2.0.0-beta06).

atifaziz:
[Documentation published](https://morelinq.github.io/2.0/ref/api/html/Overload_MoreLinq_MoreEnumerable_CountBy.htm).

leandromoh:
@atifaziz that's a good new! When do you plan to publish the release 2.0 with the other methods?

I open the link to see CountBy documentation and perceive that my name is not in the footer with others contributor. When do you plan to add it ?

atifaziz:
> When do you plan to publish the release 2.0 with the other methods?

Soon-ish.

> my name is not in the footer with others contributors.

Sorry, missed that one but fixed now by 58d6c62f1aacb4a53fa4faba2486516bacbd363e. BTW, that's an aggregate of copyright notices, not contributors list.

Issue #209, opened 2016-11-01
Namespace for experimental operators


atifaziz:
Follow up from #205:

>> I don't want to create & maintain another whole library for just one method in the async category. Perhaps a `MoreLinq.Sandbox` would be better for all ideas that are cooking and unsupported?

> This sounds like a good idea, but orthogonal to what I proposed.

Coming back to this, how about introducing a new namespace called `MoreLinq.Labs` in the current, one and only, assembly for experimental ideas with the disclaimer & understanding that anything in there is work-in-progress and could be completely removed in the future. Anyone using something in there should be prepared to deal with breaking changes. If an operator stabilizes over time and proves to be popular then it can be promoted to the `MoreLinq` namespace. The benefits of this approach would be:

- Use existing assembly & NuGet packages instead of maintaining a new and separate one
- Have one release to worry about
- Increase the release cadence of the project by shipping possibly unstable work

Risks? Confusion & some disgruntled users who took the happy path and ignored the writing on the wall.

atifaziz:
Namespace proposals:

`MoreLinq.Edge`: Stuff that's really slated for a future release but shipping part of the current release train. It is potentially unstable & not battle-tested.

`MoreLinq.Labs`: Experimental stuff that may not only be unstable & untested but also may never become part of MoreLINQ proper (therefore risks being completely removed). The reason to ever release this stuff is to summon feedback from wider field usage of existing MoreLINQ users.

fsateler:
There is a smallish problem with including in the same assembly. Visual Studio (I'm using 2015 here) will offer to include the `using` statement necessary for an extension method with the `Ctrl+.` shortcut. This increases the chance someone uses this namespace without fully understanding the consequences.

I looked for an attribute that would generate a compiler warning like `ObsoleteAttribute`, but I couldn't find any. Would it be too annoying/confusing to use `ObsoleteAttribute`?

Another mechanism would be to be more explicit in the namespace name: `MoreLinq.Experimental`.

BTW, how could one get feedback on usage of the experimental methods? It seems to me this is an important part of the point of the exercise, isn't it?

atifaziz:
> Would it be too annoying/confusing to use `ObsoleteAttribute`?

Definitely.

I can live with `MoreLinq.Experimental`. It seems like a good compromise. Not everyone may be current with hipster terms like `Labs`. 🤖 

[...]

PR #210, opened 2016-11-02
Partition for groupings

atifaziz:
Enables partitioning of groupings of Boolean keys and other any 3 cases.

Tests are pending.

## Sample 1

Partition a grouping where the key is a Boolean:

```c#
var q1 =
    Enumerable.Range(0, 10)
      .GroupBy(x => x % 2 == 0)
      .Partition((evens, odds) => new { odds, evens });

Console.WriteLine("odds  = " + q1.odds.ToDelimitedString(", "));
Console.WriteLine("evens = " + q1.evens.ToDelimitedString(", "));
```

Output:

```
odds  = 1, 3, 5, 7, 9
evens = 0, 2, 4, 6, 8
```

## Sample 2

Partition a grouping where the key is a nullable Boolean:

```c#
var q2 =
    Enumerable.Range(0, 10)
      .Select(x => (int?) x)
      .Interleave(Enumerable.Repeat((int?) null, 3))
      .GroupBy(x => x != null ? x < 5 : (bool?) null)
      .Partition((lt5, gte5, nil) => new { lt5, gte5, nil });

Console.WriteLine("lt5  = " + q2.lt5.ToDelimitedString(", "));
Console.WriteLine("gte5 = " + q2.gte5.ToDelimitedString(", "));
Console.WriteLine("nil  = " + q2.nil.Select(_ => "?").ToDelimitedString(", "));
```

Output:

```
lt5  = 0, 1, 2, 3, 4
gte5 = 5, 6, 7, 8, 9
nil  = ?, ?, ?
```

## Sample 3

Partition grouping into multiples of 3 and the rest:

```c#
var q3 =
    Enumerable.Range(0, 10)
      .GroupBy(x => x % 3)
      .Partition(0, (m3, etc) => new { m3, etc });

Console.WriteLine("m3  = " + q3.m3.ToDelimitedString(", "));
Console.WriteLine("etc = " 
    + string.Join("; ",
        from g in q3.etc
        select new { k = g.Key, ms = g.ToDelimitedString(", ") }));
```

Output:

```
m3  = 0, 3, 6, 9
etc = { k = 1, ms = 1, 4, 7 }; { k = 2, ms = 2, 5, 8 }
```

## Sample 4

Parition groupings by keys (remainders) matching 0, 1 & 2 (the non-matching groupings `_` are ignored because they can never occur):

```c#
var q4 =
    Enumerable.Range(0, 10)
      .GroupBy(x => x % 3)
      .Partition(0, 1, 2, (ms, r1, r2, _) => new { ms, r1, r2 });

Console.WriteLine("ms  = " + q4.ms.ToDelimitedString(", "));
Console.WriteLine("r1  = " + q4.r1.ToDelimitedString(", "));
Console.WriteLine("r2  = " + q4.r2.ToDelimitedString(", "));
```

Output:

```
ms  = 0, 3, 6, 9
r1  = 1, 4, 7
r2  = 2, 5, 8
```

## Sample 5

Partition of groupings using a key comparer:

```c#
var words =
    new[] { "foo", "bar", "FOO", "Bar", "baz", "quux", "bAz", "QuuX" };
    
var q5 =
    words
      .GroupBy(s => s, StringComparer.OrdinalIgnoreCase)
      .Partition("foo", "bar", StringComparer.OrdinalIgnoreCase,
                 (foos, bars, etc) => new { foos, bars, etc });

Console.WriteLine("foo = " + q5.foos.ToDelimitedString(", "));
Console.WriteLine("bar = " + q5.bars.ToDelimitedString(", "));
Console.WriteLine("etc = " 
    + string.Join("; ",
        from g in q5.etc
        select new { k = g.Key, ms = g.ToDelimitedString(", ") }));
```

Output:

```
foo = foo, FOO
bar = bar, Bar
etc = { k = baz, ms = baz, bAz }; { k = quux, ms = quux, QuuX }
```

## Sample 6

Cascaded partition of groupings:

```c#
var q6 =
    words
      .GroupBy(s => s, StringComparer.OrdinalIgnoreCase)
      .Partition("foo", "bar", StringComparer.OrdinalIgnoreCase,
                 (foo, bar, etc) =>
                    etc.Partition("quux", "baz", StringComparer.OrdinalIgnoreCase,
                        (quux, baz, _) => new { foo, bar, baz, quux }));

Console.WriteLine("foo  = " + q6.foo.ToDelimitedString(", "));
Console.WriteLine("bar  = " + q6.bar.ToDelimitedString(", "));
Console.WriteLine("baz  = " + q6.baz.ToDelimitedString(", "));
Console.WriteLine("quux = " + q6.quux.ToDelimitedString(", "));
```

Output:

```
foo  = foo, FOO
bar  = bar, Bar
baz  = baz, bAz
quux = quux, QuuX
```


leandromoh:
In Sample 1 we have

```
odds  = 1, 3, 5, 7, 9
evens = 1, 3, 5, 7, 9
```

Is it right?

fsateler:
I like this. I have needed methods like this one before, and ended up using `ToLookup`. I'll take a look at the code later.

A few overall questions though:

- Requiring a `GroupBy` before partition feels a bit strange to me. Why not do the grouping inside `Partition` itself? A possible answer is that one might want to filter the `IGrouping` prior to partitioning. Not sure which case is more common though.
- The size of the collections is (can be) known at `resultSelector` invocation time. I think passing a readonly `IList<T>` (if the order is to be guaranteed) or `ICollection<T>` (if not guaranteed) would be a better choice. Since streaming is out of the question for this operator, the passed `IEnumerable`s cannot possibly be lazy.

PS: It is unfortunate that c# does not have syntax that would allow making this truly generic over the number of possible buckets...

atifaziz:
@leandromoh Obviously not. :) Looks like a copy & paste error (printing `odds` twice). I've fixed the sample now & thanks for spotting that. 👍 

@fsateler Glad to hear you like it. I've needed this countless times too & just got tired & decided it was time to give it a go (and in a generalised way over groupings). Looking forward to your review.

> Requiring a GroupBy before partition feels a bit strange to me.

To be honest, I wanted to start with simpler signatures and not having to deal with `keySelector` and `resultSelector` (of which there is already another for `Partition`). It would have also exploded the number of overloads and it can always be added later. Since we have [`IGrouping<TKey, TElement>`](https://msdn.microsoft.com/en-us/library/bb344977(v=vs.100).aspx) given by LINQ, I thought it's best to extend on that interface. A separate version for `IEnumerable<T>` could be added later that does the grouping if this proves useful. And as you so rightly observed, these versions allow filtering & more on the groups before partitioning.

> The size of the collections is…known at `resultSelector` invocation time…passing a readonly `IList<T>`…or `ICollection<T>`…would be a better choice.

Unless you're in a concurrent world? And how can you assume how the grouping was created? An `IGrouping<TKey, TElement>` only promises to be a sequence of group member by inheriting `IEnumerable<TElement>` so I can't assume more. I know it's frustrating and I'd rather be indexing into a list than calling `ElementAt` and hoping it is one.

fsateler:
> Since we have IGrouping<TKey, TElement> given by LINQ, I thought it's best to extend on that interface. ... And as you so rightly observed, these versions allow filtering & more on the groups before partitioning.

I agree that using a type already given to us is good. OTOH, `IGrouping` is only provided as an interface, and no actual implementations, thus either forcing use of `GroupBy` or `ToLookup` to be able to consume this method.

> A separate version for IEnumerable<T> could be added later that does the grouping if this proves useful.

Ack. 

> > The size of the collections is…known at resultSelector invocation time…passing a readonly IList<T>…or ICollection<T>…would be a better choice.
>
> Unless you're in a concurrent world? And how can you assume how the grouping was created? An IGrouping<TKey, TElement> only promises to be a sequence of group member by inheriting IEnumerable<TElement> so I can't assume more. I know it's frustrating and I'd rather be indexing into a list than calling ElementAt and hoping it is one.

That is the problem of relying on `IGrouping` 😉 . More seriously, though, if the `IGrouping` overload is preserved, there is no choice but to use `IEnumerable<T>` instead of a collection interface.

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Partitions a grouping by Boolean keys into a projection of true
+        /// elements and false elements, respectively.
+        /// </summary>
+
+        public static TResult Partition<T, TResult>(this IEnumerable<IGrouping<bool, T>> source,
+            Func<IEnumerable<T>, IEnumerable<T>, TResult> resultSelector) =>
```
fsateler: This is the first use of expression-bodied members in MoreLINQ. Is this intended?

fsateler: Also, parameters indented at same level as body looks consfusing to me.

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Partitions a grouping by Boolean keys into a projection of true
+        /// elements and false elements, respectively.
+        /// </summary>
```
fsateler: Parameter and return documentation is missing, on all overloads.

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                var i = count > 0 && comparer.Equals(e.Key, k1) ? 0
+                      : count > 1 && comparer.Equals(e.Key, k2) ? 1
+                      : count > 2 && comparer.Equals(e.Key, k3) ? 2
+                      : -1;
+
+                if (i < 0)
+                    (etc ?? (etc = new List<IGrouping<TKey, TElement>>())).Add(e);
+                else
+                    groups[i] = e;
```
fsateler: If the source enumerable is not unique by keys, then the last group seen will be the one used. This should be documented.

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            foreach (var e in source)
+            {
+                var i = count > 0 && comparer.Equals(e.Key, k1) ? 0
+                      : count > 1 && comparer.Equals(e.Key, k2) ? 1
+                      : count > 2 && comparer.Equals(e.Key, k3) ? 2
+                      : -1;
+
+                if (i < 0)
+                    (etc ?? (etc = new List<IGrouping<TKey, TElement>>())).Add(e);
```
fsateler: An assignment within an lvalue looks weird, I think it is more readable to expand this:

```csharp
etc = etc ?? new List<IGrouping<TKey, TElement>>();
etc.Add(e);
```

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+                Enumerable.Empty<TElement>(),
+                Enumerable.Empty<TElement>(),
+            };
+
+            foreach (var e in source)
+            {
+                var i = count > 0 && comparer.Equals(e.Key, k1) ? 0
+                      : count > 1 && comparer.Equals(e.Key, k2) ? 1
+                      : count > 2 && comparer.Equals(e.Key, k3) ? 2
+                      : -1;
```
fsateler: I find nested `?:` operators very difficult to follow. What is the precedence of `?` and `:` ? Again, I would expand this:

```csharp
int i;
if (count > 0 && comparer.Equals(e.Key, k1))
    i = 0;
else if (count > 1 && comparer.Equals(e.Key, k2))
    i = 1;
else if (count > 2 && comparer.Equals(e.Key, k3))
    i = 2;
else
    i = -1;
```

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+            comparer = comparer ?? EqualityComparer<TKey>.Default;
+
+            List<IGrouping<TKey, TElement>> etc = null;
+
+            var groups = new[]
+            {
+                Enumerable.Empty<TElement>(),
+                Enumerable.Empty<TElement>(),
+                Enumerable.Empty<TElement>(),
+            };
```
fsateler: Why are the groups pre-inited to `Enumerable.Empty` but `etc` isn't? Any particular reason?

atifaziz:
> if the `IGrouping` overload is preserved, there is no choice but to use `IEnumerable<T>` instead of a collection interface.

Sorry, not sure if I'm following you here. Care to rephrase?

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// </summary>
+
+        public static TResult Partition<TKey, TElement, TResult>(this IEnumerable<IGrouping<TKey, TElement>> source,
+            TKey key, IEqualityComparer<TKey> comparer,
+            Func<IEnumerable<TElement>, IEnumerable<IGrouping<TKey, TElement>>, TResult> resultSelector) =>
+            PartitionImpl(source, 1, key, default(TKey), default(TKey), comparer, (a, b, c, rest) => resultSelector(a, rest));
+
+        /// <summary>
+        /// Partitions a grouping into a projection of elements matching two
+        /// sets of keys and those groups that do not.
```
fsateler: s/sets of // ? Or instead "elements matching a set of two keys". Same for the 3 key overloads

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+            Func<IEnumerable<TElement>, IEnumerable<TElement>, IEnumerable<TElement>, IEnumerable<IGrouping<TKey, TElement>>, TResult> resultSelector) =>
+            PartitionImpl(source, 3, k1, k2, k3, comparer, resultSelector);
+
+        static TResult PartitionImpl<TKey, TElement, TResult>(IEnumerable<IGrouping<TKey, TElement>> source,
+            int count, TKey k1, TKey k2, TKey k3, IEqualityComparer<TKey> comparer,
+            Func<IEnumerable<TElement>, IEnumerable<TElement>, IEnumerable<TElement>, IEnumerable<IGrouping<TKey, TElement>>, TResult> resultSelector)
+        {
+            Debug.Assert(count > 0 && count <= 3);
+
+            if (source == null) throw new ArgumentNullException(nameof(source));
```
fsateler: First instance of `nameof` in MoreLINQ, same question as for expression-bodied members.

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+            source.Partition(true, false, (t, f, _) => resultSelector(t, f));
+
+        /// <summary>
+        /// Partitions a grouping by nullable Boolean keys into a projection of
+        /// true elements, false elements and null elements, respectively.
+        /// </summary>
+
+        public static TResult Partition<T, TResult>(this IEnumerable<IGrouping<bool?, T>> source,
+            Func<IEnumerable<T>, IEnumerable<T>, IEnumerable<T>, TResult> resultSelector) =>
+            source.Partition(true, false, null, (t, f, n, _) => resultSelector(t, f, n));
```
fsateler: Here you use `_` to mark an unused parameter, I think you should do this in all overloads.

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
...
+        public static TResult Partition<TKey, TElement, TResult>(this IEnumerable<IGrouping<TKey, TElement>> source,
+            TKey k1, TKey k2, TKey k3, IEqualityComparer<TKey> comparer,
+            Func<IEnumerable<TElement>, IEnumerable<TElement>, IEnumerable<TElement>, IEnumerable<IGrouping<TKey, TElement>>, TResult> resultSelector) =>
+            PartitionImpl(source, 3, k1, k2, k3, comparer, resultSelector);
+
+        static TResult PartitionImpl<TKey, TElement, TResult>(IEnumerable<IGrouping<TKey, TElement>> source,
+            int count, TKey k1, TKey k2, TKey k3, IEqualityComparer<TKey> comparer,
+            Func<IEnumerable<TElement>, IEnumerable<TElement>, IEnumerable<TElement>, IEnumerable<IGrouping<TKey, TElement>>, TResult> resultSelector)
+        {
+            Debug.Assert(count > 0 && count <= 3);
```
fsateler: I think this should be changed to a regular check+throw, so that unit tests have a chance of catching this.

MoreLinq/GroupPartition.cs
```diff
@@ -0,0 +1,144 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2009 Atif Aziz. All rights reserved.
```
fsateler: 2009 => 2016

MoreLinq/GroupPartition.cs
fsateler: Also, for the particular boolean overloads, it should be documented which of the `resultSelector` parameters corresponds to `true` and `false`.

fsateler:
> Sorry, not sure if I'm following you here. Care to rephrase?

I mean that as long as the implementation is based on top of `IGrouping`, it is somewhat restricted to having to return `IEnumerable`s, because there is no guarantee the `IGrouping` is a collection.  The alternative would be calling `ToList` or similar before invoking the `resultSelector`, but that is potentially costly.

fsateler:
> as long as the implementation is based on top of IGrouping

Actually, now that I think of this, why not pass the `IGrouping` directly instead of an `IEnumerable` ?

```csharp
public static TResult Partition<T, TResult>(this IEnumerable<IGrouping<bool, T>> source,
            Func<IGrouping<bool, T>, IGrouping<bool, T>, TResult> resultSelector)
// ...
```

MoreLinq/GroupPartition.cs
atifaziz: I documented just the summaries for now, which are the most important bits. The rest will follow as time permits. Meanwhile, this could be released earlier, and dare I say even without tests, if it's put into the _edge_ namespace as being discussed in #209.

> for the particular boolean overloads, it should be documented which of the `resultSelector` parameters corresponds to `true` and `false`.

This is already done: _a projection of true elements and false elements, respectively_. Perhaps it's not obvious but “respectively” means in the aforementioned order.

On that note, this is different F# where, for example, [`List.partition`][listpart] returns a couple of false (first) & true (second) parts. Should we align? It does mean though that in the case of `bool?` keys, the parts become: null, false, true. Or should we go opinionated, that true then false reads better in practice because it's easier to express & think in terms of passing than failing conditions? 

[listpart]: https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/list.partition%5B't%5D-function-%5Bfsharp%5D

MoreLinq/GroupPartition.cs
atifaziz: All done in the interest of speed to get this published for early feedback.

I use expression-bodied members whenever possible though I'm willing to revert in the interest of consistency with the rest of the code base. Once 2.0 is golden, it'll be time to make a sweeping change & upgrade the code base to C# 6.

MoreLinq/GroupPartition.cs
atifaziz: I do that where it's limited to one unused parameter as that's all C# can afford us. Do you suggest to use 2 underscores for the second unused? As in:

```c#
PartitionImpl(source, 1, key, default(TKey), default(TKey), comparer, (a, _, __, rest) => resultSelector(a, rest));
```

MoreLinq/GroupPartition.cs
atifaziz: The types are different. Also the list would have to be allocated for `etc` to mean empty. In the case of `groups`, initialising to `Enumerable.Empty<TElement>()` has negligible cost (haven't checked but suspect the call may even get inlined).

MoreLinq/GroupPartition.cs
atifaziz: Good point but isn't it better to turn it into an error? Although doing so for non-matching groups would be unnecessarily expensive. I'm inclined to document this as undefined behaviour even if someone observes what you say. I guess the best test is what you'd expect `Partition` to do if the came from `GroupAdjacent` instead of `GroupBy`. Have to think about this one.

atifaziz:
> why not pass the `IGrouping` directly instead of an `IEnumerable` ?

With `Partition`, you're _destructuring groupings_ via lambda arguments. Why would you want to keep the keys as it's going to be used for a small set of _known_ keys? For non-matching cases, you get the groupings because you might still need the key (see sample 6 for how that's used for nested partitioning). For the case where keys are Boolean values (the signature you quoted), it's especially useless. It may have some value if you want to capture the original keys when using a custom comparer (e.g. the original case of a string key using a case-insensitive comparer). In the rare event that you really want to keep the keys for the destructured cases, you can attach it back.

Do you have a concrete example to demonstrate what the current design/signature would prevent?

MoreLinq/GroupPartition.cs
fsateler: >>  for the particular boolean overloads, it should be documented which of the resultSelector parameters corresponds to true and false.
>
>This is already done: a projection of true elements and false elements, respectively. Perhaps it's not obvious but “respectively” means in the aforementioned order.

Right, sorry. I meant that this needs to be in the `<param />` entry. 

> On that note, this is different F# where, for example, List.partition returns a couple of false (first) & true (second) parts. Should we align? It does mean though that in the case of bool? keys, the parts become: null, false, true. Or should we go opinionated, that true then false reads better in practice because it's easier to express & think in terms of passing than failing conditions?

Hmm. OTOH, those from a C-like background may find it more natural in the `F#` order, given that `false` is `0` and `true` is (usually) `1`, and then parameters are in order. I value consistency, but I'm not sure how many people use both F# and C#. I'm undecided.

fsateler: > I documented just the summaries for now, which are the most important bits. The rest will follow as time permits. Meanwhile, this could be released earlier, and dare I say even without tests, if it's put into the edge namespace as being discussed in #209.

Hmmm. Maybe the guidelines for acceptance should also be discussed in #209. But I figure at least the trivial tests (null arguments, etc) should be there before merging.

fsateler:
>>why not pass the IGrouping directly instead of an IEnumerable ?
>
> Do you have a concrete example to demonstrate what the current design/signature would prevent?

No. That was mostly thinking out loud, and your explanation makes a lot of sense. There is little value in preserving the `IGrouping`.

MoreLinq/GroupPartition.cs
fsateler: > Once 2.0 is golden

BTW, maybe it is a good idea to establish a milestone, to define what needs to be done before releasing 2.0. Then people can contribute towards that end, and maybe even decide which PRs to look at before or after release.

MoreLinq/GroupPartition.cs
fsateler: > Also the list would have to be allocated for etc to mean empty. In the case of groups, initialising to Enumerable.Empty<TElement>() has negligible cost

OK

MoreLinq/GroupPartition.cs
fsateler: > It is meant to be read here like a switch/case where case can be an expression and the whole switch is an expression evaluating to a single value. I'd like to keep this as-is.

OK. I still disagree, but there is nothing else to claim other than style preferences 😉

fsateler:
Alternative names to avoid collition with the already existing `Partition` (btw, the currently existing partition is not currently appearing on the api docs online):

1. `Pivot`: As in pivoting over the possible keys in the `IGrouping` collection
2. `Deconstruct`: as in tuple deconstruction.
3. `Match`: as in pattern matching (a very limited pattern, though).

atifaziz:
Thanks for those names suggestions, @fsateler. I'll give them some thought.

> …`Partition` (btw, the currently existing partition is not currently appearing on the api docs online)

If you look at the [changes between the last beta and up to the release of 2.0](https://github.com/morelinq/MoreLINQ/compare/2.0-beta-6...v2.0.0), `Partition` was removed in 7a2c72bada4ba45343e44d5291084073ba91eecf for the reason given in the commit message.

atifaziz:
We don't have an issue for this PR so let's pair it with #264.

atifaziz:
@fsateler Let me know if I missed anything from your review/feedback.

MoreLinq/Partition.cs
```diff
@@ -0,0 +1,297 @@
...
+        /// <summary>
+        /// Partitions a grouping into a projection of elements matching a key
+        /// and those groups that do not. An additional parameter specifies how
+        /// to compare keys for equality.
+        /// </summary>
+        /// <typeparam name="TKey">Type of keys in source groupings.</typeparam>
+        /// <typeparam name="TElement">Type of elements in source groupings.</typeparam>
+        /// <typeparam name="TResult">Type of the result.</typeparam>
+        /// <param name="source">The source sequence.</param>
+        /// <param name="key">The key to partition.</param>
```
fsateler: This may be my non-native English, but I fail to parse this description. I think it would be clearer to say: `The key of the matching partition`. Mutatis mutandis, the same applies to the overloads with more `key` arguments.

MoreLinq.Test/PartitionTest.cs
```diff
@@ -0,0 +1,194 @@
...
+
+    [TestFixture]
+    public class PartitionTest
+    {
+        [Test]
+        public void PartitionWithResultSelector()
+        {
+            var (evens, odds) =
+                Enumerable.Range(0, 10)
+                          .Partition(x => x % 2 == 0, ValueTuple.Create);
```
fsateler: This is a very nice example of this overload. Maybe it should be included in the xmldoc?

MoreLinq.Test/PartitionTest.cs
```diff
@@ -0,0 +1,194 @@
...
+
+            using (var r = etc.Read())
+            {
+                var r1 = r.Read();
+                Assert.That(r1.Key, Is.EqualTo(1));
+                Assert.That(r1, Is.EquivalentTo(new[] { 1, 4, 7 }));
+
+                var r2 = r.Read();
+                Assert.That(r2.Key, Is.EqualTo(2));
+                Assert.That(r2, Is.EquivalentTo(new[] { 2, 5, 8 }));
```
fsateler: This test (and all the tests that have more than one grouping in `etc`) assert that the result is ordered by the order in which keys are seen, but that is not documented. So either the documentation should be updated to include this guarantee, or change these tests to relax that requirement.

MoreLinq.Test/PartitionTest.cs
```diff
@@ -0,0 +1,194 @@
...
+            var (ms, r1, r2, etc) =
+                Enumerable.Range(0, 10)
+                    .GroupBy(x => x % 3)
+                    .Partition(0, 1, 2, ValueTuple.Create);
+
+            Assert.That(ms, Is.EquivalentTo(new[] { 0, 3, 6, 9 }));
+            Assert.That(r1, Is.EquivalentTo(new[] { 1, 4, 7 }));
+            Assert.That(r2, Is.EquivalentTo(new[] { 2, 5, 8 }));
+            using (var r = etc.Read())
+                r.ReadEnd();
```
fsateler: I would find it clearer to `Assert.That(etc, Is.Empty)`, any particular reason to use the `SequenceReader` in this context?

atifaziz: > any particular reason to use the `SequenceReader` in this context?

Besides me being sloppy, no. 😕

MoreLinq.Test/PartitionTest.cs
atifaziz: Added in 7ef844287b0883dab371bea3be71642b856d28bb but just for the simplest overload.

atifaziz:
Wonder if this is also not a good time to add an overload without any projection function that simply returns a tuple:

```c#
public static (IEnumerable<T> True, IEnumerable<T> False)
    Partition<T>(this IEnumerable<T> source, Func<T, bool> predicate) =>
    source.Partition(predicate, ValueTuple.Create);
```

The usage would then be:

```c#
var (evens, odds) = Enumerable.Range(0, 10).Partition(x => x % 2 == 0);
```

Here it is in action in a C# Interactive session:

```
Microsoft (R) Visual C# Interactive Compiler version 2.1.0.61520

> #r "MoreLinq.dll"
> using MoreLinq;
> var res = Enumerable.Range(0, 10).Partition(x => x % 2 == 0);
> res.True
Lookup<bool, int>.Grouping { 0, 2, 4, 6, 8 }
> res.False
Lookup<bool, int>.Grouping { 1, 3, 5, 7, 9 }
```

Note tuple element names of `True` & `False`.

fsateler:
> Wonder if this is also not a good time to add an overload without any projection function that simply returns a tuple

Oh, that sounds very useful.

atifaziz:
> Oh, that sounds very useful.

Added in 2a9fa5518595dff2a734df08f5ca244f59356825.

atifaziz:
@fsateler We good to merge this? Did I miss anything from the changes you requested in your review?

fsateler:
> @fsateler We good to merge this? Did I miss anything from the changes you requested in your review?

You did not comment on my [comment on the `key` description](https://github.com/morelinq/MoreLINQ/pull/210#discussion_r111692578). If you disagree the description is unclear (or you think the new description is worse than the original), then I'm fine to merge.

MoreLinq/Partition.cs
atifaziz: Here's my thinking…the method is called `Partition` so it's partitioning or extracting the group with that key. Would the “the key to partition _on_“ read clearer?

fsateler: > Would the “the key to partition on“ read clearer?

That would be a bit better. At least I can parse that :)

atifaziz: LGTY now with 97ed6541c9f5d333702b4b40b89a1c9cd280ceb9?

PR #212, opened 2016-11-09
Implemented NthOrDefault & Nth

leandromoh:
Implemented NthOrDefault and Nth, similar to FirstOrDefault and First, respectively, but you pass the "position" of the desired element as argument.

Currently missing to add method reference in "project.json" and "README.md" files.
I pretend to do it latter to avoid conflicts.

### DIFF
--- a/MoreLinq.Test/NthOrDefaultTest.cs
+++ b/MoreLinq.Test/NthOrDefaultTest.cs
@@ -1,0 +1,199 @@
+﻿#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2016 Leandro F. Vieira (leandromoh). All rights reserved.
+// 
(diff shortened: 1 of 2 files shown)


atifaziz:
Apart from the overload accepting a predicate and using a positional indexing (i.e. 1-based), how are these different from [`ElementAt`](https://msdn.microsoft.com/en-us/library/bb299233(v=vs.110).aspx) or [`ElementAtOrDefault`](https://msdn.microsoft.com/en-us/library/bb494386(v=vs.110).aspx)?

leandromoh:
Nothing. The difference is _exactly_ the items you mentioned. Also, conceptually `ElementAtOrDefault` and `NthOrDefault` are differents, because when you pass the predicate you are not caring about positional indexing itself, but about the ordinalty of an specific element of those who satisfy the predicate..

Issue #177, opened 2016-05-29
FillForward (last observation carry forward)


atifaziz:
Consider adding a LOCF (Last Observation Carry Forward) operator that replaces some definition of _blank_ (defined by a predicate) in a sequence with the last observed _non-blank_.

This can be implement today using `Scan` + `Select` + `Skip` but it's cumbersome to read, write and get right each time, and it doesn't quite capture the intended operation succinctly:

```c#
public static IEnumerable<T> Locf<T>(this IEnumerable<T> source, Func<T, bool> predicate) =>
    source.Scan(new { Seeded = false, Value = default(T) },
                (run, e) => !predicate(e)
                          ? new { Seeded = true, Value = e }
                          : !run.Seeded
                          ? new { Seeded = false, Value = e }
                          : run)
                .Select(e => e.Value)
                .Skip(1);
```

A prototype implementation could be as follows:

``` c#
partial class MoreEnumerable
{
    public static IEnumerable<Nullable<T>> Locf<T>(this IEnumerable<Nullable<T>> source)
        where T : struct =>
        source.Locf(e => e == null);

    public static IEnumerable<T> Locf<T>(this IEnumerable<T> source) where T : class =>
        source.Locf(e => e == null);

    public static IEnumerable<T> Locf<T>(this IEnumerable<T> source, Func<T, bool> predicate)
    {
        var seeded = false;
        var seed = default(T);
        foreach (var item in source)
        {
            var blank = predicate(item);
            if (blank)
            {
                yield return seeded ? seed : item;
            }
            else
            {
                seeded = true;
                seed = item;
                yield return item;
            }
        }
    }
}
```

Usage:

``` c#
var xs = new int?[] { null, null, 1, 2, null, null, null, null, 3, 4, null };
var ys = xs.Locf();
// ys will be { null, null, 1, 2, 2, 2, 2, 2, 3, 4, 4 }
```

Note that this _does not_ remove blanks at the start of the input sequence, before the first non-blank value. They can be removed subsequently using `SkipWhile`.

atifaziz:
Closed by 2772af00f67c9f13356e96318685a1ac0eb8166d.

PR #213, opened 2016-11-09
Add FillForward (last observation carry forward)

atifaziz:
See #177 for background and more. Reviews welcome.

~Tests are pending.~

- [x] Tests
- [x] Add to operators in package description and release notes
- [x] Add to operators list in README

### DIFF
--- a/MoreLinq/Locf.cs
+++ b/MoreLinq/Locf.cs
@@ -64,14 +64,14 @@ namespace MoreLinq
 
         public static IEnumerable<T> Locf<T>(this IEnumerable<T> source, Func<T, bool> predicate)
         {
+            if (source == null) throw new ArgumentNullException("source");
+            if (predicate == null) throw new ArgumentNullException("predicate");
+
             return LocfImpl(source, predicate);
         }
 
         static IEnumerable<T> LocfImpl<T>(IEnumerable<T> source, Func<T, bool> predicate)
         {
-            if (source == null) throw new ArgumentNullException("source");
-            if (predicate == null) throw new ArgumentNullException("predicate");
-
             var seeded = false;
             var seed = default(T);
             foreach (var item in source)
--- a/MoreLinq/Locf.cs
+++ b/MoreLinq/Locf.cs
@@ -1,0 +1,113 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2016 Atif Aziz. All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Returns the source sequence with null values replaced with last
+        /// the value seen in the sequence.
+        /// </summary>
+        /// <param name="source">The source sequence.</param>
+        /// <typeparam name="T">Type of the elements in the source sequence.</typeparam>
+        /// <returns>
+        /// An <see cref="IEnumerable{T}"/> with missing values replaced.
+        /// </returns>
+        /// <remarks>
+        /// This method uses deferred execution semantics and streams its
+        /// results. If values are null at the start of the sequence then
+        /// they remain missing.
+        /// </remarks>
+
+        public static IEnumerable<T?> Locf<T>(this IEnumerable<T?> source)
+            where T : struct
+        {
+            return source.Locf(e => e == null);
+        }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


atifaziz:
Before anyone objects on the name, `Locf` is named after [`na.locf`](http://svitsrv25.epfl.ch/R-doc/library/zoo/html/na.locf.html) from R and because `LastObservationCarryForward` just looked ugly.

leandromoh:
interesting.. what is the pratical use of this?

MoreLinq/Locf.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <returns>
+        /// An <see cref="IEnumerable{T}"/> with missing values replaced.
+        /// </returns>
+        /// <remarks>
+        /// This method uses deferred execution semantics and streams its
+        /// results. If values are null at the start of the sequence then
+        /// they remain missing.
+        /// </remarks>
+
+        public static IEnumerable<T?> Locf<T>(this IEnumerable<T?> source)
```
fsateler: Hmmm. Thinking out loud here. Maybe there could be a version that returns an `IEnumerable<T>` (ie, not nullable)? In that case, what should happen with leading `null`s? One could add a `seed` parameter, or error out. I would think the latter would be better as it is trival to prepend a seed using `Prepend`.

MoreLinq/Locf.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        public static IEnumerable<T> Locf<T>(this IEnumerable<T> source, Func<T, bool> predicate)
+        {
+            return LocfImpl(source, predicate);
+        }
+
+        static IEnumerable<T> LocfImpl<T>(IEnumerable<T> source, Func<T, bool> predicate)
+        {
+            if (source == null) throw new ArgumentNullException("source");
+            if (predicate == null) throw new ArgumentNullException("predicate");
```
fsateler: `LocfImpl` is lazy, these checks should be in the wrappers.

fsateler:
While the name is not super nice, alternatives I can think of are not particularly better: `FillWithLast`, `CopyPreviousWhile`...

atifaziz:
@leandromoh Imagine you've been handed a table that in the following way:

| Country | City | Value |
|---------|------|-------|
| UK | London | 12 |
| | Manchester | 34 |
| | Glasgow | 56 | 
| Germany | Munch | 78 |
| | Frankfurt | 87 |
| | Stuttgart | 65 |

Notice how **Country** is not repeated on each row (visual grouping). This is just one example of where `Locf` can be handy to get country to repeat on each row. That way each row is self-contained and operations like grouping on country become simple because you know have it available in the row to use as a key. There are other cases (all of which pretty much amount to filling gaps) but hopefully the above illustrates one practical example.

@fsateler Thanks for your thorough review. Always appreciate and will come back to it.

fsateler:
@atifaziz My pleasure :)

BTW, your example made me realize that maybe this function needs generalization with a third parameter: `Func<T, T, T> fillerCreator` , where the parameters are the last 'non-missing' value, the missing value, and it should return the value to be inserted into the list.

With the parameter, it becomes trivial to create a "normalized" table where the `Country` field is always filled:

```csharp
rows.Locf(m => m.Country == null, (lastWithCountry, curRow) => new Row(lastWithCountry.Country, curRow.City, curRow.Value));
```

MoreLinq/Locf.cs
atifaziz: > Why this overload (and the corresponding `where T : class`) in the other overload? Are there efficiency gains from the additional constraints?

No, that's just me being thick and glad someone is there to review my stupidity. 😃

atifaziz:
> While the name is not super nice, alternatives I can think of are not particularly better: `FillWithLast`, `CopyPreviousWhile`...

I'm leaning towards `FillFoward` because _fill_ captures the idea of _filling in the gaps_ and _forward_ is the direction. We lose the bit about _what's_ being filled forward (i.e. previous non-missing) and should perhaps accept that a name alone cannot embody an entire algorithm without being terribly mouthful. `FillFoward` seems to carry the essence and should just stick with time. `Nocb` (#218) would likewise just become `FillBackward`. `SpillForward` and `SpillBackward` could also do the trick as it evokes the idea that elements spill to fill the gaps in a specific direction.

[Why is this so hard?](http://martinfowler.com/bliki/TwoHardThings.html) 🤔 Oh wait, I remember now…

> There are only two hard things in Computer Science: cache invalidation and naming things.
> — Phil Karlton

fsateler:
>> While the name is not super nice, alternatives I can think of are not particularly better: FillWithLast, CopyPreviousWhile...
>
> I'm leaning towards FillFoward

This actually sounds good, I think.

> because fill captures the idea of filling in the gaps and forward is the direction. We lose the bit about what's being filled forward (i.e. previous non-missing)

I think this is not entirely bad, as it allows later adding an overload that allows [creating new elements](https://github.com/morelinq/MoreLINQ/pull/213#issuecomment-260025180).

> and should perhaps accept that a name alone cannot embody an entire algorithm without being terribly mouthful.

True.

> FillFoward seems to carry the essence and should just stick with time. Nocb (#218) would likewise just become FillBackward. SpillForward and SpillBackward could also do the trick as it evokes the idea that elements spill to fill the gaps in a specific direction.

`Fill*` sounds better to me than `Spill*`, FWIW. 

> Why is this so hard? :thinking: Oh wait, I remember now…

>>There are only two hard things in Computer Science: cache invalidation and naming things.
>>— Phil Karlton

:laughing:

atifaziz:
Closed by squashed merge 2772af00f67c9f13356e96318685a1ac0eb8166d.

Issue #217, opened 2016-11-15
FillBackward (next observation carry backward)


atifaziz:
Consider adding a NOCB (Next Observation Carry Backward) operator that replaces some definition of blank (defined by a predicate) in a sequence with the following non-blank.

This would complement LOCF from #177 (PR #213)

```c#
var xs = new int?[] { null, null, 1, 2, null, null, null, null, 3, 4, null };
var ys = xs.Nocb();
// ys will be { 1, 1, 1, 2, 3, 3, 3, 3, 3, 4, null }
```

Note that this _does not_ remove blanks at the end of the input sequence, after the last non-blank value. They can be removed subsequently using `Where`. Another option is to use `Locf` on the output sequence to fill the tail.

fsateler:
Just to understand, this would be equivalent to `src.Reverse().Locf(predicate).Reverse()` ? If so, it would be good to add a test that verifies this property (along with all the other tests).

atifaziz:
> equivalent to `src.Reverse().Locf(predicate).Reverse()` ?

Yes, except it does potentially a lot less buffering & work unless the sequence is full of lots of _holes_.

> would be good to add a test that verifies this property (along with all the other tests).

Definitely. I just wanted to park it out here as RFC and to get some early review.

atifaziz:
@fsateler While I've added a filler overload for `FillForward` in 1a9134742e6ab4043cb6eabe7ebd1fff19dbcc87, I can't seem to come up with a sensible example or use case for one for `FillBackward`. It seems keeping parity with `FillForward` would be an artificial reason to add the overload right away but I'd rather someone comes along and explains how they're missing such an overload and then adding it would be breeze. That is, unless you can already imagine a practical use case from your work?

fsateler:
Well, I see the use case for all overloads of `FillForward` much more strongly than `FillBackward`. It appears I only fill up later values :). So I think it's a good plan to delay the overload until requested.

atifaziz:
> I see the use case for all overloads of `FillForward` much more strongly than `FillBackward`.

Try a sequence lagging with respect to another and you'll see the use cases.

atifaziz:
Closed by ac5f9ff9f996cfa8805e1e49da14cdb404a30169.

PR #218, opened 2016-11-15
Add FillBackward (next observation carry backward)

atifaziz:
See #217 for background and more. Reviews welcome.

~Tests are pending.~

Note that this complements #177 (PR #213) as both `Locf` & `Nocb` can be used to help each other to fill missing element at the head and the tail of the source sequence.

- [x] Tests
- [x] Add to operators in package description and release notes
- [x] Add to operators list in README


atifaziz:
Squashed and merged in ac5f9ff9f996cfa8805e1e49da14cdb404a30169.

Issue #227, opened 2017-01-02
Replace ExpectedException with Assert.Throws in unit tests


atifaziz:
[`ExpectedException` has been removed starting with NUnit 3](https://github.com/nunit/nunit/issues/113). Unit tests need to be adjusted to use `Assert.Throws` instead, which is the recommended and better way.

Meanwhile #226 adds a local `[ExpectedException]` so tests using that attribute can compile and run unchanged with NUnit 3. The local `[ExpectedException]` implementation should be removed when migration to using `Assert.Throws` is complete.

Test files that need addressing are:

- [ ] `AcquireTest.cs`
- [ ] `AssertCountTest.cs`
- [ ] `AssertTest.cs`
- [ ] `AtLeastTest.cs`
- [ ] `BatchTest.cs`
- [ ] `CartesianTest.cs`
- [ ] `ConcatTest.cs`
- [ ] `ConsumeTest.cs`
- [ ] `CountByTest.cs`
- [ ] `DistinctByTest.cs`
- [ ] `EndsWithTest.cs`
- [ ] `EquiZipTest.cs`
- [ ] `ExceptByTest.cs`
- [ ] `ExcludeTest.cs`
- [ ] `FoldTest.cs`
- [ ] `ForEachTest.cs`
- [ ] `FullGroupJoinTest.cs`
- [ ] `GenerateTest.cs`
- [ ] `GroupAdjacentTest.cs`
- [ ] `IncrementalTest.cs`
- [ ] `IndexTest.cs`
- [ ] `InterleaveTest.cs`
- [ ] `LagTest.cs`
- [ ] `LeadTest.cs`
- [ ] `MaxByTest.cs`
- [ ] `MinByTest.cs`
- [ ] `NestedLoopTest.cs`
- [ ] `PadTest.cs`
- [ ] `PairwiseTest.cs`
- [ ] `PartitionTest.cs`
- [ ] `PermutationsTest.cs`
- [ ] `PipeTest.cs`
- [ ] `PreScanTest.cs`
- [ ] `PrependTest.cs`
- [ ] `RandomSubsetTest.cs`
- [ ] `RandomTest.cs`
- [ ] `RepeatTest.cs`
- [ ] `RunLengthEncodeTest.cs`
- [ ] `SegmentTest.cs`
- [ ] `SingleOrFallbackTest.cs`
- [ ] `SkipUntilTest.cs`
- [ ] `SortedMergeTest.cs`
[...]

PR #233, opened 2017-01-06
Fixed issue #227 Replaced ExpectedException with Assert.Throws in unit tests

GeorgeVovos:
Replaced ExpectedException with Assert.Throws in unit tests so the code is compatible with NUnit 3


atifaziz:
Thanks for this PR.

I'm curious, did you make the all the changes by hand or did you use some sort of automated process? If it was the former, I don't want to imagine the labour. :)

There seem to be more changes than expected. There are a large number of files to review, and for review purposes, it helps to focus on just the expected changes. Take [`AcquireTest.cs`](https://github.com/morelinq/MoreLINQ/blob/2772af00f67c9f13356e96318685a1ac0eb8166d/MoreLinq.Test/AcquireTest.cs), for example. There is only [one test with `[ExpectedException]`](https://github.com/morelinq/MoreLINQ/blob/2772af00f67c9f13356e96318685a1ac0eb8166d/MoreLinq.Test/AcquireTest.cs#L27-L32) but then [the diff](https://github.com/morelinq/MoreLINQ/pull/233/commits/c5ee637e5a3bfcac2023234fbf3a987806d82c08#diff-f64a1b8e175b1dbce40a4273e9c7c44f) contains lots more changes. Granted that they are harmless whitespace and styling changes but they do add noise that one has to mentally filter out. Would it be too much trouble to remove such noise? I'm hoping if you used an automated process then you can adjust it.

GeorgeVovos:
Yes,by hand
It didn't take long,about an hour.
In a similar case in the past ,where I had to wrap a code sample with some else, I created a custom code snippet (like the try/catch in VS) but I was too lazy to do that this time...

I saw that the diff tool here shows many changes but the diff tool in my Visual Studio shows only **my**  changes and nothing else*
So, unfortunately I can't fix this...


If you want, you can try the following:
In your Visual Studio take my branch and your master.  
Then,locally merge them **but don't commit the changes**
In the pending changes window ,if you VS works like mine, you should see only **my** changes


I only saw 1 deleted whitespace in AcquireTest.cs in 
`class ApplicationException : Exception {}`

atifaziz:
So it seems that `AcquireTest.cs` was the one exception with those extra change and so I may have blown the whistle too soon assuming the rest of the files may suffer from the same issue.

> but the diff tool in my Visual Studio shows only my changes and nothing else

I think VS diff might be configured by default not to highlight certain whitespace changes.

MoreLinq.Test/CartesianTest.cs
```diff
@@ -24,33 +24,39 @@ public void TestCartesianIsLazy()
...
         /// </summary>
         [Test]
-        [ExpectedException(typeof(ArgumentNullException))]
         public void TestCartesianSequenceBNullException()
         {
-            var sequence = Enumerable.Repeat(1,10);
-            sequence.Cartesian<int,int,int>(null, (a, b) => a + b);
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                var sequence = Enumerable.Repeat(1, 10);
```
atifaziz: This line should be moved outside of the lambda to `Assert.Throws` as we want to only test that `Cartesian` throws.

MoreLinq.Test/CartesianTest.cs
```diff
@@ -24,33 +24,39 @@ public void TestCartesianIsLazy()
...
         /// </summary>
         [Test]
-        [ExpectedException(typeof(ArgumentNullException))]
         public void TestCartesianResultSelectorNullException()
         {
-            var sequence = Enumerable.Repeat(1, 10);
-            sequence.Cartesian(sequence, (Func<int, int, int>) null);
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                var sequence = Enumerable.Repeat(1, 10);
```
atifaziz: This line should be moved outside of the lambda to `Assert.Throws` as we want to only test that `Cartesian` throws.

MoreLinq.Test/EquiZipTest.cs
```diff
@@ -73,43 +73,53 @@ public void ZipWithEqualLengthSequencesFailStrategy()
...
         public void ZipWithFirstSequenceShorterThanSecondFailStrategy()
         {
-            var zipped = new[] { 1, 2 }.EquiZip(new[] { 4, 5, 6 }, Tuple);
-            Assert.That(zipped, Is.Not.Null);
-            zipped.Consume();
+            Assert.Throws<InvalidOperationException>(() =>
+            {
+                var zipped = new[] { 1, 2 }.EquiZip(new[] { 4, 5, 6 }, Tuple);
+                Assert.That(zipped, Is.Not.Null);
+                zipped.Consume();
```
atifaziz: Only this line should be inside the lambda to `Assert.Throws` as that's what will cause and being tested for an exception.

MoreLinq.Test/EquiZipTest.cs
```diff
@@ -73,43 +73,53 @@ public void ZipWithEqualLengthSequencesFailStrategy()
...
         {
-            var zipped = new[] { 1, 2, 3 }.EquiZip(new[] { 4, 5 }, Tuple);
-            Assert.That(zipped, Is.Not.Null);
-            zipped.Consume();
-            zipped.AssertSequenceEqual(Tuple(1, 4), Tuple(2, 5));
+            Assert.Throws<InvalidOperationException>(() =>
+            {
+                var zipped = new[] { 1, 2, 3 }.EquiZip(new[] { 4, 5 }, Tuple);
+                Assert.That(zipped, Is.Not.Null);
+                zipped.Consume();
```
atifaziz: Only this line should be inside the lambda to `Assert.Throws` as that's what will cause and being tested for an exception.

MoreLinq.Test/EquiZipTest.cs
```diff
@@ -73,43 +73,53 @@ public void ZipWithEqualLengthSequencesFailStrategy()
...
+                zipped.AssertSequenceEqual(Tuple(1, 4), Tuple(2, 5));
```
atifaziz: This is strange as code execution should never get here, having thrown on the previous line. If so, it should be simply deleted.

MoreLinq.Test/InterleaveTest.cs
```diff
@@ -24,35 +24,41 @@ public void TestInterleaveIsLazy()
...
         public void TestInterleaveNullOtherSequencesArgument()
         {
-            const int count = 10;
-            var sequence = Enumerable.Range(1, count);
-            sequence.Interleave(null);
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                const int count = 10;
+                var sequence = Enumerable.Range(1, count);
+                sequence.Interleave(null);
```
atifaziz: Only this line should be inside the lambda to `Assert.Throws` as that's what will cause and being tested for an exception.

MoreLinq.Test/InterleaveTest.cs
```diff
@@ -24,35 +24,41 @@ public void TestInterleaveIsLazy()
...
         public void TestInterleaveNullEntryInOtherSequences()
         {
-            const int count = 10;
-            var sequence = Enumerable.Range(1, count);
-            sequence.Interleave(Enumerable.Range(1, count), null);
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                const int count = 10;
+                var sequence = Enumerable.Range(1, count);
+                sequence.Interleave(Enumerable.Range(1, count), null);
```
atifaziz: Only this line should be inside the lambda to `Assert.Throws` as that's what will cause and being tested for an exception.

MoreLinq.Test/RandomSubsetTest.cs
```diff
@@ -25,42 +25,50 @@ public void TestRandomSubsetIsLazy()
...
         /// Verify that involving RandomSubsets with a subset size less than 0 results in an exception.
         /// </summary>
         [Test]
-        [ExpectedException(typeof(ArgumentOutOfRangeException))]
         public void TestRandomSubsetNegativeSubsetSize()
         {
-            Enumerable.Range(1, 10).RandomSubset(-5);
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                Enumerable.Range(1, 10).RandomSubset(-5);
```
atifaziz: The call to `Range` should go outside the lambda to `Assert.Throws` and the one really under test for an exception, `RandomSubset`, should go inside. Otherwise, there is the risk of a false positive since we don't know if `Range` or `RandomSubset` threw `ArgumentOutOfRangeException`.

MoreLinq.Test/RandomSubsetTest.cs
```diff
@@ -25,42 +25,50 @@ public void TestRandomSubsetIsLazy()
...
         /// Verify that involving RandomSubsets with a subset size less than 0 results in an exception.
         /// </summary>
         [Test]
-        [ExpectedException(typeof(ArgumentOutOfRangeException))]
         public void TestRandomSubsetNegativeSubsetSize2()
         {
-            Enumerable.Range(1, 10).RandomSubset(-1, new Random());
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                Enumerable.Range(1, 10).RandomSubset(-1, new Random());
```
atifaziz: The call to `Range` should go outside the lambda to `Assert.Throws` and the one really under test for an exception, `RandomSubset`, should go inside. Otherwise, there is the risk of a false positive since we don't know if `Range` or `RandomSubset` threw `ArgumentOutOfRangeException`.

MoreLinq.Test/RandomSubsetTest.cs
```diff
@@ -113,29 +121,33 @@ public void TestRandomSubsetShorterThanSequence()
...
-            const int subsetSize = count + 5;
-            var sequence = Enumerable.Range(1, count);
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                const int count = 100;
+                const int subsetSize = count + 5;
+                var sequence = Enumerable.Range(1, count);
 
-            sequence.RandomSubset(subsetSize).Count();
+                sequence.RandomSubset(subsetSize).Count();
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/RandomSubsetTest.cs
```diff
@@ -113,29 +121,33 @@ public void TestRandomSubsetShorterThanSequence()
...
-            const int subsetSize = count + 5;
-            var sequence = Enumerable.Range(1, count);
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                const int count = 100;
+                const int subsetSize = count + 5;
+                var sequence = Enumerable.Range(1, count);
 
-            sequence.RandomSubset(subsetSize, new Random(1234)).Count();
+                sequence.RandomSubset(subsetSize, new Random(1234)).Count();
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/RandomTest.cs
```diff
@@ -17,74 +17,86 @@ public class RandomTest
...
         {
-            const int maxValue = 10;
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                const int maxValue = 10;
 
-            Assert.Greater( maxValue, 0 );
-            MoreEnumerable.Random(null, maxValue);
+                Assert.Greater(maxValue, 0);
+                MoreEnumerable.Random(null, maxValue);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/RandomTest.cs
```diff
@@ -17,74 +17,86 @@ public class RandomTest
...
-
-            Assert.LessOrEqual(minValue, maxValue);
-            MoreEnumerable.Random(null, minValue, maxValue);
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                const int minValue = 10;
+                const int maxValue = 100;
+
+                Assert.LessOrEqual(minValue, maxValue);
+                MoreEnumerable.Random(null, minValue, maxValue);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/RandomTest.cs
```diff
@@ -17,74 +17,86 @@ public class RandomTest
...
         public void TestNegativeMaxValueException()
         {
-            const int maxValue = -10;
-            Assert.Less( maxValue, 0 );
-            MoreEnumerable.Random(maxValue);
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                const int maxValue = -10;
+                Assert.Less(maxValue, 0);
+                MoreEnumerable.Random(maxValue);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/RandomTest.cs
```diff
@@ -17,74 +17,86 @@ public class RandomTest
...
-
-            Assert.Greater(minValue, maxValue);
-            MoreEnumerable.Random(minValue, maxValue);
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                const int minValue = 100;
+                const int maxValue = 10;
+
+                Assert.Greater(minValue, maxValue);
+                MoreEnumerable.Random(minValue, maxValue);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/SubsetTest.cs
```diff
@@ -24,25 +24,29 @@ public void TestSubsetsIsLazy()
...
         public void TestNegativeSubsetSize()
         {
-            const int count = 10;
-            var sequence = Enumerable.Range(1, count);
-            sequence.Subsets(-5);
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                const int count = 10;
+                var sequence = Enumerable.Range(1, count);
+                sequence.Subsets(-5);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/SubsetTest.cs
```diff
@@ -24,25 +24,29 @@ public void TestSubsetsIsLazy()
...
         {
-            const int count = 10;
-            var sequence = Enumerable.Range(1, count);
-            var result = sequence.Subsets(count + 5);
-            result.Count(); // this particular exception is deferred until sequence evaluation
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                const int count = 10;
+                var sequence = Enumerable.Range(1, count);
+                var result = sequence.Subsets(count + 5);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/SubsetTest.cs
```diff
@@ -24,25 +24,29 @@ public void TestSubsetsIsLazy()
...
+                result.Count(); // this particular exception is deferred until sequence evaluation
```
atifaziz: This line doesn't make sense and should be reviewed for deletion. An argument exception should be caught eagerly during the call to `Subsets` and not when it is iterated. To be checked and removed.

MoreLinq.Test/ToDataTableTest.cs
```diff
@@ -72,70 +72,86 @@ public ToDataTableTest()
...
         {
-            var dt = new DataTable();
-            dt.Columns.Add("Test");
+            Assert.Throws<ArgumentException>(() =>
+            {
+                var dt = new DataTable();
+                dt.Columns.Add("Test");
 
-            m_TestObjects.ToDataTable(dt);
+                m_TestObjects.ToDataTable(dt);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/ToDataTableTest.cs
```diff
@@ -72,70 +72,86 @@ public ToDataTableTest()
...
         {
-            var dt = new DataTable();
-            dt.Columns.Add("AString", typeof(int));
+            Assert.Throws<ArgumentException>(() =>
+            {
+                var dt = new DataTable();
+                dt.Columns.Add("AString", typeof(int));
 
-            m_TestObjects.ToDataTable(dt, t=>t.AString);
+                m_TestObjects.ToDataTable(dt, t => t.AString);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/WindowedTest.cs
```diff
@@ -24,22 +24,26 @@ public void TestSlidingWindowIsLazy()
...
         [Test]
-        [ExpectedException(typeof(ArgumentOutOfRangeException))]
         public void TestSlidingWindowNegativeWindowSizeException()
         {
-            var sequence = Enumerable.Repeat(1, 10);
-            sequence.Windowed(-5);
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                var sequence = Enumerable.Repeat(1, 10);
+                sequence.Windowed(-5);
```
atifaziz: Only this line should be inside the lambda to Assert.Throws as that's what will cause and being tested for an exception.

MoreLinq.Test/AcquireTest.cs
```diff
@@ -37,15 +39,15 @@ public void AcquireAll()
             Disposable a = null;
             Disposable b = null;
             Disposable c = null;
-
```
atifaziz: Remove unrelated style & whitespace changes in this file after this line.

atifaziz:
`Assert.Throws` is more powerful than `ExpectedException` in that it allows us to assert further things about the thrown exception. Where available, the parameter names should be asserted too, as in:

```c#
        [Test]
        public void AcquireNullSequence()
        {
            var e = Assert.Throws<ArgumentNullException>(() =>
                MoreEnumerable.Acquire<IDisposable>(null));
            Assert.That(e.ParamName, Is.EqualTo("source"));
        }
```

The value of doing this is more pronounced in cases where multiple arguments can throw the same exception type.

Technically, this could be made into a separate PR unless you're up to tackling it as part of this refactoring?

GeorgeVovos:
- I updated Assert.Throw to test only 1 statement
- I still haven't managed to fix the whitespaces
- I haven't update SubsetTest.cs with
    This line doesn't make sense and should be reviewed for deletion. 
    An argument exception should be caught eagerly during the call to Subsets and not when it is iterated. 
    To be checked and removed.
	

About the validation of ArgumentException
Which style do you prefer
Classic:

 ```
var e = Assert.Throws<ArgumentNullException>(() =>
                  {
                     MoreEnumerable.Acquire<IDisposable>(null);
                  });
         Assert.That(e.ParamName, Is.EqualTo("source"));
```

Or something more "fluent"		  

 ```
Assert.Throws<ArgumentNullException>(() =>
          {
                MoreEnumerable.Acquire<IDisposable>(null);
          })
          .AndParameterNameIs("source");
				
```

atifaziz:
So `AndParameterNameIs` is something you'd add, right?

If you want to go for a fluent style then I'd say better to embody the whole thing into a single assertion:

```c#
Assert.Throws<ArgumentNullException>("source", () =>
    MoreEnumerable.Acquire<IDisposable>(null));
```

Otherwise the classic is just fine and what everyone will be familiar with.

Since it looks like you're going to also assert on parameter names (and thanks for that), let's also remove the block ceremony on single-statement lambdas. That is, instead of:

```c#
Assert.Throws<ArgumentNullException>(() =>
{
    MoreEnumerable.Acquire<IDisposable>(null);
});
```

let's go with:

```c#
Assert.Throws<ArgumentNullException>(() =>
    MoreEnumerable.Acquire<IDisposable>(null));
```

Note that lambda body is indented. And should you choose to go with the classic approach:

```c#
var e = Assert.Throws<ArgumentNullException>(() =>
    MoreEnumerable.Acquire<IDisposable>(null));
Assert.That(e.ParamName, Is.EqualTo("source"));
```

> I still haven't managed to fix the whitespaces

You should be able to see these by running a `git diff` on the command-line and in the directory.

GeorgeVovos:
_So AndParameterNameIs is something you'd add, right?_
It was going to be an Extension method on ArgumentException.

_If you want to go for a fluent style then I'd say better to embody the whole thing into a single assertion_
This is a nice idea but unfortunately we can't add Extension methods on static classes...
If we want to go with the single assertion i can only think 2 solutions

1. Introduce our own static Assert class (Something like MoreAssert ) and add a Throw method there
2. Introduce a base Test class where we can add a AssertThrowsArgumentNullException method

of course option 3 is the "classic" approach

The only thing I like with the non classic solutions is that we don't have to introduce a new variable and we don't have to repeat  `Assert.That(e.ParamName, Is.EqualTo("XYZ")); `everywhere

Whatever you like more,I really don't mind

atifaziz:
> This is a nice idea but unfortunately we can't add Extension methods on static classes

`Assert` is not static. I believe you can derive and static methods and is the recommended way to extend `Assert` (or used to be).

GeorgeVovos:
_Assert is not static._
Damn!!!You are right,I was under the impression it was static for so many years...
OK I'll create a derived class


```
    public class AssertException : Assert
    {
        public static void ThrowsArgumentNull(string argumentName,TestDelegate code)
        {
            ThrowsArgumentNullException<ArgumentNullException>(argumentName, code);
        }

        public static void ThrowsArgument(string argumentName, TestDelegate code)
        {
            ThrowsArgumentNullException<ArgumentException>(argumentName, code);
        }

        public static void ThrowsArgumentNullException<TActual>(string argumentName,TestDelegate code) where TActual : ArgumentException
        {
            var e = Throws<TActual>(code);

            That(e.ParamName, Is.EqualTo(argumentName));
        }
    }
```

Do the name look funny or long to you?Do you have any suggestions?
It will be used like this:
```
AssertException.ThrowsArgumentNull("predicate",() => 
{
      new object[0].Assert(null);
});
```

MoreLinq.Test/AssertException.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            ThrowsArgumentException<ArgumentException>(argumentName, code);
+        }
+
+        public static void OutOfRange(string argumentName, TestDelegate code)
+        {
+            ThrowsArgumentException<ArgumentOutOfRangeException>(argumentName, code);
+        }
+
+        public static void InvalidOperation(TestDelegate code)
```
atifaziz: I don't find it's worth wrapping every exception like `InvalidOperationException`, `SequenceException` and `Exception`. These should be reverted back to simply using `Assert.Throws`.

GeorgeVovos:
OK,we should be close...
Note

- I deleted the ExpectedException class

- I haven't updated SubsetTest.cs with
This line doesn't make sense and should be reviewed for deletion.
An argument exception should be caught eagerly during the call to Subsets and not when it is iterated.

atifaziz:
Squash-merged in 07a3420b48ef2af28295cb21f794870b763cffb7.

@GeorgeVovos Thanks for your help with this and working through the review items.

PR #243, opened 2017-01-30
Append-method as counterpart to the existing prepend-method.

Timmitry:
I was wondering why there isn't an append-method while the prepend-method is supplied. I created the append-method by mostly copying and pasting the prepend method, including unit tests.

### DIFF
--- a/MoreLinq.Test/AppendTest.cs
+++ b/MoreLinq.Test/AppendTest.cs
@@ -32,7 +32,7 @@ namespace MoreLinq.Test
         }
 
         [Test]
-        public void AppendWithEmptyTailSequence()
+        public void AppendWithEmptyHeadSequence()
         {
             string[] head = { };
             var tail = "first";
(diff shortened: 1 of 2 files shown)


atifaziz:
@Timmitry It's already there and called [`Concat`](https://morelinq.github.io/1.x/ref/api/html/M_MoreLinq_MoreEnumerable_Concat__1.htm). I wish you had checked in first by opening an issue and proposing the idea. It would saved you duplicating the work and the trouble of submitting a PR.

Timmitry:
Ah, right, there is an overload for `Concat`, I was only looking for the method `Append`...
You're right, I should first open an issue in the future. However, since I'm new to Git and Github, this was a good exercise, nonetheless.

atifaziz:
BTW, [`Append` is available in LINQ to Objects](https://github.com/dotnet/corefx/blob/33e00e186bfad0041b63da382bbcb014f26d2d70/src/System.Linq/src/System/Linq/AppendPrepend.cs#L12-L26) when targeting .NET Core so I'm guessing it's coming to the full .NET Framework soon.

Issue #231, opened 2017-01-04
Continuous build for non-Windows platforms


atifaziz:

fsateler:
I finally found some time to look into this. Installing dotnet core in my debian machine proved hard (binaries are only good for debian stable, not debian unstable, which is what I use). Therefore I went the easy way and installed the [docker image](https://hub.docker.com/r/microsoft/dotnet/).

1. The dotnet runtime as shipped in the latest image appears to not support `netcoreapp1.0`, but rather `netcoreapp1.1`. Therefore the test project fails to build. I will try now with the 1.0.3 image, but this suggests that maintaining the required versions will not be trivial. Why doesn't the 1.1 runtime satisfy `netcoreapp1.0`? Is this something that should be reported upstream?
2. Running `dotnet build` tries to build all possible configurations, which include the full framework and portable. The first is easily fixed by installing the [official xamarin mono runtime](http://www.mono-project.com/docs/getting-started/install/linux/) (note to self, one needs to enable the gdiplus repository too), but the portable runtime appears to be proprietary and thus not freely available. I don't think extracting the relevant dlls from a windows installer is reasonable to perform during CI, so we probably will need to disable Portable.Net during linux builds.

Sorry I've been slow on this, I hope I can make more progress soonish.

fsateler:
Doh, as I re-read the install page I noticed there is a `referenceassemblies-pcl` package, which should fix the Portable.Net problem.

atifaziz:
> Why doesn't the 1.1 runtime satisfy `netcoreapp1.0`?

From what I understand, they are side-by-side installations that don't interfere with each other anymore:

>  .NET Core is entirely side-by-side, with each installation completely isolated from other installations. You can now be even more flexible in how you deploy systems to machines, and be reassured that introducing another version of .NET Core won't break your other applications. Flexibility is the name of the game here.
>
> Source: [Side by side installation](https://github.com/dotnet/features/blob/6f7bd19b65bc16c4d89a4e26768f3664b3489f68/features/netcore/side-by-side.md)

So you'll need to work with the image for 1.0.

fsateler:
> From what I understand, they are side-by-side installations that don't interfere with each other anymore

OK, I see. I wonder how one is intended to manage the `dotnet` tool itself though.

atifaziz:
The tooling, like `dotnet`, is still in preview so I'm guessing the exact details on the servicing story will be revealed with the release.

[...]

PR #244, opened 2017-02-01
Add travis-ci configuration to test in a linux environment

fsateler:
Fixes: #231 

--

Results can be seen at https://travis-ci.org/fsateler/MoreLINQ  (s/fsateler/morelinq/ when configured for this repository).


travis-test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -e
+
+dotnet test -c Debug -f netcoreapp1.0 MoreLinq.Test
```
atifaziz: Isn't this a duplicate of the 3rd instruction (`dotnet test -f netcoreapp1.0 MoreLinq.Test`) under `script:` section of `.travis.yml'?

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+mono: latest
+dist: trusty
+sudo: required
+dotnet: 1.0.0-preview2-003131
+
+install:
+  - dotnet restore
+
+script:
+  - dotnet build -c Debug MoreLinq
```
atifaziz: Any reason to exclude release builds?

travis-test.sh
```diff
@@ -0,0 +1,9 @@
...
+RID=`dotnet --info | grep RID | sed -E 's/ *RID: *(.+)/\1/'`
```
atifaziz: [I used the same technique](https://github.com/morelinq/MoreLINQ/blob/c9e3c09b738ce162d4a90e8d16edcba4732ea46d/test.sh#L6) at first but it didn't work out well on Windows. While `dotnet --info` reported an RID of `win10-x64` or `win10-x86` on Windows 10, the test project was still built to a directory based on the RID of Windows 7. I ended up hard-coding it first in 7c0a0f9f03de696c9feb010adbc296c45ad586e8 (:speak_no_evil:) before picking it up dynamically in 7b8551c5b4bd30d6e3f67fd5d456b8ee7f4f1ec6 so it works for `win7-x64` or `win7-x86` depending on the bit-ness of the OS.

atifaziz: …point being, it seems to work for non-Windows platforms if you rely on the RID reported by `dotnet --info` but it could be that it might have to be eventually picked up dynamically by just scanning the directory below `net451`. Meanwhile, I'm hoping this is just a kink due to the tooling being preview.

travis-test.sh
fsateler: Indeed! I'll remove that part from the script and rename the script to `travis-mono-test.sh`. The only reason for its existence is that I did not manage to properly encode the commands in yaml syntax

.travis.yml
fsateler: Not really. I could add both if you want to. I did not manage to test the Release configuration though. `dotnet` insisted on trying to use the debug build.

travis-test.sh
fsateler: I could switch it to iterate over all values:

```sh
for dir in MoreLinq.Test/bin/Debug/net451/*/ ; do
    mono $dir/dotnet-test-nunit.exe $dir/MoreLinq.Test.dll
done
```

This would be future proof against building multiple times.

.travis.yml
atifaziz: > I could add both if you want to.

I'd say it's even more important to test release but if it's not too much trouble to have both then we should strive for it.

> `dotnet` insisted on trying to use the debug build.

Really? 😮 I guess we could leave it as a separate issue to be investigated?

fsateler: > > dotnet insisted on trying to use the debug build.
>
>Really? :open_mouth: I guess we could leave it as a separate issue to be investigated?

Hmm. This works on my local machine but did not when I tried on travis. Let's try again to rule out typos. Otherwise it would look like a tooling bug (as you can see from the yml file, the tooling version is not the latest one).

fsateler: ....And yes, it appears it was a typo :speak_no_evil: . Now we build and test debug and release configurations.

travis-mono-test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -e
+set -x
+
+for dir in MoreLinq.Test/bin/$CONFIGURATION/net451/*/ ; do
+	mono $dir/dotnet-test-nunit.exe $dir/MoreLinq.Test.dll
```
atifaziz: This adds a double slash here, as in:

    mono MoreLinq.Test/bin/Release/net451/ubuntu.14.04-x64//dotnet-test-nunit.exe \
         MoreLinq.Test/bin/Release/net451/ubuntu.14.04-x64//MoreLinq.Test.dll

Threw me off when searching through the log. Would it be nitpicking to trim the trailing slash using `{$dir%/}`, like below?

    mono ${dir%/}/dotnet-test-nunit.exe ${dir%/}/MoreLinq.Test.dll

And would it be even more nitpicking to replace the leading tab on that line with spaces instead?

fsateler: No problem. However, instead of using `${dir/%}` I used `${dir}dotnet-test-nunit.exe`. I find shell variable expansion to be hard to read (I don't use shell programming everyday so I tend to forget what does each modifier does), so I prefer to avoid it if there is a more readable alternative. Moreover, I'm not sure if this is a `bash` extension or part of POSIX shell (although debian's `dash` seems to allow this particular expansion).

atifaziz: > I'm not sure if this is a `bash` extension or part of POSIX shell

Yeah, I get lost in that too.

atifaziz:
Thanks for helping with this.

PR #250, opened 2017-02-13
travis: Build on mac too

fsateler:
This way [the badge text](https://github.com/morelinq/morelinq.github.io/blob/master/index.html#L25) becomes true :smile:


atifaziz:
Thanks for keeping us honest! 😄

PR #252, opened 2017-02-19
CompareCount: compares lengths of sequences

jnyrup:
These three overloads enable comparing lengths of two sequences.
* If *both* sequences implements `ICollection<T>` the `Count` properties will be compared.
* If at least one sequence implements `ICollection<T>`, the current versions of `AtLeast`, `AtMost` and `Exactly` will be used.
* Otherwise, the sequences will be iterated as long as both has more elements.


leandromoh:
I dont look the tests yet, but the implementation seems great! 👍

atifaziz:
@jnyrup How is this any better than sending in the count of the second sequence?

jnyrup:
@leandromoh The tests are all the current tests extended with all four combinations of collections and lazy sequences as `first` and `second` sequences.

@atifaziz This takes care of whether one or both of the sequences implements `ICollection` and uses `Count` when possible regardless of the order of `first` and `second`.
When neither `first` nor `second` implements `ICollection`, `first` has fewer elements than `second` and `AtMost(first, second)` is called, this stops iterating `second` earlier than passing `second.Count()` to the current `AtMost(first, int)`.

leandromoh:
I was thinking and if user what to do something like the following, it can't:

```
IEnumerable<T1> enumerable1;
IEnumerable<T2> enumerable2;
...
if (...) // if two sequence has same length
{
}
else if (..)// if enumerable1 is longer
{
}
else // if enumerable2 is longer
{
}
```

to solve this, we can create the follow function in the spirit of [IComparable](https://msdn.microsoft.com/pt-br/library/system.icomparable(v=vs.110).aspx)

- `int CompareCount<T1, T2>(this IEnumerable<T1> first, IEnumerable<T2> second)`

and so, we can define `AtLeast`, `AtMost` and `Exactly` in terms of `CompareCount`:

```
public static bool AtLeast<T>(this IEnumerable<T> first, IEnumerable<T> second)
{
	return first.CompareCount(second) >= 0;
}

public static bool AtMost<T>(this IEnumerable<T> first, IEnumerable<T> second)
{
	return first.CompareCount(second) <= 0;
}

public static bool Exactly<T>(this IEnumerable<T> first, IEnumerable<T> second)
{
	return first.CompareCount(second) == 0;
}
```
`CompareCount` source code can be something like:

```
public static int CompareCount<T1, T2>(this IEnumerable<T1> first, IEnumerable<T2> second)
{
	if (first == null) throw new ArgumentNullException(nameof(first));
	if (second == null) throw new ArgumentNullException(nameof(second));

	var firstCol = first as ICollection<T1>;
	var secondCol = second as ICollection<T2>;

	if (firstCol != null)
	{
		if (secondCol != null)
		{
			return CompareLength(firstCol.Count, secondCol.Count);
		}

		return CompareLength(firstCol.Count, QuantityIterator(second, firstCol.Count + 1));
	}

	if (secondCol != null)
	{
		return CompareLength(QuantityIterator(first, secondCol.Count + 1), secondCol.Count);
	}

	bool firstHasNext, secondHasNext;

	using (var e1 = first.GetEnumerator())
	using (var e2 = second.GetEnumerator())
	{
		do
		{
			firstHasNext = e1.MoveNext();
			secondHasNext = e2.MoveNext();
		}
		while (firstHasNext && secondHasNext);
	}

	return CompareLength(Convert.ToInt32(firstHasNext), Convert.ToInt32(secondHasNext));
}

private static int CompareLength(int first, int second)
{
	return first == second ? 0 : first > second ? 1 : -1;
}

private static int QuantityIterator<T>(IEnumerable<T> source, int limit)
{
	var count = 0;

	using (var e = source.GetEnumerator())
	{
		while (e.MoveNext())
		{
			if (++count == limit)
			{
				break;
			}
		}
	}

	return count;
}
```

jnyrup:
@leandromoh That's beautiful. I really like that `CompareCount` does not require the two sequence to be of same type.
I was to thinking about a CompareTo based approach, but apparently I forgot it. Is there a special reason why you are using `CompareLength` instead of just `int.CompareTo`?

leandromoh:
> I really like that CompareCount does not require the two sequence to be of same type.

Yes, they really dont need, neither the `AtLeast`, `AtMost`, etc overloads.

> Is there a special reason why you are using CompareLength instead of just int.CompareTo?

No, i just forgot about its existence. LOL

jnyrup:
I've updated the PR with Leandro's implementation.

leandromoh:
@jnyrup did you add tests for `CompareCount`?

jnyrup:
@leandromoh Now I have. 
I've also updated readme.md and project.json with `CompareCount`.
I haven't updated readme.md with the new "overloads" for `AtLeast`, `AtMost` and `Exactly` as I'm unsure whether they would fit under the "The methods has x overloads" description.

leandromoh:
@jnyrup great job! the PR looks good for me! 
I think it fit under the  "The methods has x overloads" description. 
But @atifaziz can say with more certainty

fsateler:
I'm not sure I like overloading the names `AtLeast`, `AtMost` and `Exactly` for sequences. `seq1.Exactly(seq2)` suggests the sequences are equal but only the lengths are being compared. Using a name that makes this explicit  (`AtLeastAsLongAs` is what I can come up with, but I'm not entirely happy with that either as its too long) would make usage more self-explanatory.

leandromoh:
What do you think to add the "CountOf" suffix? 

seq 1.AtLeastCountOf(seq2);
seq 1.AtMostCountOf(seq2);
seq 1.ExactlyCountOf(seq2);

Sounds self-explanatory to me.

jnyrup:
What about using SameCountAs as suffix?
Without being a native English speaker, I think it sounds more natural.

* `seq1.AtLeastSameCountAs(seq2)`
* `seq1.AtMostSameCountAs(seq2)`
* `seq1.ExactlySameCountAs(seq2)`

leandromoh:
Maybe this PR should just add only `CompareCount`, since the other methods (AtLeastxxx, AtMostxxx, Exactlyxxx) just calls it and compare its result against zero. That is, they are trivial comparations with `CompareCount` in hands. What do you think?

jnyrup:
As no one has come up with really satisfying names for the AtLeast/AtMost/Exactly I agree it is better to leave them out as you suggest. 
As you point out they are trivial comparisons and as the return value of `CompareCount` is similar to other `CompareTo`'s the comparison should not be unfamiliar.

leandromoh:
@atifaziz @fsateler what do you think ??

atifaziz:
@jnyrup @leandromoh I still struggle with understanding the utility of this. If you're comparing lengths of two sequences then it's usually because you want to make some decision based on the result of the comparison and which would still involve either of the sequences going forward. At that stage, you'll be faced with the problem multiple iterations, especially if the sequences are not cheap. To avoid that, I usually buffer the results in advance but then I have access to either `Length` of the array or `Count` of the list; why would I need `CompareCount`? @jnyrup, since you submitted this PR, I'm guessing you had a good need for the proposed `CompareCount` that is different than I'm imagining. Could you provide a little more insight into one or two use cases? I'm missing my _ah-ha_ moment. :smile:

jnyrup:
@atifaziz Sure, I should've provided better examples earlier. 
The code that inspired me was something like:
```
var sumOfCounts = sequence1.Count() + sequence2.Count() + sequence3.Count();
return sequence4.Count() >= sumOfCounts;
```

But I didn't care about the _actual_ counts of the sequences, only the _relationship_ between them, i.e. finding the longer sequence.
With the proposed `CompareCount` this could be rewritten as
```
var elements = sequence1.Concat(sequence2).Concat(sequence3);
return sequence4.CompareCount(elements) >= 0;
```

So in this case, with `CompareCount`, I don't need to worry about which of `elements` or `sequence4` is the most expensive to enumerate to end. 
Furthermore this will utilize the `.Count` property if available, again less to worry about.

Right now I can see two ways to find the longer sequence, depending on how expensive `sequence4` is to enumerate:

* If `sequence4` is _cheap_ to enumerate:
```
var count4 = sequence4.Count();

var count1 = sequence1.Count();
if(count1 > count4) return false;

var count2 = sequence2.Count();
if((count1 + count2) > count4) return false;

var count3 = sequence3.Count();
if((count1 + count2 + count3) > count4) return false;

return true;
```

* If `sequence4` is _expensive_ to enumerate:
```
var sumOfCounts = sequence1.Count() + sequence2.Count() + sequence3.Count();
return sequence4.AtLeast(sumOfCounts);
```

leandromoh:
@atifaziz It is usefull for definning the following function for example:

```
IEnuemerable<T>TakeOfGreaterSequence<T>(IEnumerable<T> first, IEnumerable<T> second, int count)
{
    return (first.CompareCount(second) >= 0  ? first : second).Take(count);
}
```

It is not worth to materialize (with `ToList` or `ToArray`) the whole sequence (**_that might be long, expensive or even  infinite_**) if we just want to work with a part of this.

leandromoh:
@atifaziz what about merge this? 

After change quantity iterator method for the current in branco master.

atifaziz:
I've brought this PR up to date with the `master` branch so we can continue the review and eventually merge.

jnyrup:
@atifaziz

As I see it, the tests can't be turned into `TestCase`s, as it requires the parameters to be constant expression.

I'm a bit unsure how to update this PR since you committed directly to the PR.

atifaziz:
> I'm a bit unsure how to update this PR since you committed directly to the PR.

You just pull my changes into your PR branch and continue from there.

> As I see it, the tests can't be turned into `TestCase`s, as it requires the parameters to be constant expression.

According to the [docs](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/attributes):

> Parameters to an attribute constructor are limited to simple types/literals: `bool`, `int`, `double`, `string`, `Type`, enums, etc and arrays of those types. 

So what I had in mind was test cases with arrays of integers or strings. If that doesn't work out, you could still resort to [`[TestCaseSource]`](https://github.com/nunit/docs/wiki/TestCaseData).

jnyrup:
I've updated the PR to have more explicit tests. 

I introduced `TrackingTestCollection<T> : ICollection<T>` 
and `TrackingTestEnumerable<T> : IEnumerable<T>` to ensure which branches are hit in `CompareCount`.
They both use a `TrackingEnumerator` which exposes the number of times `MoveNext` was invoked and whether `Dispose` was invoked.

It doesn't seem to me that `[TestCaseSource]` is suitable for this kind of testing.

leandromoh:
@jnyrup You may use the [UnenumerableList](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq.Test/UnenumerableList.cs) to check if the collection was not iterated.

When need to test the consistency of the operator for lazy sequences as well collections, I like the way we did at [TakeLastTest](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq.Test/TakeLastTest.cs#L74). We can add a `using` statement to check about "disposability" too.

About `[TestCaseSource]`, [MoveTest](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq.Test/MoveTest.cs#L53) has an example if necessary.

jnyrup:
@leandromoh Thanks for the comments.

I have updated the tests to use `[TestCase]` for the four cases:
* comparing collection with collection
* comparing collection with lazy sequence
* comparing lazy sequence with collection
* comparing lazy sequence with lazy sequence

And then three tests to explictly tests the sequnce(s) are properly disposed if the first, second or both `IEnumerables` are lazy sequences.

I have used `BreakingCollection` for the collections instead of `UnenumerableList` as `CompareCount` should take the cheap paths for `ICollection<T>`.

I modified the `TestingSequence` to expose how many times `MoveNext()` was invoked on the underlying enumerator, as I need to verify that lazy sequences are not enumerated more than necessary.

MoreLinq/MoreLinq.csproj
```diff
@@ -54,7 +54,7 @@
   <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'net451' ">
     <DefineConstants>$(DefineConstants);MORELINQ</DefineConstants>
   </PropertyGroup>
-  
+
```
atifaziz: Delete this unrelated change.

MoreLinq.Test/TestingSequence.cs
```diff
@@ -81,6 +85,7 @@ class DisposeTestingSequenceEnumerator : IEnumerator<T>
             readonly IEnumerator<T> _sequence;
 
             public event EventHandler Disposed;
+            public event EventHandler OnMoveNext;
```
atifaziz: This should be renamed to `MovedNext`.

MoreLinq.Test/TestingSequence.cs
```diff
@@ -89,9 +94,14 @@ public DisposeTestingSequenceEnumerator(IEnumerator<T> sequence)
 
             public T Current => _sequence.Current;
             object IEnumerator.Current => Current;
-            public bool MoveNext() => _sequence.MoveNext();
             public void Reset() => _sequence.Reset();
 
+            public bool MoveNext()
+            {
+                OnMoveNext?.Invoke(this, EventArgs.Empty);
```
atifaziz: The event should be raised _if_ `_sequence.MoveNext()` returned `true`. Otherwise `MoveNextCounter` is misleading is really a counter of unconditional invocations of `MoveNext` on `_sequence`. I don't think that's what's intended.

jnyrup: My intention was indeed to count *all* invocations of `MoveNext()`, and not *only* the number of positives when invoking `MoveNext().`
I wanted to verify that `MoveNext()` was invoked *exactly* the minimum number of times necessary.

atifaziz: I see. Perhaps then calling it `MoveNextCallCount` would be more accurate?

jnyrup: I've  renamed the property and backing field on `TestingSequence` from `MoveNextCount` to `MoveNextCallCount`.

MoreLinq.Test/TestingSequence.cs
atifaziz: Rename to `MoveNextCalled`.

jnyrup: Done.

jnyrup:
I'm happy to be contributing and grateful for the many reviews and comments.

Issue #254, opened 2017-02-24
SkipLast missing doc for type parameter T 


atifaziz:
The [documentation for `SkipLast`](https://morelinq.github.io/2.2/ref/api/html/M_MoreLinq_MoreEnumerable_SkipLast__1.htm) is missing ` <typeparam name="T"/>`.

@leandromoh Can you have a look at this?

Issue #214, opened 2016-11-09
Conflict-free extension methods


atifaziz:
**Note:** This is forked from a [comment](https://github.com/morelinq/MoreLINQ/issues/60#issuecomment-133518416) in issue #60.

With extension methods, there's always a potential for conflict with some future version of .NET or another library so following is a proposal to come up with a future-proof and conflict-free solution that puts the user in control.

At the present, when you import the `MoreLinq` namespace, you get all operators whether you like it or not. What would be helpful is to be able to choose exactly the operators one would like to import and avoid those that are not needed or would otherwise conflict with either the .NET Framework or another library referenced in the user's project. This could be done by breaking up `MoreEnumerable` into distinct classes (as opposed to the single partial class it is today), hosting only a single operator along with its overloads. Each `MoreEnumerable` class would then be wrapped in a namespace named after the operator it hosts.

The following sample illustrates this approach:

```c#
#pragma warning disable 1591

using System;
using System.Collections.Generic;

namespace MoreLinq
{
    namespace NoConflict
    {
        namespace DistinctBy
        {
            public static class MoreEnumerable
            {
                public static IEnumerable<TSource> DistinctBy<TSource, TKey>(this IEnumerable<TSource> source,
                    Func<TSource, TKey> keySelector)
                {
                    return MoreLinq.MoreEnumerable.DistinctBy(source, keySelector);
                }

                public static IEnumerable<TSource> DistinctBy<TSource, TKey>(this IEnumerable<TSource> source,
                    Func<TSource, TKey> keySelector, IEqualityComparer<TKey> comparer)
                {
                    return MoreLinq.MoreEnumerable.DistinctBy(source, keySelector, comparer);
                }
            }            
        }

        namespace EquiZip
        {
            public static class MoreEnumerable
            {
                public static IEnumerable<TResult> EquiZip<TFirst, TSecond, TResult>(this IEnumerable<TFirst> first,
                        IEnumerable<TSecond> second,
                        Func<TFirst, TSecond, TResult> resultSelector)
                {
                    return MoreLinq.MoreEnumerable.EquiZip(first, second, resultSelector);
                }
            }
        }
    }
}
```

[...]

PR #211, opened 2016-11-02
Memoize implementation (#100)

leandromoh:
Implemented Memoize extension method (#100).

The Memoize method returns an IBufferedEnumerable<T> and has two overloads: one without parameters and one which receive an bool that indicates if the buffering must happen even if the source implements ICollection<T>. (Otherwise it is assumed that the source already serves as a buffer adequately). (for, in future (after this branch is merge), deal with the TODO's of issue #87)

Currently missing to add method reference in "project.json" and "README.md" files.
I pretend to do it latter to avoid conflicts.

PR #255, opened 2017-02-25
Fix documentation (issue #254)

leandromoh:
Added `<typeparam name="T"/>` for `SkipLast` .

I also added some operator that was missing on `project.json`:

- AtLeast
- AtMost
- CountBetween
- Exactly
- FallbackIfEmpty
- FullGroupJoin


atifaziz:
Thanks for spotting those missing operators and adding them. :+1: For next time, though, would be best to open separate PR for unrelated issues.

atifaziz:
Closed as 0bdb7bb3e078d1f54f10095f2cd345f4d20a10cd and 0bdb7bb3e078d1f54f10095f2cd345f4d20a10cd

leandromoh:
>  For next time, though, would be best to open separate PR for unrelated issues.

@atifaziz thanks for the tip. when do you plan to see the aggregateright and memoize PR?

atifaziz:
@leandromoh I've moved `AggregateRight` to the [next release](https://github.com/morelinq/MoreLINQ/milestone/4) because I figured I won't get time before end of the month to review it. I'm trying to keep to a schedule of a release per month and would rather delay features than releases. As for `Memoize`, last I recall I felt that Ix already gave us that so was hesitant to add it to MoreLinq; unless of course #214 is closed via #235.

leandromoh:
@atifaziz I understood. I also would like to see `Memoize`, `Tranpose` and `UnAggregate` in the next release. Do you think it is possible? As for `Memoize`, we can simple rename it to `Cache` to avoid conflict, as you already suggested.

atifaziz:
> I also would like to see `Memoize`, `Tranpose` and `UnAggregate` in the next release. Do you think it is possible?

We can certainly try provided we've already agreed on adding them. Review takes time so what would help to speed up thing is to up the quality of the PRs using feedback we've been over in past PRs.

> As for `Memoize`, we can simple rename it to `Cache` to avoid conflict, as you already suggested.

This belong with #211.

leandromoh:
> Review takes time so what would help to speed up thing is to up the quality of the PRs using feedback we've been over in past PRs.

Of course, MoreLinq was the first project I contributed and I learnt a lot of things with the reviews:

implementations:

1. bank line between XML documentation and method for clarity
2. exception/documentation descriptions always end with full-stop or period
3. avoid enumerating twice non-list collections
4. xxxImpl method don't must be an extension method
5. quote all params on summary 
6. use var where type is explicit
7. descritive variable names always
8. check comparer.GetHashCode before comparer.Equals
9. using statements go inside the namespace block

tests:

1. non-extension invocation style when testing if source is null
2. use TestCase more appropriately
3. alignment test cases when it has many parameters for clarity
4. no use random data on tests
5. no "xxxSimpleTest"
6. use  sequence equality assertion
7. give a different name for each overload on tests
8. Test methods don't need to end in the Test suffix
9. act-arrange-assert pattern when necessary

atifaziz:
> MoreLinq was the first project I contributed and I learnt a lot of things

You're doing well and thanks for all your contributions so far! :+1: Keep it up! I'm also really glad to see that you've taken note of all the feedback as it'll help to avoid repeating ourselves.

Issue #246, opened 2017-02-05
Unfold/Disaggregate


leandromoh:
`Unfold` function (present and many functional languages like Haskell and F#), is a dual to `fold/aggregate`. 

`Unfold` builds a list from a seed value. It takes a function which either returns `null `if it is done producing the list, or returns the tuple `(x, y)`, `x` which is prepended to the list, and `y` is used as the next element in the recursive call. `x` and `y` may be of different types.

`Unfold` signature is:

`IEnumerable<TResult> UnFold<TSource, TResult>(TSource seed, Func<TSource, Tuple<TResult, TSource>> generator)`

Examples of usage:

```
var naturalNumbers = UnFold(0, n => Tuple.Create(n, n + 1));
var naturalNumbersAsString = UnFold(0, n => Tuple.Create(n.ToString(), n + 1));
var squaresLowerThan100 = UnFold(1, n => (n * n) < 100 ? Tuple.Create(n * n, n + 1) : null);
```

for some cases, the following overload may also be usefull/readable. it bassically split the function into 3 functions:

`IEnumerable<TResult> UnFold<TSource, TResult>(TSource seed, Func<TSource, bool> predicate, Func<TSource, TResult> resultGenerator, Func<TSource, TSource> seedGenerator)`

for example, we can do 

`var fibonacciNumbersLowerThan100 = UnFold(Tuple.Create(0, 1), n => n.Item1 < 100, n => n.Item1, n => Tuple.Create(n.Item2, n.Item1 + n.Item2));`

instead of 

`var fibonacciNumbersLowerThan100 = UnFold(Tuple.Create(0, 1), n => n.Item1 < 100 ? Tuple.Create(n.Item1, Tuple.Create(n.Item2, n.Item1 + n.Item2)) : null);`

I'm planning to submit a PR for this.

What do you think?

fsateler:
It would be great if we could use a type with better member names that `Item1` and `Item2`:

```csharp
class UnfoldMemo<TState, TResult> {
    public TState State { get; }
    public TResult Result { get; }

   public UnfoldMemo(TState state, TResult result) {
       State = state;
       Result = result;
   }
}
static class UnfoldMemo {
    public static UnfoldMemo<TState, TResult>Create<TState, TResult>(TState state, TResult result) => new UnfoldMemo(state, result);
}
```

atifaziz:
Or make it completely generic so you can build on top and use anonymous types, tuples and what not. I would go with the following signature:

[...]

PR #260, opened 2017-03-07
Implemented Unfold (Issue 246) 

leandromoh:
Implemented Unfold method (#246).


MoreLinq/UnFold.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// Returns a sequence generated by applying a state to the generator function, 
+        /// and from its result, determines if the sequence should have a next element and its value,
+        /// and the next state in the recursive call.
+        /// </summary>
+        /// <typeparam name="TState">Type of state elements.</typeparam>
+        /// <typeparam name="T">Type of the elements generated by the generator function.</typeparam>
+        /// <typeparam name="TResult">The type of the elements of the result sequence.</typeparam>
+        /// <param name="state">The initial state.</param>
+        /// <param name="generator">
+        /// Function that takes a state and procudes an object containing all
```
fsateler: s/procudes/produces/

MoreLinq/UnFold.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <typeparam name="TState">Type of state elements.</typeparam>
+        /// <typeparam name="T">Type of the elements generated by the generator function.</typeparam>
+        /// <typeparam name="TResult">The type of the elements of the result sequence.</typeparam>
+        /// <param name="state">The initial state.</param>
+        /// <param name="generator">
+        /// Function that takes a state and procudes an object containing all 
+        /// necessary data for the other functions (predicate, stateSelector, resultSelector).
+        /// </param>
+        /// <param name="predicate">
+        /// Function used to determine if state is valid.
```
fsateler: I would say: `Function to determine if the unfolding should continue`. "Validness" is not something this function really cares about, as it is not defined. For completeness, it should also describe the inputs: `Function to determine if the unfolding should continue based on the current state`.

MoreLinq/UnFold.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </param>
+        /// <param name="stateSelector">
+        /// Function to select the next state.
```
fsateler: I would use generator suffix instead of selector (and change the description accordingly), but that may be just cosmetics. What do you think?
Also, the input should be described as well: `based on the current state`.

MoreLinq/UnFold.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </param>
+        /// <param name="resultSelector">
+        /// Function to select the next result.
```
fsateler: Same here: `Function to select the result from the current state`.

MoreLinq/UnFold.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </param>
+        /// <returns>A sequence containing the results generated by the resultSelector function.</returns>
```
fsateler: Strictly speaking, the results are generated by applying the resultSelector to the state, so this should be noted too.

MoreLinq/UnFold.cs
```diff
@@ -0,0 +1,111 @@
...
+            Func<TState, T> generator,
+            Func<T, bool> predicate,
+            Func<T, TState> stateSelector,
+            Func<T, TResult> resultSelector)
+        {
+            while (true)
+            {
+                var step = generator(state);
+
+                if (!predicate(step))
```
fsateler: The initial state is never returned. I wonder if this is common in other implementations of this function. F# seems to include it:

```f#
Seq.unfold (fun state -> if (state > 0) then None else Some(state, state + 1)) 0
```
That returns a sequence with the initial state.

fsateler:
I also think it would be great to have overloads for the (maybe?) common scenarios where the type arguments are the same (that is, either `resultSelector` or `generator` are identity functions).

MoreLinq/UnFold.cs
atifaziz: > I would use generator suffix instead of selector

@fsateler The computation _based on the current state_ will usually already be done by the `generator` and returned as `T`. However, since `Unfold` doesn't know the domain of `T`, it uses this function to go from `T` to `TState` and so I'd argue that _selector_ is the right suffix here. The concept at work here is really deconstruction, if you will, than generation.

fsateler: Oh, you are correct. Scratch this comment then.

MoreLinq/UnFold.cs
atifaziz: @fsateler It does return the initial state. The F# example translated to this `Unfold` implementation would be:

```c#
Unfold(0, x => new { x, s = x + 1 }, e => e.x == 0, e => e.s, e => e.x);
```

I pasted the relevant bits of an `Unfold` implementation into a C# Interactive session and then evaluated the above:

```
Microsoft (R) Visual C# Interactive Compiler version 1.3.1.60616

>         private static IEnumerable<TResult> UnfoldImpl<TState, T, TResult>(
.             TState state,
.             Func<TState, T> generator,
.             Func<T, bool> predicate,
.             Func<T, TState> stateSelector,
.             Func<T, TResult> resultSelector)
.         {
.             while (true)
.             {
.                 var step = generator(state);
.
.                 if (!predicate(step))
.                     yield break;
.
.                 yield return resultSelector(step);
.                 state = stateSelector(step);
.             }
.         }
>         public static IEnumerable<TResult> Unfold<TState, T, TResult>(
.             TState state,
.             Func<TState, T> generator,
.             Func<T, bool> predicate,
.             Func<T, TState> stateSelector,
.             Func<T, TResult> resultSelector)
.         {
.             if (generator == null) throw new ArgumentNullException(nameof(generator));
.             if (predicate == null) throw new ArgumentNullException(nameof(predicate));
.             if (stateSelector == null) throw new ArgumentNullException(nameof(stateSelector));
.             if (resultSelector == null) throw new ArgumentNullException(nameof(resultSelector));
.
.             return UnfoldImpl(state, generator, predicate, stateSelector, resultSelector);
.         }
> Unfold(0, x => new { x, s = x + 1 }, e => e.x == 0, e => e.s, e => e.x)
UnfoldImpl { 0 }
```

As you can see, the last line prints `UnfoldImpl { 0 }`. Are you seeing something different?

fsateler: Oh I'm terribly sorry. I mistyped my test in LINQPad, and got the wrong conclusion. The fixed test would be:

```f#
// note the >=
Seq.unfold (fun state -> if (state >= 0) then None else Some(state, state + 1)) 0
```

Which yields an empty sequence, just like this implementation. So this means this implementation agrees with F#.

/me makes a note to do reviews *after* coffee...

BTW, we should have a test for this behavior too.

atifaziz: > BTW, we should have a test for this behavior too.

Agreed! I haven't been over the tests or any of the PR files since I'm still waiting for the [build to be fixed](https://github.com/morelinq/MoreLINQ/pull/260#pullrequestreview-25443733) first. 😆

leandromoh:
> Can you fix the following error that is breaking the build?

@atifaziz  I forgot to say I use an older version of MoreLinq (before migration to .NET Core) to create methods and tests because I couldn't install .Net Core. 

Since the error is releated with .Net Core I dont know how to fix it. Could you fix the build please?

atifaziz:
@leandromoh The error isn't related to .NET Core but to .NET 3.5, which doesn't have `System.Tuple<,>`. The baseline for MoreLINQ is still .NET 3.5. If we want to move it to 4.x then that needs to be a separate decision. For now, I would remove the overloads with involving tuples and get it out. We can then add the overloads back later as a future improvement, which I think we'll have to do as a sweeping review across the library.

> I use an old version of MoreLinq (before migration to .NET Core) to create methods and tests

If you're using the older version then that should've also failed to compile since it targets 3.5 that doesn't have any type for tuples.

MoreLinq/UnFold.cs
leandromoh: done

MoreLinq/UnFold.cs
leandromoh: done. thanks for improve de description.
sadly, i am not so good to describe parameters.

MoreLinq/UnFold.cs
leandromoh: done

MoreLinq/UnFold.cs
leandromoh: tests added

leandromoh:
> I also think it would be great to have overloads for the (maybe?) common scenarios where the type arguments are the same (that is, either resultSelector or generator are identity functions).

@fsateler, Can you give me an example of usage?

leandromoh:
> If you're using the older version then that should've also failed to compile since it targets 3.5 that doesn't have any type for tuples.

I set the version for 4.5, because I dont have 3.5 version installed.

MoreLinq/UnFold.cs
```diff
@@ -0,0 +1,87 @@
...
+                    yield break;
+
+                yield return resultSelector(step);
+                state = stateSelector(step);
```
fsateler: While coming up with an example for the overloads where one function is an identity, I realized I am confused a bit by the type distinction between `state` and `step`. Given that `state` is only used to calculate the next `step`, what do we gain by having a `stateSelector` at all? Or in other words, what is the advantage of the current version over:

```csharp
private static IEnumerable<TResult> UnfoldImpl<TState, TResult>(
    TState state,
    Func<TState, TState> generator,
    Func<TState, bool> predicate,
    Func<TState, TResult> resultSelector)
{
    while (true)
    {
        var step = generator(state);

        if (!predicate(step))
            yield break;

        yield return resultSelector(step);
        state = step;
    }
}
```

What am I missing?

atifaziz: How do you get back the result to yield for the current iteration from the state for the next that is `step`? Suppose your seed is 0 and `generator` is a simple increment. Then `step` will be 1 for the _next_ iteration. Your `predicate` and `resultSelector` would have to account for that. For example, `resultSelector` will have to decrement to get the result, so at best, it'll feel clumsy. The `squaresLowerThan100` example from #246 will now read like this: 

```c#
var squaresLowerThan100 =
    Unfold(1, x => x + 1, x => ((x - 1) * (x - 1)) < 100, x => (x - 1) * (x - 1));
```

> where one function is an identity

Which function?

fsateler: > How do you get back the result to yield for the current iteration from the state for the next that is step?

Well, you start with a seed *state*, not seed *value*:

```csharp
var squaresLowerThan100 =
    Unfold(new { x = 1, sq = 0 }, s => new { x = s.x+1, sq = s.x*s.x }, s => s.sq < 100, s => s.sq);
```

I think what is confusing me a lot is that I find that what is currently described as `T step` I would define as `TState state`. After all, it is `step` that carries all the information necessary to generate results, determine if iteration should continue, and generate the next iteration (if we assume `stateSelector` is usually a property access).

>>where one function is an identity
>
> Which function?

I originally wrote "[either resultSelector or generator are identity functions](https://github.com/morelinq/MoreLINQ/pull/260#issuecomment-284725887)". However, having a `generator` being an identity is not likely to be very useful (and certainly not common). This got me thinking I did not understand correctly what the parameters really mean, and I could not resolve my confusion.

BTW, what I really meant is `stateSelector` and `resultSelector` as identities.

leandromoh:
@atifaziz 
I already removed the overloads involving tuples. 
The current PR is passing in tests.

MoreLinq/UnFold.cs
atifaziz: Is your problem really with the word _step_ and what's throwing you off?

> Well, you start with a seed *state*, not seed *value*:

> I think what is confusing me a lot is that I find that what is currently described as `T step` I would define as `TState state`. 

The reason for naming `step` distinctly so is that it contains the next state but it isn't the next state by itself. That comes out later in `state = stateSelector(step)`, and yes, it'll usually be a simple property access.

Do you have a better name to suggest?

fsateler: @atifaziz I'm not sure further discussion is helpful at this point. It's pretty clear that its only me that is confused, and a sample size of 1 is not indicative of a problem 😉 .
If I had written this function, however, I would have probably written it more like this:

```csharp
private static IEnumerable<TResult> UnfoldImpl<TState, T, TResult>(
	T seed,
	Func<T, TState> generator,
	Func<TState, bool> predicate,
	Func<TState, T> valueSelector,
	Func<TState, TResult> resultSelector)
{
	T value = seed;
	while (true)
	{
		var state = generator(value);

		if (!predicate(state))
			yield break;

		yield return resultSelector(state);
		value = valueSelector(state);
	}
}
```

MoreLinq.Test/UnfoldTest.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Linq;
+using NUnit.Framework;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class UnfoldTest
+    {
+        [Test]
+        public void UnfoldGenericWithNullGenerator()
```
atifaziz: Remove `Generic` from this and all other such named tests.

MoreLinq.Test/UnfoldTest.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            Assert.ThrowsArgumentNullException("generator", () =>
+                MoreEnumerable.Unfold<int, Tuple<int, int>, int>(0, null, _ => true, e => e.First, e => e.Second));
+        }
+
+        [Test]
+        public void UnfoldGenericWithNullPredicate()
+        {
+            Assert.ThrowsArgumentNullException("predicate", () =>
+                MoreEnumerable.Unfold(0, e => new Tuple<int, int>(e, e + 1), null, e => e.First, e => e.Second));
```
atifaziz: Use [`Tuple.Create`](https://msdn.microsoft.com/en-us/library/system.tuple.create.aspx) instead of `new Tuple<int, int>` for readability. Ditto elsewhere.

MoreLinq/Unfold.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// Returns a sequence generated by applying a state to the generator function, 
+        /// and from its result, determines if the sequence should have a next element and its value,
+        /// and the next state in the recursive call.
+        /// </summary>
+        /// <typeparam name="TState">Type of state elements.</typeparam>
+        /// <typeparam name="T">Type of the elements generated by the generator function.</typeparam>
+        /// <typeparam name="TResult">The type of the elements of the result sequence.</typeparam>
+        /// <param name="state">The initial state.</param>
+        /// <param name="generator">
+        /// Function that takes a state and produces an object containing all
```
atifaziz: Change to:

> Function that takes a state and computes the next state and the next element of the sequence.

MoreLinq/Unfold.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <param name="state">The initial state.</param>
+        /// <param name="generator">
+        /// Function that takes a state and produces an object containing all 
+        /// necessary data for the other functions (predicate, stateSelector, resultSelector).
+        /// </param>
+        /// <param name="predicate">
+        /// Function to determine if the unfolding should continue based on the current input.
+        /// </param>
+        /// <param name="stateSelector">
+        /// Function to select the state from the current input.
```
atifaziz: Change to:

```c#
        /// <param name="stateSelector">
        /// Function to select the state from the result of the
        /// <paramref name="generator"/> function.
        /// </param>
```

MoreLinq/Unfold.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </param>
+        /// <param name="resultSelector">
+        /// Function to select the result from the current input.
```
atifaziz: Adapt as suggested for `stateSelector`.

MoreLinq/Unfold.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </param>
+        /// <returns>A sequence containing the results generated by the resultSelector function.</returns>
```
atifaziz: Use `<paramref>` to quote `resultSelector`.

MoreLinq.Test/UnfoldTest.cs
leandromoh: As you said, .NET framework's Tuple are after version 4, so it is not avaible here.

MoreLinq/Unfold.cs
leandromoh: done

MoreLinq/Unfold.cs
leandromoh: done

MoreLinq/Unfold.cs
leandromoh: done

MoreLinq/Unfold.cs
leandromoh: done

MoreLinq.Test/UnfoldTest.cs
leandromoh: done

MoreLinq.Test/UnfoldTest.cs
leandromoh: done

leandromoh:
@atifaziz all adjusts were done.

MoreLinq/Unfold.cs
```diff
@@ -32,19 +32,18 @@ static partial class MoreEnumerable
         /// <typeparam name="TResult">The type of the elements of the result sequence.</typeparam>
         /// <param name="state">The initial state.</param>
         /// <param name="generator">
-        /// Function that takes a state and produces an object containing all 
-        /// necessary data for the other functions (predicate, stateSelector, resultSelector).
+        /// Function that takes a state and computes the next state and the next element of the sequence.
         /// </param>
         /// <param name="predicate">
         /// Function to determine if the unfolding should continue based on the current input.
```
atifaziz: Change to:

```c#
        /// Function to determine if the unfolding should continue based the
        /// result of the <paramref name="generator"/> function.
```

leandromoh: done

leandromoh:
@atifaziz the adjust were done. IMHO would be more interesting you request all changes once and I do the adjusts once. The review process would be faster.

atifaziz:
> IMHO would be more interesting you request all changes once and I do the adjusts once. The review process would be faster. What do you think?

@leandromoh Three things:

1. I get bursts of time and I try to cover as much as I can in those when conducting a review. If I were to wait for the perfect moment to provide all the feedback in one session then it could take days. Also, I don't know your agenda so the assumption is that you'll get round to addressing the feedback when you can (time zone, duties, priorities, etc.).

2. I expect you to extrapolate from some of the feedback and address similar/related issues but when those opportunities are missed, it requires another cycle of review.

3. Some new items come to light only when the initial set of feedback has been addressed.

Again, appreciate you working through the reviews.

atifaziz:
@fsateler Are the changes you requested from your review now in?

fsateler:
Oops, sorry, I still have some comments

MoreLinq/Unfold.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Returns a sequence generated by applying a state to the generator function, 
+        /// and from its result, determines if the sequence should have a next element and its value,
+        /// and the next state in the recursive call.
```
fsateler: if the sequence should have a next element, its value, and the next state. (that is, remove the extra "and").

fsateler: Also, I just noticed the "in the recursive call" wording. We are not doing any recursing really. I'm confused about what this means.

MoreLinq/Unfold.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </param>
+        /// <param name="predicate">
+        /// Function to determine if the unfolding should continue based the
+        /// result of the <paramref name="generator"/> function.
+        /// </param>
+        /// <param name="stateSelector">
+        /// Function to select the state from the result of the <paramref name="generator"/> function.
+        /// </param>
+        /// <param name="resultSelector">
+        /// Function to select the result from the result of the <paramref name="generator"/> function.
```
fsateler: This turned out a bit awkward due to the repetition of "result". Unfortunately, I have not come up with a better wording.

"Function to select the result from the output of the generator function"? That feels weird too.

MoreLinq/Unfold.cs
leandromoh: "in the recursive call" here means that it will repeat the whole process again and again, not literally recursion.

MoreLinq/Unfold.cs
leandromoh: " from the output" sound better for me. if no better idea appears, we can go with it.

atifaziz: Unless we rename `resultSelector` to `itemSelector`, “from the output” is good enough for me too.

leandromoh:
@atifaziz @fsateler the xmldoc improviments were done and I ready update `README.md` and `project.json` to get changes from `AggregateRight` PR.  🏁

MoreLinq/project.json
```diff
@@ -1,14 +1,14 @@
...
   "language": "en-US",
   "authors": [
     "MoreLINQ Developers."
   ],
   "copyright": "\u00a9 2008 Jonathan Skeet. Portions \u00a9 2009 Atif Aziz, Chris Ammerman, Konrad Rudolph. Portions \u00a9 2010 Johannes Rudolph, Leopold Bushkin. Portions \u00a9 2015 Felipe Sateler, \u201csholland\u201d. Portions \u00a9 2016 Leandro F. Vieira (leandromoh). Portions \u00a9 Microsoft. All rights reserved.",
   "packOptions": {
-    "releaseNotes": "Adds new operators: Sequence, SkipLast, Unfold. See also https://github.com/morelinq/MoreLINQ/wiki/API-Changes.",
+    "releaseNotes": "Adds new operators: AggregateRight, Sequence, SkipLast, Unfold. See also https://github.com/morelinq/MoreLINQ/wiki/API-Changes.",
```
atifaziz: This change is unrelated to this PR.

atifaziz: Perhaps this came as a result of you rebasing?

leandromoh: Yes, It does

atifaziz: Okay. Just waiting on the go from @fsateler.

fsateler:
LGTM. Thanks for working on this!

And thanks to both of you for being patient with me while I got very confused 😄

atifaziz:
Closed by squashed-merge 0ade8d77399cbb4e41348a53ed362de123bfbea4.

Thanks for your contribution (@leandromoh), review (@fsateler) and prompt actions to make this happen in time for the [2.3.0 release](https://github.com/morelinq/MoreLINQ/milestone/4) coming tomorrow! :champagne:

Issue #65, opened 2015-08-21
Better seeding for Random


atifaziz:
Currently it creates its randomness source using the default constructor of the `Random` class. But the well known problem with that is that this is seeded with the system time, which typically doesn't change for 16ms. So all instances of the random class created during that interval will return the same sequence.

My suggestion is creating one static instance of random and using it(with locking of course) to create the seed for the individual instances of `Random`.

---

Originally reported on Google Code with ID 65

Reported by `CodeInChaos@quantentunnel.de` on 2011-03-02 21:07:16

fsateler:
I disagree with this request. [There is an overload that takes a `Random` parameter](https://morelinq.github.io/2.2/ref/api/html/M_MoreLinq_MoreEnumerable_Random_3.htm), so this should be fairly easy to implement in your own program. Moreover, not all programs need `Random` sequences to be different in quick sucession bursts. This would thus slow down `Random` for these users for no gain.

fsateler:
I'm going to close this. Please reopen if you disagree with me.

atifaziz:
@fsateler I think the issue is a valid concern because while [there's an overload that accepts a `Random`](https://morelinq.github.io/2.2/ref/api/html/M_MoreLinq_MoreEnumerable_Random_3.htm) as you point out, `Random` isn't thread-safe so it's not like you can put it in a static and supply the same instance from multiple threads to get _better seeding_. If we don't want to get bogged down with locking, the best balance here might be to go with a lazy-initialized thread-static so there'll be a `Random` per thread. This is non-trivial to ask everyone to get right and so we should just do it for the simpler overloads that instantiate and use a new `Random` on behalf of the user. A thread-static will also have the least impact on the existing implementation.

Another advantage of sharing a `Random` is that the initialization cost ([space](https://github.com/dotnet/corefx/blob/v1.0.4/src/System.Runtime.Extensions/src/System/Random.cs#L38) and [time](https://github.com/dotnet/corefx/blob/v1.0.4/src/System.Runtime.Extensions/src/System/Random.cs#L57-L86)) and while this may be lesser of a motivation, it's just another bonus.

Since there's only gain, I'd like to re-open and take a crack at this.

> This would thus slow down `Random` for these users for no gain.

You didn't say why this is slow but I'm guessing you were thinking locking?

fsateler:
[...]

PR #265, opened 2017-03-23
Better seeding for Random (addresses #65)

atifaziz:
This PR addresses #65.

### DIFF
--- a/MoreLinq/Random.cs
+++ b/MoreLinq/Random.cs
@@ -23,20 +23,15 @@ namespace MoreLinq
 
     public static partial class MoreEnumerable
     {
-        static int _seed = unchecked((int) DateTime.Now.Ticks);
-        [ThreadStatic] static Random _sharedRandom;
-
-        static Random ThreadRandom => _sharedRandom ?? (_sharedRandom = new Random(Interlocked.Increment(ref _seed)));
-
         /// <summary>
         /// Returns an infinite sequence of random integers using the standard 
         /// .NET random number generator.
         /// </summary>
         /// <returns>An infinite sequence of random integers</returns>
-        
+
         public static IEnumerable<int> Random()
         {
-            return Random(ThreadRandom);
+            return Random(GlobalRandom.Instance);
         }
 
         /// <summary>
@@ -64,7 +59,7 @@ namespace MoreLinq
         {
             if (maxValue < 0) throw new ArgumentOutOfRangeException(nameof(maxValue));
 
-            return Random(ThreadRandom, maxValue);
+            return Random(GlobalRandom.Instance, maxValue);
         }
 
         /// <summary>
@@ -94,7 +89,7 @@ namespace MoreLinq
         
         public static IEnumerable<int> Random(int minValue, int maxValue)
         {
-            return Random(ThreadRandom, minValue, maxValue);
+            return Random(GlobalRandom.Instance, minValue, maxValue);
         }
 
         /// <summary>
@@ -124,7 +119,7 @@ namespace MoreLinq
         
         public static IEnumerable<double> RandomDouble()
         {
-            return RandomDouble(ThreadRandom);
+            return RandomDouble(GlobalRandom.Instance);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


fsateler:
CI failure appears to be a network outage and thus unrelated.

atifaziz:
> this shared random is then passed directly, which would be unsafe if an async context results in a thread jump

I don't see the issue. If there's a thread jump then the `Random` instance associated with that thread will be used.

fsateler:
>> this shared random is then passed directly, which would be unsafe if an async context results in a thread jump
>
> I don't see the issue. If there's a thread jump then the Random instance associated with that thread will be used.

If `Random` is invoked in thread T1 but the result enumerated in thread T2, then the `ThreadRandom` instance of T1 is used in T2, because `RandomImpl` will capture the passed `Random` instance from T1. 

This change will make it required to enumerate from the same thread that the enumerable was created. Previously, it was only required that the enumerable was only enumerated by a single thread at a time.

atifaziz:
> If `Random` is invoked in thread T1 but the result enumerated in thread T2, then the `ThreadRandom` instance of T1 is used in T2, because `RandomImpl` will capture the passed `Random` instance from T1. 

Oh man, you're totally spot on and good catch! I'm so glad another pair of eyes looked over. Silly me that the very problem I set out to solve ended up rearing its ugly head via the closure. 👹

atifaziz:
@fsateler In 172629953d44b15ddc78f2fdd143ee3893dda048, I've taken another approach. Fortunately, not only is `Random` open to sub-classing but all its `NextXXX` methods are also virtual. I used that to create a singleton implementation that internally uses thread-local `Random` instances. So while the global is still captured, each call into the various `NextXXX` methods uses the `Random` of the thread _at the time of each iteration_.

Let me know what new blunders I've created. 😄

fsateler:
> Fortunately, not only is Random open to sub-classing but all its NextXXX methods are also virtual. I used that to create a singleton implementation that internally uses thread-local Random instances. 

This is avoids this problem. However, by potentially using different `Random` objects while enumerating, the distribution of random numbers might not have the properties expected of a `Random` (uniform distribution is probably the most important). I have no idea how much the distribution could change.

Is this a problem worth worrying about? The conditions under which this problem would arise are not common (constant thread-jumping while enumerating a single sequence).

atifaziz:
> Is this a problem worth worrying about? The conditions under which this problem would arise are not common (constant thread-jumping while enumerating a single sequence).

@fsateler You raise a fair concern but I feel @stephentoub put it best:

> If I really cared about the quality of the numbers, I would use `RNGCryptoServieProvider`.  I’m just trying to escape the hazards of getting back all 0s, or having all of the threads seeing the exact same numbers.  This is a fair compromise if the statistical properties beyond that aren’t that important.

fsateler:
@atifaziz Right, I agree for most users this will not be an issue. Maybe it is only necessary to add a remark to the documentation for the relevant overloads?

atifaziz:
@fsateler What do you think of this?

```c#
        /// <remarks>
        /// The implementation internally uses a thread-local instance of
        /// <see cref="System.Random" /> to generate a random number on each
        /// iteration. The actual <see cref="System.Random" /> instance used
        /// therefore will depend on the thread on which a single iteration is
        /// taking place; that is the call to
        /// <see cref="System.Collections.IEnumerator.MoveNext()" />. If the
        /// overall iteration takes place on different threads (e.g.
        /// asynchronous await completing on different threads) then various
        /// different <see cref="System.Random" /> instances will be involved
        /// in the generation of the sequence of random numbers.
        /// </remarks>
```

fsateler:
@atifaziz Looks good to me, modulo a minor nit. I'd also note that the thread-local instance is shared among enumerables in the same thread:

```csharp
        /// <remarks>
        /// The implementation internally uses a shared, thread-local instance of
        /// <see cref="System.Random" /> to generate a random number on each
        /// iteration. The actual <see cref="System.Random" /> instance used
        /// therefore will depend on the thread on which a single iteration is
        /// taking place; that is the call to
        /// <see cref="System.Collections.IEnumerator.MoveNext()" />. If the
        /// overall iteration takes place on different threads (e.g.
        /// via asynchronous awaits completing on different threads) then various
        /// different <see cref="System.Random" /> instances will be involved
        /// in the generation of the sequence of random numbers. Because the
        /// <see cref="System.Random" /> instance is shared, if multiple sequences
        /// are generated in the same thread, the order of enumeration affects the
        /// resulting sequences.
        /// </remarks>
```

Issue #274, opened 2017-04-13
ToLookup overloads for key-value pairs & tuples of 2


atifaziz:
Following on from #272, it makes sense to have overloads of [`ToLookup`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tolookup--3) for key-value pairs and tuples of 2.

atifaziz:
Available for testing in [2.4.0 beta 2](https://www.nuget.org/packages/morelinq/2.4.0-beta-02).

[Documentation also available online](https://morelinq.github.io/2.4/ref/api/html/Overload_MoreLinq_MoreEnumerable_ToLookup.htm).

PR #275, opened 2017-04-13
ToLookup overloads for key-value pairs & tuples of 2 (#274)

atifaziz:
PR for #274

- [x] Implementation
- [x] Unit tests
- [x] Update operators list
- [x] Update release notes
- [x] Add quick description to README

### DIFF
--- a/MoreLinq.Test/ToLookupTest.cs
+++ b/MoreLinq.Test/ToLookupTest.cs
@@ -16,7 +16,6 @@
 #endregion
 
 using System;
-using System.Collections.Generic;
 using NUnit.Framework;
 
 namespace MoreLinq.Test
@@ -24,13 +23,6 @@ namespace MoreLinq.Test
     [TestFixture]
     public class ToLookupTest
     {
-        [Test]
-        public void ToLookupWithNullKeyValuePairSequence()
-        {
-            Assert.ThrowsArgumentNullException("source", () =>
-                MoreEnumerable.ToLookup((IEnumerable<KeyValuePair<object, object>>) null));
-        }
-
         [Test]
         public void ToLookupWithKeyValuePairs()
         {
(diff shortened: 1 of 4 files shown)


MoreLinq.Test/ToLookupTest.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class ToLookupTest
+    {
+        [Test]
+        public void ToLookupWithNullKeyValuePairSequence()
+        {
+            Assert.ThrowsArgumentNullException("source", () =>
+                MoreEnumerable.ToLookup((IEnumerable<KeyValuePair<object, object>>) null));
```
fsateler: This tests an exception that is actually thrown by the real `ToLookup`, as the implementation doesn't check its arguments. I now see the same is true about the `ToDictionary` overload. Conceptually, I find it wrong to test behavior from the underlying library. What if in some version of the runtime the first argument to `ToLookup` is not named "source"?

atifaziz: > Conceptually, I find it wrong to test behavior from the underlying library

You could also look at it as testing an assumption. As long as the test never went red initially, there was nothing to fix!

> What if in some version of the runtime the first argument to `ToLookup` is not named "source"?

Then the tests would break and we'll know!

And while you raise a fair point, I've realized that these tests are now redundant with [`NullArgumentTest`](https://github.com/morelinq/MoreLINQ/blob/b25eef043c6a54adead8428dcf2db5cf23f2dd6d/MoreLinq.Test/NullArgumentTest.cs). Perhaps all such tests should be removed? :thinking:

atifaziz:
@fsateler We good to merge this?

fsateler:
LGTM

Issue #286, opened 2017-05-06
ScanRight function


leandromoh:
Now we have AggregateRight. We should provide  ScanRight function to provide symmetry.

I would like to submit a PR for this.

What do you think?

atifaziz:
Good idea so go for the PR!

PR #287, opened 2017-05-09
implemented scanRight (issue #286)

leandromoh:
This close issue #286

### DIFF
--- a/MoreLinq.Test/ScanRightTest.cs
+++ b/MoreLinq.Test/ScanRightTest.cs
@@ -1,6 +1,6 @@
 ﻿#region License and Terms
 // MoreLINQ - Extensions to LINQ to Objects
-// Copyright (c) 2016 Leandro F. Vieira (leandromoh). All rights reserved.
+// Copyright (c) 2017 Leandro F. Vieira (leandromoh). All rights reserved.
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/MoreLinq.Test/ScanRightTest.cs
+++ b/MoreLinq.Test/ScanRightTest.cs
@@ -71,6 +71,12 @@ namespace MoreLinq.Test
             Assert.That(result, Is.EqualTo(expectations));
         }
 
+        [Test]
+        public void ScanRightIsLazy()
+        {
+            new BreakingSequence<int>().ScanRight(BreakingFunc.Of<int, int, int>());
+        }
+
         // Overload 2 Test
 
         [Test]
@@ -114,6 +120,12 @@ namespace MoreLinq.Test
             var expectations = new[] { "(1+(2+(3+(4+5))))", "(2+(3+(4+5)))", "(3+(4+5))", "(4+5)", "5" };
 
             Assert.That(result, Is.EqualTo(expectations));
+        }
+
+        [Test]
+        public void ScanRightSeedIsLazy()
+        {
+            new BreakingSequence<int>().ScanRight(string.Empty, BreakingFunc.Of<int, string, string>());
         }
     }
 }
--- a/MoreLinq.Test/ScanRightTest.cs
+++ b/MoreLinq.Test/ScanRightTest.cs
@@ -1,0 +1,119 @@
+﻿#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2016 Leandro F. Vieira (leandromoh). All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class ScanRightTest
+    {
+        // Overload 1 Test
+
+        [Test]
+        public void ScanRightWithNullSequence()
+        {
+            Assert.ThrowsArgumentNullException("source",
+                () => MoreEnumerable.ScanRight<int>(null, (a, b) => a + b));
+        }
+
+        [Test]
+        public void ScanRightWithNullFunc()
+        {
+            Assert.ThrowsArgumentNullException("func",
+                () => Enumerable.Range(1, 5).ScanRight(null));
+        }
+
+        [Test]
+        public void ScanRightWithEmptySequence()
+        {
+            var result = new int[0].ScanRight((a, b) => a + b);
+
+            Assert.That(result, Is.EqualTo(new int[0]));
+        }
+
+        [Test]
+        public void ScanRightFuncIsNotInvokedOnSingleElementSequence()
+        {
+            const int value = 1;
+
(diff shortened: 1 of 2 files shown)


MoreLinq.Test/ScanRightTest.cs
```diff
@@ -0,0 +1,119 @@
+﻿#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2016 Leandro F. Vieira (leandromoh). All rights reserved.
```
fsateler: It's 2017 😉

MoreLinq/ScanRight.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <remarks>
+        /// This operator uses deferred execution and streams its results.
+        /// </remarks>
+
+        public static IEnumerable<TSource> ScanRight<TSource>(this IEnumerable<TSource> source, Func<TSource, TSource, TSource> func)
+        {
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (func == null) throw new ArgumentNullException(nameof(func));
+
+            var list = (source as IList<TSource>) ?? source.ToList();
```
fsateler: This will enumerate the source immediately.

MoreLinq.Test/ScanRightTest.cs
```diff
@@ -0,0 +1,119 @@
...
+
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class ScanRightTest
```
fsateler: There is no test that all overloads are lazy.

MoreLinq/ScanRight.cs
fsateler: Not only that, but if the resulting enumerable is enumerated multiple times, the source sequence will be enumerated only once.

MoreLinq/ScanRight.cs
```diff
@@ -0,0 +1,105 @@
...
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Peforms a right-associative scan (inclusive prefix) on a sequence of elements.
+        /// This operator is the right-associative version of the 
+        /// <see cref="MoreEnumerable.Scan{TSource}(IEnumerable{TSource}, Func{TSource, TSource, TSource})"/> LINQ operator.
+        /// </summary>
+        /// <typeparam name="TSource">Type of elements in source sequence.</typeparam>
+        /// <param name="source">Source sequence.</param>
+        /// <param name="func">A right-associative accumulator function to be invoked on each element.</param>
```
fsateler: I think it would be useful to describe here the order of the arguments (first argument is the current value in the sequence, second argument is the previous accumulator value), because the types are not giving us any hints here.

MoreLinq/ScanRight.cs
```diff
@@ -0,0 +1,105 @@
...
+            if (list.Count == 0)
+                return Enumerable.Empty<TSource>();
+
+            return ScanRightImpl(list, list.Last(), func, list.Count - 1);
+        }
+
+        /// <summary>
+        /// Peforms a right-associative scan (inclusive prefix) on a sequence of elements.
+        /// The specified seed value is used as the initial accumulator value.
+        /// <see cref="MoreEnumerable.Scan{TSource, TState}(IEnumerable{TSource}, TState, Func{TState, TSource, TState})"/> LINQ operator.
```
fsateler: I think you are missing the "This operator is the right-associative version of the" part of the sentence here.

MoreLinq/ScanRight.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <typeparam name="TSource">The type of the elements of source.</typeparam>
+        /// <typeparam name="TAccumulate">The type of the accumulator value.</typeparam>
+        /// <param name="source">Source sequence.</param>
+        /// <param name="seed">The initial accumulator value.</param>
+        /// <param name="func">A right-associative accumulator function to be invoked on each element.</param>
+        /// <returns>The scanned sequence.</returns>
+        /// <example>
+        /// <code>
+        /// var numbers = Enumerable.Range(1, 4);
+        /// string result = numbers.AggregateRight("5", (a, b) => string.Format("({0}/{1})", a, b));
```
fsateler: ScanRight, not AggregateRight.

MoreLinq/ScanRight.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// This operator is the right-associative version of the 
+        /// <see cref="MoreEnumerable.Scan{TSource}(IEnumerable{TSource}, Func{TSource, TSource, TSource})"/> LINQ operator.
+        /// </summary>
+        /// <typeparam name="TSource">Type of elements in source sequence.</typeparam>
+        /// <param name="source">Source sequence.</param>
+        /// <param name="func">A right-associative accumulator function to be invoked on each element.</param>
+        /// <returns>The scanned sequence.</returns>
+        /// <example>
+        /// <code>
+        /// string result = Enumerable.Range(1, 5).Select(i => i.ToString()).ScanRight((a, b) => string.Format("({0}/{1})", a, b));
```
fsateler: Result is an `IEnumerable<string>`

atifaziz: That's why `var` is almost always better.

leandromoh:
@fsateler thanks for review! all items were fixed.

MoreLinq/ScanRight.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <remarks>
+        /// This operator uses deferred execution and streams its results.
+        /// </remarks>
+
+        public static IEnumerable<TAccumulate> ScanRight<TSource, TAccumulate>(this IEnumerable<TSource> source, TAccumulate seed, Func<TSource, TAccumulate, TAccumulate> func)
+        {
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (func == null) throw new ArgumentNullException(nameof(func));
+
+            return _(); IEnumerable<TAccumulate> _()
```
atifaziz: I sense a lot of duplication here that can be consolidate with [the same in the first overload](https://github.com/leandromoh/MoreLINQ/blob/e9d321f372ae4292ea143a84263bab3aec355ed5/MoreLinq/ScanRight.cs#L53-L62).

leandromoh: I agree. I already have tried to do this lógic in the function that returns the stack, but its code become more verbose and less clear, só i let as It is. Can you show me How you would do?

atifaziz: > Can you show me How you would do?

See 2656082881c2897e504c8a42b08a13b27f1dc8e3 although [the `struct` is ugly] and does add bloat. We can revert if both you and @fsateler find it worse. I do find it however that it puts the whole algorithm in one place neatly.

leandromoh: IMO it became worse. Why we can't use tuples here?

leandromoh:
@atifaziz  all adjusts done, thanks!

MoreLinq/ScanRight.cs
atifaziz: > Now it is more verbose (more lines) and a bit harder to understand.

I don't think there's more lines. There's a new data structure.

> Why we can't use tuples here?

Because as [I commented in the code, we can once we drop .NET Framework 3.5 as a target](https://github.com/morelinq/MoreLINQ/commit/2656082881c2897e504c8a42b08a13b27f1dc8e3#diff-d45d5bd92cdeef152abc5d8a68122510R89).

leandromoh: I read, but what the problem with use tuples together with . NET 3.5? Would se need a more recent version?

atifaziz: [System.ValueTuple isn't available for .NET 3.5](https://www.nuget.org/packages/System.ValueTuple/4.3.1).

atifaziz:
@fsateler Are we good to merge this?

Issue #291, opened 2017-05-11
Create an enumerable given an enumerator factory


atifaziz:
`Create` will return an `IEnumerable<>` given a function that creates an `IEnumerator<>`. Since this is essentially the same signature as for [`IEnumerable<>`'s only method `GetEnumerator`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1.getenumerator), it enables getting a boilerplate streaming `IEnumerable<>` implementation whenever you have function returning an `IEnumerator<>`, but really needed an `IEnumerable<>`. For example, suppose you have an extension method like this:

```c#
static class Extensions
{
    public static IEnumerator<string> ReadLines(this TextReader reader)
    {
        using (reader)
        for (var line = reader.ReadLine(); line != null; line = reader.ReadLine())
            yield return line;
    }
}
```

`Create` would enable using the above function in a query:

```c#
var q =
    from fp in Directory.EnumerateFiles(Environment.GetEnvironmentVariable("windir"), "*.ini")
    select new
    {
        Path  = fp,
        Content =
            from line in MoreEnumerable.Create(() => File.OpenText(fp).ReadLines())
            select line.TrimStart() into line
            where line.Length > 0 && line[0] != ';' // skip blanks + comments
            select line
    };
```

It would also enable writing in-line, ad-hoc and anonymous iterators using [local functions (introduced with C# 7)](https://docs.microsoft.com/en-us/dotnet/articles/csharp/whats-new/csharp-7#local-functions), as shown here to conduct multiple aggregations in a single pass without materialization:

```c#
var summary =
    from x in Enumerable.Range(1, 20)
    group x by x % 5 == 0 into g
    select new
    {
        MultiplesOf5 = g.Key,
        Run =
            from x in MoreEnumerable.Create(() =>
            {
                return _(); IEnumerator<(int Num, int Min, int Max, int Count, int Sum)> _()
                {
                    var min = int.MaxValue;
                    var max = int.MinValue;
                    var sum = 0;
                    var count = 0;
                    using (var e = g.GetEnumerator())
                    while (e.MoveNext())
                    {
                        var x = e.Current;
                        if (x < min) min = e.Current;
                        if (x > max) max = e.Current;
                        sum += e.Current;
                        yield return (x, min, max, ++count, sum);
                    }
                }
            })
            select new
            {
                x.Num, x.Min, x.Max, x.Count, x.Sum,
                Average = (double) x.Sum / x.Count
[...]

PR #292, opened 2017-05-11
Create implementation (#291)

atifaziz:
This PR is the implementation of `Create` as proposed in #291.

### DIFF
--- a/MoreLinq.Test/CreateTest.cs
+++ b/MoreLinq.Test/CreateTest.cs
@@ -1,3 +1,20 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2017 Atif Aziz. All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
 using System.Linq;
 using System.Collections.Generic;
 using NUnit.Framework;
--- a/MoreLinq.Test/CreateTest.cs
+++ b/MoreLinq.Test/CreateTest.cs
@@ -49,5 +49,14 @@ namespace MoreLinq.Test
                 r.ReadEnd();
             }
         }
+
+        [Test]
+        public void CreateDisposesEnumeratorWhenOwnEnumeratorDisposed()
+        {
+            var xs = new[] { 123, 456 };
+            using (var ts = xs.AsTestingSequence())
+            using (var e = MoreEnumerable.Create(() => ts.GetEnumerator()).Read())
+                Assert.That(e.Read(), Is.EqualTo(xs[0]));
+        }
     }
 }
--- a/MoreLinq.Test/CreateTest.cs
+++ b/MoreLinq.Test/CreateTest.cs
@@ -1,0 +1,36 @@
+using System.Linq;
+using System.Collections.Generic;
+using NUnit.Framework;
+
+namespace MoreLinq.Test
+{
+    [TestFixture]
+    public class CreateTest
+    {
+        [Test]
+        public void CreateIsLazy()
+        {
+            Assert.That(MoreEnumerable.Create(BreakingFunc.Of<IEnumerator<object>>()), Is.Not.Null);
+        }
+
+        [Test]
+        public void CreateYieldsEnumeratorElements()
+        {
+            var xs = Enumerable.Range(1, 10);
+            var result = MoreEnumerable.Create(() => xs.GetEnumerator());
+            Assert.That(result, Is.EquivalentTo(xs));
+        }
+
+        [Test]
+        public void CreateDisposesEnumerator()
+        {
+            var item = new object();
+            using (var items = Enumerable.Repeat(item, 1).AsTestingSequence())
+            {
+                var r = MoreEnumerable.Create(() => items.GetEnumerator()).Read();
+                Assert.That(r.Read(), Is.EqualTo(item));
+                r.ReadEnd();
+            }
+        }
+    }
+}
--- a/MoreLinq.Test/CreateTest.cs
+++ b/MoreLinq.Test/CreateTest.cs
@@ -43,8 +43,8 @@ namespace MoreLinq.Test
         {
             var item = new object();
             using (var items = Enumerable.Repeat(item, 1).AsTestingSequence())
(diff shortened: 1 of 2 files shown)


MoreLinq.Test/CreateTest.cs
```diff
@@ -0,0 +1,36 @@
+using System.Linq;
```
fsateler: Copyright header is missing.

MoreLinq.Test/CreateTest.cs
```diff
@@ -0,0 +1,36 @@
...
+        [Test]
+        public void CreateYieldsEnumeratorElements()
+        {
+            var xs = Enumerable.Range(1, 10);
+            var result = MoreEnumerable.Create(() => xs.GetEnumerator());
+            Assert.That(result, Is.EquivalentTo(xs));
+        }
+
+        [Test]
+        public void CreateDisposesEnumerator()
```
fsateler: There should be a test that it is disposed even if the sequence is not read entirely (as long as the wrapper is disposed).

MoreLinq/Create.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        /// <typeparam name="T">Type of returned sequence elements.</typeparam>
+        /// <param name="factory">
+        /// Function that creates an <see cref="IEnumerator{T}"/>.</param>
+        /// <remarks>
+        /// This method uses deferred execution semantics and streams its
+        /// results.
+        /// </remarks>
+
+        public static IEnumerable<T> Create<T>(Func<IEnumerator<T>> factory)
```
fsateler: I'm not sold on the `Create` name. It is too generic. After all, most methods in MoreLinq create an `IEnumerable<T>`. If we add an operator for wrapping a `Func<IEnumerable<T>>` [as I suggested](https://github.com/morelinq/MoreLINQ/issues/291#issuecomment-300809598) while trying to understand the need for this method, would it also be called `Create`? I don't think it would be a good name. An alternative would be to call it Wrap, since it can be thought of as a thin wrapper over the `IEnumerator<T>`, or Lazy, since it is also hiding a lazily-created sequence.

MoreLinq/Create.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            if (factory == null) throw new ArgumentNullException(nameof(factory));
+            return _(); IEnumerable<T> _()
```
fsateler: I think it would be cheaper (performance-wise) to have a dedicated class, as we know it can be simpler than the one generated by the compiler:

```csharp
private class EnumeratorWrapper<T> : IEnumerable<T>
{
    private readonly Func<IEnumerator<T>> Underlying;
    public EnumeratorWrapper(Func<IEnumerator<T>> underlying)
    {
        Underlying = underlying;
    }

    public IEnumerator<T> GetEnumerator() => Underlying();

    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
}
```

MoreLinq/Create.cs
atifaziz: `Wrap` makes sense when you're not changing the type.

> After all, most methods in MoreLinq create an `IEnumerable<T>`

Yeah and most of them are named after the operation they perform and so I would have wanted to call it `Enumerable` (because it turns an `IEnumerator<>` returning function into an `IEnumerable<>`) but `MoreEnumerable.Enumerable` seemed ugly. I also thought about `Delegate` because it returns an `IEnumerable<T>` that merely delegates its implementation to a function. However, it's really the most basic _creation function_ you can possibly imagine/implement and so I'd say `Create` sounds about right. Oh and look, [System.Interactive has one going by the same name](https://github.com/Reactive-Extensions/Rx.NET/blob/f47d85ef7cb70f6ba9eecc5f82462a13f3b8e821/Ix.NET/Source/System.Interactive/Create.cs). 😄

fsateler: >  Oh and look, System.Interactive has one going by the same name. 😄

Hmm, and the implementation is the same. Does it make sense to have this in morelinq too?

atifaziz: > Does it make sense to have this in morelinq too?

I agree, it's arguable and I wouldn't be completely upset if we drop it. I had plans to build a couple of operators on top of it but have scrapped those. The only advantage of having it in MoreLINQ, however, is that System.Interactive targets .NET 4.5 as a minimum, which is a lot to ask for something as trivial as `Create`.

fsateler: > I agree, it's arguable and I wouldn't be completely upset if we drop it. I had plans to build a couple of operators on top of it but have scrapped those.

OOC, what other operators? If we have something that builds on top of `Create`, it would be weird if we didn't provide it.

> The only advantage of having it in MoreLINQ, however, is that System.Interactive targets .NET 4.5 as a minimum, which is a lot to ask for something as trivial as Create.

Well, it's not like 4.5 is such a hard requirement to make. It has been around for 5 years already and can be installed [on any currently supported OS](https://www.microsoft.com/en-us/download/details.aspx?id=42642). 3.5 SP1 is the [only older version that is still supported](https://support.microsoft.com/en-us/lifecycle/search?alpha=.net%20framework), and AFAICT, it is only offered for [old OS](https://www.microsoft.com/en-us/download/details.aspx?id=22), which are already only [on extended support](https://support.microsoft.com/en-us/lifecycle/search?sort=PN&alpha=Windows%20Server%202008&Filter=FilterNO).

I wonder if we should just bite the bullet and do the same for MoreLinq. It would save [having to create single-purpose, internal structs](https://github.com/morelinq/MoreLINQ/pull/287/files#diff-d45d5bd92cdeef152abc5d8a68122510R89), and would eliminate a couple of `#if`s.

atifaziz: > OOC, what other operators?

High-level operators that summarize when you are exploring a data set, somewhat akin to [`summary` in R](https://stat.ethz.ch/R-manual/R-devel/library/base/html/summary.html):

```
> summary(cars)
     speed           dist       
 Min.   : 4.0   Min.   :  2.00  
 1st Qu.:12.0   1st Qu.: 26.00  
 Median :15.0   Median : 36.00  
 Mean   :15.4   Mean   : 42.98  
 3rd Qu.:19.0   3rd Qu.: 56.00  
 Max.   :25.0   Max.   :120.00 
```

Also posted an example to #291 but they're really half-baked ideas so not ready to roll into MoreLINQ and possibly even out of scope (like we decided for linear regression from #298).

> If we have something that builds on top of `Create`, it would be weird if we didn't provide it.

True though most won't care how one operator uses another internally. It doesn't discount `Create` to stand on its own. I was eager to add this for further exploration and while I'm perfectly happy with using the one from System.Interactive, it causes conflicts with MoreLinq in some cases (pending #214).

> I wonder if we should just bite the bullet and do the same for MoreLinq. It would save [having to create single-purpose, internal structs](https://github.com/morelinq/MoreLINQ/pull/287/files#diff-d45d5bd92cdeef152abc5d8a68122510R89), and would eliminate a couple of `#if`s.

Let's take that up in a new issue.

atifaziz:
Closing because #291 has been closed.

Issue #293, opened 2017-05-12
FallbackIfEmpty<T>(this IEnumerable<T> source, Func<T> fallback)?


abatishchev:
Hi,
Currently FallBackIfEmpty has the following signature:
```csharp
    public static IEnumerable<T> FallbackIfEmpty<T>(
	this IEnumerable<T> source,
	T fallback)
```
Does it mean if `T` is a method call `private static T CreateDefault()` it would be called anyway, even if `source` is not empty? I think yes, it would be.

So I propose an overload:
```csharp
    public static IEnumerable<T> FallbackIfEmpty<T>(
	this IEnumerable<T> source,
	Func<T> fallbackFactory)
```
which would call it if only `source` is indeed empty.

Are you willing to accept a pr from me once agreed on the principles. Thanks!

fsateler:
Indeed. In my local projects I have such an overload. I should probably clean it up a bit and post the PR...

abatishchev:
Cool! Please add. Eager to assist/review if necessary.

atifaziz:
There's a few and possibly more reusable and interesting ways to solve this.

First of all, the desired effect can be achieved today using [the `FallbackIfEmpty` overload that takes a sequence of fallback values](http://morelinq.github.io/2.0/ref/api/html/M_MoreLinq_MoreEnumerable_FallbackIfEmpty__1.htm). If you're on the latest and greatest toys like C# 7 then you can use [local functions](https://docs.microsoft.com/en-us/dotnet/articles/csharp/whats-new/csharp-7#local-functions) together with [`Create`](https://github.com/Reactive-Extensions/Rx.NET/blob/f47d85ef7cb70f6ba9eecc5f82462a13f3b8e821/Ix.NET/Source/System.Interactive/Create.cs) from [System.Interactive](https://www.nuget.org/packages/System.Interactive) to supply the fallback values lazily via an iterator, like this:

```c#
var q =
    Enumerable.Empty<int>()
              .FallbackIfEmpty(EnumerableEx.Create(() => { return _(); IEnumerator<int> _()
              {
                  yield return 1;
                  yield return 2;
                  yield return 3;
              }}));
```

The only downside is that it may be cumbersome to have all the syntactic ceremony if you're just returning one or two fallback values. But before you dismiss it on that basis, keep in mind that the above example is a toy one. In a more complex & realistic example such as the next one, the ceremony takes the back seat:

```c#
[...]

PR #294, opened 2017-05-12
Some FallbackIfEmpty improvements

fsateler:
This includes a fix for #293, and an optimization to preserve the source and fallback collections if possible.

/cc @abatishchev


fsateler:
So, #333 was merged. This makes e39fc66 less interesting, but b2c7bec still looks like something that should be merged. Once a decision is made in #293, this can be rebased or closed. In the meantime, I have pushed the commit preserving arguments if possible to a separate PR (#340), as I think it is a less controversial improvement.

PR #303, opened 2017-05-27
Fix documentation for ScanRight and ToArrayByIndex

leandromoh:
put `ScanRight `and `ToArrayByIndex `on README.md and MoreLinq.csproj


atifaziz:
Thanks!

PR #304, opened 2017-05-27
Collection Optmization for TakeLast

leandromoh:


leandromoh:
can I do it in the same way that [SliceTest](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq.Test/SliceTest.cs) does ? That is, testing the collection in the same method that the enumerable.

atifaziz:
@leandromoh If I've understood you correctly then yes, sure.

leandromoh:
@atifaziz  done!

MoreLinq/TakeLast.cs
```diff
@@ -50,25 +51,28 @@ public static IEnumerable<TSource> TakeLast<TSource>(this IEnumerable<TSource> s
...
 
-            return _(); IEnumerable<TSource> _()
-            {
-                Debug.Assert(source != null);
+            return 
+                source is ICollection<TSource> col
+                ? col.Skip(col.Count - count)
+                : _(); IEnumerable<TSource> _()
+                {
+                    Debug.Assert(source != null);
```
fsateler: This is redundant now that the function is inline.

atifaziz: Removed in d7a4d39cb6433c84a0769eae6e10049b91a3bc93

leandromoh:
done!

atifaziz:
> I see now that the CLR does not optimize for `ICollection<>`, …

@fsateler It can't really because an `ICollection<>` can't be indexed into.

> …only for `IList<>`
> https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Skip.cs#L45

Except [`Skip` on the Desktop CLR (at least up to 4.6.x) does not optimise for `IList<>`](http://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,519):

```c#
        //   Copyright (c) Microsoft Corporation.  All rights reserved.

        public static IEnumerable<TSource> Skip<TSource>(this IEnumerable<TSource> source, int count) {
            if (source == null) throw Error.ArgumentNull("source");
            return SkipIterator<TSource>(source, count);
        }
 
        static IEnumerable<TSource> SkipIterator<TSource>(IEnumerable<TSource> source, int count) {
            using (IEnumerator<TSource> e = source.GetEnumerator()) {
                while (count > 0 && e.MoveNext()) count--;
                if (count <= 0) {
                    while (e.MoveNext()) yield return e.Current;
                }
            }
        }
```

leandromoh:
> I find the tests confusing now. It's not very clear why the there's a local function for assertion that's being exercised in two different ways. The optimisation test should be clear, simple and separate. 

I rewrote the tests in the "traditional" approach. 

> Also agree with @fsateler that the tests should fail if the optimisation is removed and yet they don't.

I think this is a good idea, however I dont know to test it. Can any one help me ?

MoreLinq.Test/TakeLastTest.cs
```diff
@@ -73,11 +57,29 @@ public void TakeLastDisposesSequenceEnumerator()
             {
                 seq.TakeLast(1).Consume();
             }
+        }
 
-            using (var seq = new[]{ 1,2,3 }.AsTestingSequence())
-            {
-                seq.TakeLast(1).Consume();
-            }
+        // ICollection<T> Optimization Tests
```
atifaziz: These tests don't test for _optimisation_ so the comment is misleading. They only test that the _behaviour_ of `TakeLast` does not change when passed a collection, which is why removing the optimisation in `TakeLast` does not fail them.

It seems futile to test that `TakeLast` is optimised for collections. The best one can do is to assert that the `Count` property was used, which doesn't seem like a very useful assurance. My main problem with 4ce583392328975ea5a5d197e893fdf829654a5f was that it seemed to abuse local functions when a simpler and more classic approach to have a (locally) reusable assertion function would be sufficient, like this:

```c#
static void AssertTakeLast<T>(IEnumerable<T> input, int count, IEnumerable<T> expected)
{
    // TODO comment why we are testing twice, once via an identity projection
    input.Select(x => x).TakeLast(count).AssertSequenceEqual(expected);
    input.TakeLast(count).AssertSequenceEqual(expected);
}

public void TakeLast()
{
    AssertTakeLast(new[] { 12, 34, 56, 78, 910, 1112 }, 3, new[] { 78, 910, 1112 });
}
```

f68966e54af292acfd2819583832c4694b669675 seems to add a lot of duplication of test data and code with little clarity.

leandromoh: > These tests don't test for optimisation so the comment is misleading. They only test that the behaviour of TakeLast does not change when passed a collection, which is why removing the optimisation in TakeLast does not fail them.

I see what you and @fsateler meant now, thanks for clarify.

> It seems futile to test that TakeLast is optimised for collections. The best one can do is to assert that the Count property was used, which doesn't seem like a very useful assurance.

I agree, it doesnt seem useful.

> My main problem with 4ce5833 was that it seemed to abuse local functions when a simpler and more classic approach to have a (locally) reusable assertion function would be sufficient

I agree. probrably I've got too excited about local functions that I didn't use it appropriately.
I will replace the current tests by that classic approach you show.

> Also agree with @fsateler that the tests should fail if the optimisation is removed and yet they don't.

how can I make a test that fail if the optimisation is removed?

leandromoh: > how can I make a test that fail if the optimisation is removed?

Any suggestion?

jnyrup:
A way to test it would be to make two classes that implements `ICollection` and just `IEnumerable` respectively, with a custom enumerator that counts the number of invocations of `MoveNext()` and exposes is as e.g. `public Iterations { get; private set; }`

For 
`myCustomCollection.TakeLast(k)` `myCustomCollection.Iteration` should be `k`,
while for 
`enumerable.TakeLast(k)` `enumerable.Iterations` should be `n`.

For inspiration see https://github.com/fluentassertions/fluentassertions/blob/develop/Tests/Shared.Specs/CollectionAssertionSpecs.cs#L4582

MoreLinq.Test/TakeLastTest.cs
```diff
@@ -57,5 +63,27 @@ public void TakeLastDisposesSequenceEnumerator()
                 seq.TakeLast(1).Consume();
             }
         }
+
+        [Test]
+        public void TakeLastOptimizedForCollections()
+        {
+            var collection = new OnceEnumerationCollection<int>(Enumerable.Range(1, 10));
```
atifaziz: We already have `UnenumerableList<>` that can be used here to test the same thing:

```c#
var collection = new UnenumerableList<int> { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
```

`OnceEnumerationCollection<>` should be removed.

MoreLinq.Test/TakeLastTest.cs
```diff
@@ -57,5 +63,27 @@ public void TakeLastDisposesSequenceEnumerator()
...
+
+            foreach (var _ in collection.TakeLast(3))
+            {
+                break;
+            }
+
+            Assert.IsTrue(collection.Count > 0);
+        }
+
+        static void AssertEqual<T>(ICollection<T> input, Func<IEnumerable<T>, IEnumerable<T>> op, Action<IEnumerable<T>> action)
```
atifaziz: `AssertEqual` is a misleading name for this method as it doesn't conduct an equality test itself. It leaves that up to `action` and in the case of `TakeLastWithNegativeCount`, the actual assertion is one of termination. I would suggest renaming this to `AssertTakeLast` or even `TakeLastDo` and since `op` is always going to be over `TakeLast`, change that parameter to be simply a count argument to be passed to `TakeLast`:

```c#
static void TakeLastDo<T>(ICollection<T> input, int count, Action<IEnumerable<T>> action)
{
    // Test that the behaviour does not change whether a collection
    // or a sequence is used as the source.

    action(input.TakeLast(count));
    action(input.Select(x => x).TakeLast(count));
}
```

leandromoh:
@atifaziz, I tested the code with .Net Framework 4.5 and .Net core 1.0 and both passed.

UnenumerableList and OnceEnumerationCollection have diferente propouses. The First cant Be used in a simple foreach. It has the restrictin that It must be consumed by its indexer. OnceEnumerationCollection can be consumed in both ways, but If and item is got by going through its IEnumerator, It is removed and its Count of the collection is decreased. Its Very similar to my EnumerateOnce method, but having the Count property avaible.

In this test, since the collection receive a source with length 10 and we use Takelast 3, the optmization becomes a Skip (7).

If Skip is not optmized for IList<>. It gonna iterate the my collection (what removes the 7 iterared  elements. So Count property of my collection is now 3.

If Skip is optmized, It doesn not consume the elements by iterativo so the Count remains 10.

atifaziz:
> `UnenumerableList `and `OnceEnumerationCollection `have diferente propouses. The First can not Be used in a simple foreach. It has the restriction that It must be consumed by its indexer. `OnceEnumerationCollection `can be consumed in both ways, but If the item is got by going through its `IEnumerator`, It is removed and its `Count `of the collection is **decreased**. Its Very similar to my [`AsEnumerateOnce`](https://github.com/morelinq/MoreLINQ/issues/318) method, but having the `Count `property avaible.

I understand the mechanics, but here `UnenumerableList<>` is more suitable.

> I am not sure why the test is failing

Could you kindly investigate then?

Perhaps you shouldn't rely on `Skip` at all and instead implement that logic in an _internal_ `Skip` that does optimize for lists?

atifaziz:
@leandromoh Is there a chance we can close this soon? If so, I'd like to schedule it for the [upcoming 2.8 release](https://github.com/morelinq/MoreLINQ/milestone/14).

atifaziz:
@fsateler LGTY?

leandromoh:
@atifaziz great! I would not suspect about that solution!  There is a chance to schedule also InserAt and MoveRange for 2.8 ?

atifaziz:
> There is a chance to schedule also InserAt and MoveRange for 2.8 ?

@leandromoh I doubted. My plate is full but good to finish them up early anyhow instead of waiting until close to the next deadline.

leandromoh:
@atifaziz I agree, but MoveRange is done since some months ago

MoreLinq.Test/TakeLastTest.cs
```diff
@@ -57,5 +66,22 @@ public void TakeLastDisposesSequenceEnumerator()
                 seq.TakeLast(1).Consume();
             }
         }
+
+        [Test]
+        public void TakeLastOptimizedForCollections()
+        {
+            var sequence = new UnenumerableList<int> { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
+
+            sequence.TakeLast(3).AssertSequenceEqual(8, 9, 10);
```
fsateler: Here we are assuming that the final implementation will indeed skip enumerating. Are we confident all framework versions will ever do that, and consider it broken until they do? Or maybe we should just test that TakeLast did its thing, which is to check the Count property before enumerating?

atifaziz: Not sure what you mean. The tests are now checking against all supported frameworks. Also, [`TakeLast` now internally relies on our `Slice`](https://github.com/morelinq/MoreLINQ/pull/304/commits/5eb30a2d70fbfaa294f2c23cb07085f9cd670441#diff-46c4b502587f8dc7793481ad0dbc74e8R55) (since 5eb30a2d70fbfaa294f2c23cb07085f9cd670441) instead of `Skip` so we are more sure about what's going to happen.

fsateler: Oh, right. I somehow missed this change, so scratch my comment.

LGTM

atifaziz: @fsateler Cool and thanks for the one last look nonetheless. :+1:

atifaziz:
@leandromoh Thanks!

atifaziz:
> I agree, but MoveRange is done since some months ago

@leandromoh Yes except I haven't had a chance to review your latest changes based on the reviews. If I can get round to it and all seems good to go then I'll try and release it earlier. This screams for the following bullet point from #209:

> - Increase the release cadence of the project by shipping possibly unstable work

Issue #312, opened 2017-06-08
PadLeft function


leandromoh:
Only to provide symmetry with Pad function

atifaziz:
Sure, go ahead. In addition to `PadLeft`, I would also consider the following names:

- `PadHead`
- `PrePad`

atifaziz:
Please note that this method was renamed to [`PadStart`](https://morelinq.github.io/2.6/ref/api/html/Overload_MoreLinq_MoreEnumerable_PadStart.htm) at time of [2.6 release in which it shipped](https://github.com/morelinq/MoreLINQ/releases/tag/v2.6.0).

PR #313, opened 2017-06-09
Padleft (issue #312) 

leandromoh:
This close issue #312

### DIFF
--- a/MoreLinq.Test/PadLeftTest.cs
+++ b/MoreLinq.Test/PadLeftTest.cs
@@ -1,0 +1,130 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2017 Leandro F. Vieira (leandromoh). All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq.Test
+{
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class PadLeftTest
+    {
+        // PadLeft(source, width)
+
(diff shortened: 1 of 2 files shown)


MoreLinq/PadLeft.cs
```diff
@@ -0,0 +1,162 @@
...
+                                     .Concat(col);
+                
+                return col;
+            }
+
+            return _(); IEnumerable<T> _()
+            {
+                using (var e = source.GetEnumerator())
+                {
+                    var list = new List<T>(width);
```
atifaziz: Why not use an array instead of a `List<T>` since it is always going to be a fixed size? It will also save an extra allocation.

MoreLinq/PadLeft.cs
```diff
@@ -0,0 +1,162 @@
...
+
+                    for (int i = 0; i < width && e.MoveNext(); i++)
+                    {
+                        list.Add(e.Current);
+                    }
+
+                    if (list.Count < width)
+                    {
```
atifaziz: Within this block, the enumerator (`e`) can be disposed early since the items are now all buffered in `list`.

MoreLinq.Test/PadLeftTest.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        [Test]
+        public void PadLeftIsLazy()
+        {
+            new BreakingSequence<int>().PadLeft(0);
+        }
+
+        [Test]
+        public void PadLeftWideSourceSequence()
```
atifaziz: Turn `PadLeftWideSourceSequence`, `PadLeftEqualSourceSequence` and `PadLeftNarrowSourceSequence` into test cases? Same for when testing the other overloads.

leandromoh:
@atifaziz all improviments were done! thanks for review.

atifaziz:
@leandromoh I've gone ahead and made a few tweaks, which I thought would be quicker than going over the review system.

I'm happy to revert 9b75a7c87c6bfeae518685bc6e87347a4e9afd91 if you or @fsateler find it more unreadable.

Can we also make a final call on the name? Here are some suggestions:

- `PadHead`
- `PrePad`
- `PadStart`

`PadLeft` wouldn't be entirely shocking but I think when it comes to sequences, it's more natural to think in terms of head & tail, pre- & post- or start & end rather than left & right.

leandromoh:
>  I think when it comes to sequences, it's more natural to think in terms of head & tail, pre- & post- or start & end rather than left & right.

I agree, but I vote for `PadLeft`, since user Immediately understand its intention.

Issue #315, opened 2017-06-13
SplitAt (split a sequence in two at a specified index)


atifaziz:
I'd like to propose a new operator that splits a sequence in two at a given index. The first sequence will contain all the items that appear before the index in the source sequence and the second will contain those items that appear at and after the index. A negative index would be allowed and would have the equivalent behavior of specifying an index of 0.

The follow table illustrates the splits that `SplitAt` will produce for a sample of indexes on a given source:

| Source                 | Index | Split 1                | Split 2                | 
|------------------------|-------|------------------------|------------------------| 
| [1,2,3,4,5,6,7,8,9,10] | 0     | []                     | [1,2,3,4,5,6,7,8,9,10] | 
| [1,2,3,4,5,6,7,8,9,10] | 2     | [1,2]                  | [3,4,5,6,7,8,9,10]     | 
| [1,2,3,4,5,6,7,8,9,10] | 5     | [1,2,3,4,5]            | [6,7,8,9,10]           | 
| [1,2,3,4,5,6,7,8,9,10] | 10    | [1,2,3,4,5,6,7,8,9,10] | []                     | 
| [1,2,3,4,5,6,7,8,9,10] | 20    | [1,2,3,4,5,6,7,8,9,10] | []                     | 
| [1,2,3,4,5,6,7,8,9,10] | -5    | []                     | [1,2,3,4,5,6,7,8,9,10] | 

## Prototype

```c#
public static (IEnumerable<T>, IEnumerable<T>)
    SplitAt<T>(this IEnumerable<T> source, int index) =>
    source.Index()
          .Partition(e => e.Key < index, (xs, ys) => (xs.Select(x => x.Value),
                                                      ys.Select(y => y.Value)));
```

## Example

```c#
var xs = Enumerable.Range(1, 10);
var results =
    from i in new[] { 0, 2, 5, 10, 20, -5 }
    select xs.SplitAt(i, (h, t) => new
    {
        Index = i,
        Head  = "[" + string.Join(",", h) + "]",
        Tail  = "[" + string.Join(",", t) + "]",
    });

foreach (var result in results)
    Console.WriteLine(test);
```

### Output

```
{ Index = 0, Head = [], Tail = [1,2,3,4,5,6,7,8,9,10] }
{ Index = 2, Head = [1,2], Tail = [3,4,5,6,7,8,9,10] }
{ Index = 5, Head = [1,2,3,4,5], Tail = [6,7,8,9,10] }
[...]

PR #316, opened 2017-06-14
SplitAt implementation

atifaziz:
This PR is an implementation of `SplitAt` as proposed in #315.

### DIFF
--- a/MoreLinq.Test/MoreLinq.Test.csproj
+++ b/MoreLinq.Test/MoreLinq.Test.csproj
@@ -51,7 +51,7 @@
       <Compile Include="SampleData.cs" />
       <Compile Include="Scope.cs" />
       <Compile Include="SequenceReader.cs" />
-      <Compile Include="TestingSequence.cs" />
+      <Compile Include="Testing*.cs" />
       <Compile Include="TestExtensions.cs" />
       <Compile Include="UnenumerableList.cs" />
       <Compile Include="VerifiableEnumerator.cs" />
--- a/MoreLinq.Test/SplitAtTest.cs
+++ b/MoreLinq.Test/SplitAtTest.cs
@@ -1,0 +1,118 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq.Test
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class SplitAtTest
+    {
+        [Test]
+        public void SplitAtIsLazy2()
+        {
+            var (first, second) = new BreakingSequence<object>().SplitAt(0);
+            Assert.That(first, Is.Not.Null);
+            Assert.That(second, Is.Not.Null);
+            Assert.That(first, Is.Not.SameAs(second));
+        }
+
+        [Test]
+        public void SplitAtWithResultSelectorIsLazy()
+        {
+            var (first, second) = new BreakingSequence<object>().SplitAt(0, ValueTuple.Create);
+            Assert.That(first, Is.Not.Null);
+            Assert.That(second, Is.Not.Null);
+            Assert.That(first, Is.Not.SameAs(second));
+        }
+
+        [TestCase( 0, new int[0]                             , new[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 })]
+        [TestCase( 2, new[] { 1, 2                          }, new[] { 3, 4, 5, 6, 7, 8, 9, 10       })]
+        [TestCase( 5, new[] { 1, 2, 3, 4, 5                 }, new[] { 6, 7, 8, 9, 10                })]
+        [TestCase(10, new[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }, new int[0]                             )]
+        [TestCase(20, new[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }, new int[0]                             )]
+        [TestCase(-5, new int[0]                             , new[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 })]
+        public void SplitAt(int index, int[] expected1, int[] expected2)
+        {
+            var xs = Enumerable.Range(1, 10).ToArray();
+
+            AssertParts(xs.AsTestingList()    , input => input.SplitAt(index, ValueTuple.Create));
+            AssertParts(xs.AsTestingSequence(), input => input.SplitAt(index, ValueTuple.Create));
+            AssertParts(xs.AsTestingList()    , input => input.SplitAt(index));
+            AssertParts(xs.AsTestingSequence(), input => input.SplitAt(index));
+
+            void AssertParts<T>(T input, Func<IEnumerable<int>, (IEnumerable<int>, IEnumerable<int>)> splitter)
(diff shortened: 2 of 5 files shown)


leandromoh:
There is missing a test for infinite sequence. Something like:

```
[Test]
public void SplitAtWithInfiniteSequence()
{
	var naturals = MoreEnumerable.Generate(1, x => x + 1);

	var (first, second) = naturals.SplitAt(10);

	Assert.That(first, Is.EquivalentTo(Enumerable.Range(1, 10)));
	Assert.That(second.Take(10), Is.EquivalentTo(Enumerable.Range(11, 10)));
}
```

atifaziz:
@leandromoh That test can never work. Note the remarks comments:

```c#
        /// <remarks>
        /// This method uses deferred execution semantics but iterates the
        /// <paramref name="source"/> entirely when either of the sequences
        /// resulting from split are iterated.
        /// </remarks>
```

Since the sequence will be iterated entirely, an infinite one will never end or stop when memory has been exhausted.

leandromoh:
I replaced your definition by the following and all tests passed, including that I comment for Infinite sequence.

```
public static TResult SplitAt<T, TResult>(
	this IEnumerable<T> source, int index,
	Func<IEnumerable<T>, IEnumerable<T>, TResult> resultSelector)
{
	if (resultSelector == null) throw new ArgumentNullException(nameof(resultSelector));

	if (index <= 0)
		return resultSelector(Enumerable.Empty<T>(), source);

	if (source is ICollection<T> col && index >= col.Count)
		return resultSelector(source, Enumerable.Empty<T>());

	var list =  new List<T>();
	var cached = list.Concat(source.AsEnumerateOnce().Pipe(x => list.Add(x)));

	return resultSelector(cached.TakeWhile((x, i) => i < index), cached.SkipWhile((x, i) => i < index));
}
```


where `AsEnumerateOnce` definition is 


```
private static IEnumerable<T> AsEnumerateOnce<T>(this IEnumerable<T> source)
{
	IEnumerator<T> ector = null;

	{
		ector = ector ?? source.GetEnumerator();

		while (ector.MoveNext())
			yield return ector.Current;

		ector.Dispose();
	}
}
```

atifaziz:
@leandromoh So you completely changed the implementation to make it work with infinite sources. I'd say it's a little detail you missed out [earlier](https://github.com/morelinq/MoreLINQ/pull/316#issuecomment-308775281). 😆 I have a hard time reasoning about the code now due to all the side-effects (`Pipe`) and will have inspect more closely. I also see a mini `Memoize` in there too. Perhaps we should finish #211 that seems to have stalled again? It's a noble goal to try and work with infinite sequences but your proposal does make the second part somewhat inefficient due to the `SkipWhile`. It may be acceptable for small values of `index` but not otherwise. Also the implementation I'm proposing returns lists for the two parts, which will be efficient for any operators optimized for lists or collection. However, I'm not sure anyone should count on that; I consider it an an implementation detail at the moment.

Thanks for your suggestions & ideas. Keep 'em coming!

leandromoh:
> @leandromoh So you completely changed the implementation to make it work with infinite sources. 

The idea was to create the simplest draft to pass in all tests. Complex code turns hard to reason about. As  C.A.R. Hoare quote: “**There are two ways of constructing a software design: One way is to make it so simple that there are obviously no deficiencies, and the other way is to make it so complicated that there are no obvious deficiencies. The first method is far more difficult.**”

>  I have a hard time reasoning about the code now due to all the side-effects (Pipe) and will have inspect more closely. 

Since my draft passed in all wroten test, I suppose it is a perfectly valid implementation.

>  I also see a mini Memoize in there too. Perhaps we should finish #211 that seems to have stalled again? 

Being honest, the fact of #211 had been submitted 6 or 7 months ago and it is not still be acceptable for merge made me a bit exhausted. I will be happy  when it be released, but for now, I think it is better dont account or wait for it.

> It's a noble goal to try and work with infinite sequences 

I think all LINQ operators **must** work with infinit sequences, except aggregation functions (`count`, `max`, `min`, `aggregate`, etc). otherwise the method has a limitation that is must not have. LINQ operators must be lazy as possible.

> your proposal does make the second part somewhat inefficient due to the SkipWhile

As I said, its only a draft without all possible optmizations. If we take it as the current implementation for this PR, we can add more optimizations for it, and In the **worst case**, we can apply the mini memoize to it too and turn it efficient;

> Since the sequence will be iterated entirely

I think it shoud be avoided, and the sequence must be as lazy iterated as possible. Imagine that source is **long**, for example 10.000 elements, and user just split it at 10 and consume only its 25 first elements (first = 10, and second = 15). the current implementation **iterate 9975 elements for nothing**. This dont make sense and user of course dont expect that. Of course, it is also not performatic.

> the implementation I'm proposing returns lists for the two parts, which will be efficient for any operators optimized for lists or collection. However, I'm not sure anyone should count on that; I consider it an an implementation detail at the moment.

I think it is more important that methods works with infinit sequences than to have optmizations. Users are not too concerned about optimizations but expect method works with an infinit list if he is not iterating it entirely.

atifaziz:
Love that quote by Hoare every time I read it.

> Since my draft passed in all wroten test, I suppose it is a perfectly valid implementation.

True, but aren't you assuming that the tests are perfect? 😃 Look how in #319 I made a [capital mistake](https://github.com/morelinq/MoreLINQ/pull/319#issuecomment-308818769) because I was missing a test (not a problem of coverage).

I agree with all your points about staying as lazy as possible. I realized the limitations and implementation choices, which is why I called them out in the remarks sections. I'll reconsider based on your feedback.

> Being honest, the fact of #211 had been submitted 6 or 7 months ago and it is not still be acceptable for merge made me a bit exhausted. I will be happy when it be released, but for now, I think it is better dont account or wait for it.

Sorry to hear that. If it's any consolation, it's been very consuming to review and fix.

> I will be happy when it be released, but for now, I think it is better dont account or wait for it.

That's a shame because a lot of work & thought went into it. I feel now that [my initial inclination to just go with the version in System.Interactive](https://github.com/morelinq/MoreLINQ/pull/211#issuecomment-259364631) may have been the right call.

leandromoh:
> Sorry to hear that. If it's any consolation, it's been very consuming to review and fix.

No problem. All we have a life outside github, with duties, and priorities.

> I feel now that my initial inclination to just go with the version in System.Interactive may have been the right call.

I think the problem is not about our work spent, but about define a MVP. The Memoize PR's release  is stopped 4 months only because ["Dispose should really mean Reset,"](https://github.com/morelinq/MoreLINQ/pull/211#issuecomment-303565368).

leandromoh:
@atifaziz  Any progress?

atifaziz:
@leandromoh Not much news, I'm afraid. 😞 I don't know if you followed #337, but I think [the idea of using a reader monad to express partial consumption](https://github.com/morelinq/MoreLINQ/issues/337#issuecomment-320438526) would work here with large and infinite sequences. For example:

```c#
var source = MoreEnumerable.Generate(0, x => x + 1);
var (xs, ys) = source.SplitAt(10, ys => ys.Take(25));
```

So `xs` will have ≤ 10 elements and `ys` will have ≤ 25 and at most 35 elements of `source` will have been consumed even though it's an infinite sequence.

atifaziz:
2a094a87fda3b268fb68b1a4954d7298b8b0ad93 updates the implementation based on [my revised proposal for #315](https://github.com/morelinq/MoreLINQ/issues/315#issuecomment-379003305) where split parts are returned as sub-sequences instead of a pivoted couple (tuple of 2) of parts.

:memo: Notes:

- Clean-up implementation
- Consider changing `offsets` type to `IEnumerable<int>` (`params int[]` was out of laziness)
- Add more test cases of various offsets
- Optionally, optimise for corner cases where `source` can be returned verbatim, though this can be done post-merge of this PR

leandromoh:
I really likes this new implementation... How about receive a predicate (Func<T, int, book>) to indicate a split? This allow user split by other criterious. You can write the Index overload in terms of this new overload wrapping int[] offesets with the predicate. 

Ex.: (e, Index) => offsets.Contains(Index).

atifaziz:
> I liked a lot this new implementation.

@leandromoh Glad to hear.

> How about receive a predicate (Func<T, int, bool>) to indicate a split? This allow user split by other criterious. 

It's an idea but have to be careful about avoiding confusion with `Split` and `Segment`.

leandromoh:
I did not know that Segment can receive an predicate. Why dont write SplitAt in terms of Segment?

atifaziz:
> Why dont write SplitAt in terms of Segment?

It would almost work but it can't be used because `Segment` doesn't _always_ produce a sub-sequence. With `SplitAt`, if the source sequence is too short then you will still get a blank sub-sequence.

Orace:
`Dispatch` #633 may have been used to implement this.

PR #319, opened 2017-06-15
Flatten consecutive chaining of Concat & Prepend

atifaziz:
This PR optimises [`Prepend`](https://morelinq.github.io/1.x/ref/api/html/M_MoreLinq_MoreEnumerable_Prepend__1.htm) and [`Concat`](https://morelinq.github.io/1.x/ref/api/html/Overload_MoreLinq_MoreEnumerable_Concat.htm) such that if they are chained to the same operation or either operation in succession then there's a flat (as opposed to nested today) iteration that takes place.

Before this PR, calling `xs.Prepend(2).Prepend(1).Concat(10)` will create 3 distinct enumerables, which when iterated will make nested (`MoveNext`) calls into each iterator. Done in a loop (e.g. `xs = xs.Concat(x);` and the iteration performance will explode. With this PR, there will be a single enumerable resulting across all _consecutive_ calls and the final iteration will be flat (a single call through `MoveNext`).


fsateler:
>Before this PR, calling xs.Prepend(2).Prepend(1).Concat(10) will create 3 distinct enumerables, which when iterated will make nested (MoveNext) calls into each iterator. Done in a loop (e.g. xs = xs.Concat(x); and the iteration performance will explode. With this PR, there will be a single enumerable resulting across all consecutive calls and the final iteration will be flat (a single call through MoveNext).

I like the idea, but the implementation is problematic due to the shared buffer:

```csharp
var firstSequence = new [] { 1, }.Concat(2);
var second = firstSequence.Concat(3);
var third = firstSequence.Concat(4);

third.Dump(); // prints 1, 2, 4
second.Dump(); // also prints 1, 2, 4
```

Usually LINQ operators are immutable, in that applying an additional operator does not alter the source sequence, and this PR breaks this expectation.

fsateler:
Using [ImmutableList](https://www.nuget.org/packages/System.Collections.Immutable/) would fix the issue, but that would add a dependency.

atifaziz:
@fsateler That's a great catch! Pity that the tests didn't pick it up. Goes to show you'll always miss something so thanks for the review.

BTW, `third.Dump()` yields [1, 2, 3, 4].

atifaziz:
@fsateler I have added tests and fixed the sharing bug. The implementation may be suboptimal, especially for concatenations, but that can be tweaked and improved with time (unless you have some suggestions right away).

atifaziz:
@fsateler Are you happy with this now?

atifaziz:
@fsateler I want to get this into [the June release](https://github.com/morelinq/MoreLINQ/milestone/8) that I was planning for today. I'm assuming you haven't found the time to finish your review but I'll go ahead and merge nonetheless and if you see a glaring issue later then it can always be addressed in a 2.6 patch version.

Issue #311, opened 2017-06-03
Move: move range of elements to a new offset


leandromoh:
It pick up a range of a sequence and put it at some other index. 
It's a combination of `Exclude + Slice + InsertAt`, but the input sequence is enumerated once.

Signature:
`static IEnumerable<T> SwapRange<T>(this IEnumerable<T> source, int index, int count, int putAt)`

A naive implementation (of this initial propose) would be

```
static IEnumerable<T> SwapRange<T>(this IEnumerable<T> source, int index, int count, int putAt)
{
    var slc = source.Slice(index, count);
    var exc = source.Exclude(index, count);

    return exc.InsertAt(slc, putAt);
}
```

Example:
`Enumerable.Range(0, 10).SwapRange(5, 2, 7) // [0, 1, 2, 3, 4, 7, 8, 5, 6, 9]`

Since the [InsertAt](https://github.com/morelinq/MoreLINQ/issues/306) function it is not complete defined yet, this propose is also not , so suggestions for the behaviour of the method is very welcome!

leandromoh:
I was thinking, Since the objective of `SwapRange` is just to move a range within to another place, that is, **dont add new elements or remove (preserving source's length)**, only swap, i think make sense that 

`Enumerable.Range(0, 10).SwapRange(5, 2, 999)` returns `[0, 1, 2, 3, 4, 7, 8, 9 , 5, 6]`

I would like to submit a PR for this.
What do you think?

atifaziz:
> I suggest submit a PR that throws, and latter if we think valid, turn it less strict (accpet negative index) with another PR.

Yes, let's start with the strict version and we can relax the rules in a later version.

leandromoh:
Ok, I Will submit the PR

PR #323, opened 2017-06-20
Move implementation  (issue #311) 

leandromoh:
PR for #311


MoreLinq/SwapRange.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (index < 0) throw new ArgumentOutOfRangeException(nameof(index), "Index cannot be negative.");
+            if (count < 0) throw new ArgumentOutOfRangeException(nameof(count), "Count cannot be negative.");
+            if (putAt < 0) throw new ArgumentOutOfRangeException(nameof(putAt), "PutAt cannot be negative.");
+
+            if (putAt == index || count == 0)
+                return source;
+
+            if (putAt < index)
```
fsateler: I don't agree that decision trees based on nested `?:` operators are easy to follow, as precedence is not intuitively obvious (even if unambiguous in the language). But YMMV of course. I think we discussed this earlier but I can't remember where and if we reached a conclusion.

In general, I find early exit pattern to be highly readable: list all 'special' conditions with their special return value, and then have the meat of the process without handling those cases.

MoreLinq/SwapRange.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// This operator uses deferred execution and streams its results.
+        /// </remarks>
+        /// <example>
+        /// <code>
+        /// var result = Enumerable.Range(0, 6).SwapRange(3, 2, 0);
+        /// </code>
+        /// The <c>result</c> variable will contain <c>{ 3, 4, 0, 1, 2, 5 }</c>.
+        /// </example>
+
+        public static IEnumerable<T> SwapRange<T>(this IEnumerable<T> source, int index, int count, int putAt)
```
fsateler: I would also rename `index` and `count` to `sourceIndex` and `sourceCount` (or `oldIndex` and `oldCount`) for symmetry.

MoreLinq/SwapRange.cs
```diff
@@ -0,0 +1,85 @@
...
+                return source;
+
+            if (putAt < index)
+                return _(putAt, index - putAt, count);
+            else
+                return _(index, count, putAt - index);
+
+            IEnumerable<T> _(int startBuffer, int bufferSize, int beforeYieldBuffer)
+            {
+                using (var ector = source.GetEnumerator())
```
leandromoh: Done! 
**ector** is something like an abreviation for "**_e_** numera **_ctor_**"

leandromoh:
@atifaziz  all adjusts done (except that about ternary operator, because I agree with @fsateler ).

MoreLinq/SwapRange.cs
atifaziz: But `count` remains constant so seems somewhat superfluous to call it `sourceCount`.

MoreLinq/SwapRange.cs
atifaziz: > In general, I find early exit pattern to be highly readable: list all 'special' conditions with their special return value, and then have the meat of the process without handling those cases.

I agree, that's how I do it but when I'm thinking _imperatively_ and wish to avoid code snake dance of complex if-else tree causing side-effects. The meat here, however, is in the local iterator function so what's left is a return expression. I think it's good to consider each case separately. By the way, #304 also [uses the conditional expression](https://github.com/leandromoh/MoreLINQ/blob/f0168dfc828d143efb16f90880e1ccebfef028d0/MoreLinq/TakeLast.cs#L55-L57).

Can we strike a compromise? :smile: Either make the tree obvious via indentation:

```c#
return newIndex == oldIndex || count == 0
     ? source
     : newIndex < oldIndex
       ? _(newIndex, oldIndex - newIndex, count)
       : _(oldIndex, count, newIndex - oldIndex);
```

or use the condition expression just for the call to the local function:

```c#
if (newIndex == oldIndex || count == 0)
    return source;

return newIndex < oldIndex
     ? _(newIndex, oldIndex - newIndex, count)
     : _(oldIndex, count, newIndex - oldIndex);
```

This would also make it consistent with [what we have today in `SkipLast`](https://github.com/morelinq/MoreLINQ/blob/v2.5.0/MoreLinq/SkipLast.cs#L40-L46).

And if it's still distasteful then I'm happy to let go.

MoreLinq/MoveRange.cs
```diff
@@ -0,0 +1,85 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
```
atifaziz: Import is unused.

MoreLinq.Test/MoveRangeTest.cs
```diff
@@ -0,0 +1,121 @@
...
+        [TestCase(10, 0,  5)]
+        [TestCase(10, 5,  0)]
+        [TestCase(10, 6,  1)]
+        [TestCase(10, 1,  6)]
+        [TestCase(10, 0, 10)]
+        [TestCase(10, 10, 0)]
+        [TestCase(10, 3, 10)]
+        [TestCase(10, 10, 3)]
+        [TestCase(10, 99, 2)]
+        public void MoveRange(int length, int index, int count)
```
atifaziz: It's not obvious what's going on here really. The `length` argument is always 10 and there are no expectations in the test data. Later in the test method, expectations are generated dynamically so there's mix of strategies here. If you want to generate test cases, I recommend you do so with a [test case source](https://github.com/nunit/docs/wiki/TestCaseData), or see the [`ConcatMany` test](https://github.com/atifaziz/MoreLINQ/blob/1afdf8394fa18ce6af146b7a816d6843f62616b4/MoreLinq.Test/ConcatTest.cs#L101-L119) for an example.

MoreLinq.Test/MoveRangeTest.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public void MoveRangeWithSequenceShorterThanNewIndex()
+        {
+            const int length = 10;
+            const int index = 5;
+            const int count = 2;
+
+            var source = Enumerable.Range(0, length);
+
+            Enumerable.Range(length, length + 5).ForEach(newIndex =>
```
atifaziz: Ditto as for my previous comment about test case generation.

MoreLinq.Test/MoveRangeTest.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        [Test]
+        public void MoveRangeWithCountEqualsZero()
+        {
+            var source = Enumerable.Range(0, 10);
+            var result = source.MoveRange(5, 0, 5);
+            
+            Assert.That(source, Is.SameAs(result));
+        }
+
```
atifaziz: Remove blank line (has tabs).

MoreLinq/MoveRange.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// This operator uses deferred execution and streams its results.
+        /// </remarks>
+        /// <example>
+        /// <code>
+        /// var result = Enumerable.Range(0, 6).MoveRange(3, 2, 0);
+        /// </code>
+        /// The <c>result</c> variable will contain <c>{ 3, 4, 0, 1, 2, 5 }</c>.
+        /// </example>
+
+        public static IEnumerable<T> MoveRange<T>(this IEnumerable<T> source, int oldIndex, int count, int newIndex)
```
atifaziz: This is about moving, `fromIndex` and `toIndex` go better than `oldIndex` and `newIndex`.

MoreLinq/MoveRange.cs
```diff
@@ -0,0 +1,85 @@
...
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Take a range of the source sequence and moves it to another position.
```
atifaziz: Re-word to:

> Return a sequence with a range of elements in the source sequence moved to a new offset.

MoreLinq/MoveRange.cs
```diff
@@ -0,0 +1,85 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Take a range of the source sequence and moves it to another position. 
+        /// </summary>
+        /// <typeparam name="T">Type of the source sequence</typeparam>
```
atifaziz: Missing a period (.) at the end.

MoreLinq/MoveRange.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <param name="source">The source sequence.</param>
+        /// <param name="oldIndex">The zero-based and original index where the range to be moved begins.</param>
+        /// <param name="count">The count of items to move.</param>
+        /// <param name="newIndex">The index where the specified range will be placed.</param>
+        /// <returns>The sequence with the specified range moved to the new position.</returns>
```
atifaziz: Instead of “The sequence”, say, “A sequence”.

MoreLinq/MoveRange.cs
```diff
@@ -0,0 +1,85 @@
...
+    using System.Linq;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Take a range of the source sequence and moves it to another position. 
+        /// </summary>
+        /// <typeparam name="T">Type of the source sequence</typeparam>
+        /// <param name="source">The source sequence.</param>
+        /// <param name="oldIndex">The zero-based and original index where the range to be moved begins.</param>
```
atifaziz: Re-word to:

> The zero-based index identifying the first element in the range of elements to move.

MoreLinq/MoveRange.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <param name="count">The count of items to move.</param>
+        /// <param name="newIndex">The index where the specified range will be placed.</param>
```
atifaziz: Change _placed_ to _moved_ for consistency of vocabulary.

leandromoh:
@atifaziz  All adjusts done! Thanks for show me about "test case source", I didnt know about it.

leandromoh:
any problem with this PR ?

leandromoh:
@atifaziz done!

atifaziz:
> take a look at `InsertAt` too, I think it may ship in the 2.8v

@leandromoh I'll take a final look after @fsateler has cleared his review.

PR #333, opened 2017-07-13
Add From methods that can create sequences from functions

fsateler:
This builds on top of #331 so that should be reviewed and merged first.

### DIFF
--- a/MoreLinq.Test/FromTest.cs
+++ b/MoreLinq.Test/FromTest.cs
@@ -20,6 +20,7 @@ namespace MoreLinq.Test
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using NUnit.Framework;
 
     class FromTest
@@ -81,14 +82,10 @@ namespace MoreLinq.Test
             results.Consume();
             results.Consume();
 
-            for (int i = 0; i < numArgs; i++)
-            {
-                Assert.That(evals[i], Is.EqualTo(3));
-            }
-            for (int i = numArgs; i < 4; i++)
-            {
-                Assert.That(evals[i], Is.EqualTo(0));
-            }
+            // numArgs functions were evaluated...
+            evals.Take(numArgs).AssertSequenceEqual(Enumerable.Repeat(3, numArgs));
+            // safety check: we didn't evaluate functions past numArgs
+            evals.Skip(numArgs).AssertSequenceEqual(Enumerable.Repeat(0, evals.Length - numArgs));
         }
     }
 }
--- a/MoreLinq.Test/FromTest.cs
+++ b/MoreLinq.Test/FromTest.cs
@@ -40,38 +40,19 @@ namespace MoreLinq.Test
         [TestCase(4)]
         public void TestFromInvokesMethods(int numArgs)
         {
-            var factories = new Func<int>[]
-            {
-                () => -2,
-                () => 4,
-                () => int.MaxValue,
-                () => int.MinValue,
-            };
-            IEnumerable<int> results;
-            int[] expectedResults;
+            int F1() => -2;
+            int F2() => 4;
+            int F3() => int.MaxValue;
+            int F4() => int.MinValue;
+
             switch (numArgs)
             {
-                case 1:
-                    results = MoreEnumerable.From(factories[0]);
-                    expectedResults = new[] { factories[0]() };
-                    break;
-                case 2:
-                    results = MoreEnumerable.From(factories[0], factories[1]);
-                    expectedResults = new[] { factories[0](), factories[1]() };
-                    break;
-                case 3:
-                    results = MoreEnumerable.From(factories[0], factories[1], factories[2]);
-                    expectedResults = new[] { factories[0](), factories[1](), factories[2]() };
-                    break;
-                case 4:
-                    results = MoreEnumerable.From(factories[0], factories[1], factories[2], factories[3]);
-                    expectedResults = new[] { factories[0](), factories[1](), factories[2](), factories[3]() };
-                    break;
-                default:
-                    throw new ArgumentOutOfRangeException(nameof(numArgs));
+                case 1: MoreEnumerable.From(F1).AssertSequenceEqual(F1()); break;
+                case 2: MoreEnumerable.From(F1, F2).AssertSequenceEqual(F1(), F2()); break;
+                case 3: MoreEnumerable.From(F1, F2, F3).AssertSequenceEqual(F1(), F2(), F3()); break;
+                case 4: MoreEnumerable.From(F1, F2, F3, F4).AssertSequenceEqual(F1(), F2(), F3(), F4()); break;
+                default: throw new ArgumentOutOfRangeException(nameof(numArgs));
             }
-
-            results.AssertSequenceEqual(expectedResults);
         }
 
         [TestCase(1)]
@@ -81,20 +62,18 @@ namespace MoreLinq.Test
         public void TestFromInvokesMethodsMultipleTimes(int numArgs)
         {
             var evals = new [] { 0, 0, 0, 0 };
-            var factories = new Func<int>[]
-            {
-                () => { evals[0]++; return -2; },
-                () => { evals[1]++; return -2; },
-                () => { evals[2]++; return -2; },
-                () => { evals[3]++; return -2; },
-            };
+            int F1() { evals[0]++; return -2; }
+            int F2() { evals[1]++; return -2; }
+            int F3() { evals[2]++; return -2; }
+            int F4() { evals[3]++; return -2; }
+
             IEnumerable<int> results;
             switch (numArgs)
             {
-                case 1: results = MoreEnumerable.From(factories[0]); break;
-                case 2: results = MoreEnumerable.From(factories[0], factories[1]); break;
-                case 3: results = MoreEnumerable.From(factories[0], factories[1], factories[2]); break;
-                case 4: results = MoreEnumerable.From(factories[0], factories[1], factories[2], factories[3]); break;
+                case 1: results = MoreEnumerable.From(F1); break;
+                case 2: results = MoreEnumerable.From(F1, F2); break;
+                case 3: results = MoreEnumerable.From(F1, F2, F3); break;
+                case 4: results = MoreEnumerable.From(F1, F2, F3, F4); break;
                 default: throw new ArgumentOutOfRangeException(nameof(numArgs));
             }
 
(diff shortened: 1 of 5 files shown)


MoreLinq.Test/FromTest.cs
```diff
@@ -0,0 +1,125 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+
+namespace MoreLinq.Test
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text;
```
atifaziz: Remove unused import.

MoreLinq.Test/FromTest.cs
```diff
@@ -0,0 +1,125 @@
...
+            results.AssertSequenceEqual(expectedResults);
+        }
+
+        [TestCase(1)]
+        [TestCase(2)]
+        [TestCase(3)]
+        [TestCase(4)]
+        public void TestFromInvokesMethodsMultipleTimes(int numArgs)
+        {
+            var evals = new int[] { 0, 0, 0, 0 };
```
atifaziz: FYI, `int` specification is redundant here.

MoreLinq.Test/FromTest.cs
```diff
@@ -0,0 +1,125 @@
...
+            var factories = new Func<int>[]
+            {
+                () => { evals[0]++; return -2; },
+                () => { evals[1]++; return -2; },
+                () => { evals[2]++; return -2; },
+                () => { evals[3]++; return -2; },
+            };
+            IEnumerable<int> results;
+            switch (numArgs)
```
atifaziz: This switch reads very long (nitpicking on readability) with lots of whitespace breaks when what we want to say is simply:

```c#
var results
    = numArgs == 1 ? MoreEnumerable.From(factories[0])
    : numArgs == 2 ? MoreEnumerable.From(factories[0], factories[1])
    : numArgs == 3 ? MoreEnumerable.From(factories[0], factories[1], factories[2])
    : numArgs == 4 ? MoreEnumerable.From(factories[0], factories[1], factories[2], factories[3])
    : throw new ArgumentOutOfRangeException(nameof(numArgs));
```

Even if you wish to maintain switch, I would suggest reformatting as follows:

```c#
switch (numArgs)
{
    case 1: results = MoreEnumerable.From(factories[0]); break;
    case 2: results = MoreEnumerable.From(factories[0], factories[1]); break;
    case 3: results = MoreEnumerable.From(factories[0], factories[1], factories[2]); break;
    case 4: results = MoreEnumerable.From(factories[0], factories[1], factories[2], factories[3]); break;
    default: throw new ArgumentOutOfRangeException(nameof(numArgs));
}
```

MoreLinq/From.cs
```diff
@@ -0,0 +1,126 @@
...
+#endregion
+
+
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    partial class MoreEnumerable
```
atifaziz: Review documentation for missing periods at the end of sentences.

MoreLinq/From.cs
```diff
@@ -0,0 +1,126 @@
...
+    {
+        /// <summary>
+        /// Returns a single-element sequence containing the result of invoking the function function
```
atifaziz: Repeated: function

MoreLinq/From.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This operator uses deferred execution and streams the results.
+        /// If the resulting sequence is enumerated multiple times, the function will be
+        /// invoked multiple times too.
+        /// </remarks>
+        /// <typeparam name="T">The type of the object returned by the function</typeparam>
+        /// <param name="function">The function to evaluate</param>
+        /// <returns>A sequence with the value created from <paramref name="function"/></returns>
```
atifaziz: Instead of:

> A sequence with the value created from…

say:

> A sequence with the value resulting from invoking…

Ditto for other overloads.

fsateler:
> There is a fundamental inconsistency between the overloads of From. The ones taking individual functions as arguments don't allow for null references. The one taking the variable array of functions, however, does! 

Good catch! Indeed, this needs to be resolved. Given that we can reasonably expect callers to be creating relatively short sequences, I think we should check for null in the `params` overload too. What do you think?

I pushed commits fixing your comments. I also picked up the change you suggested for [`Evaluate` docstring](https://github.com/morelinq/MoreLINQ/pull/331#discussion_r130069652).

> This switch reads very long (nitpicking on readability)

Maybe it would be better to split the testcase into separate tests...

atifaziz:
> I'm not much of a fan of nested `? :` operators.

Sometimes, there's no choice, e.g., when using anonymous types:

```c#
[TestCase(1)]
[TestCase(2)]
[TestCase(3)]
[TestCase(4)]
public void TestFromInvokesMethods(int numArgs)
{
    int F1() => -2;
    int F2() => 4;
    int F3() => int.MaxValue;
    int F4() => int.MinValue;

    var r
        = numArgs == 1 ? new { Actuals = MoreEnumerable.From(F1)            , Expectations = new[] { F1()                   } }
        : numArgs == 2 ? new { Actuals = MoreEnumerable.From(F1, F2)        , Expectations = new[] { F1(), F2()             } }
        : numArgs == 3 ? new { Actuals = MoreEnumerable.From(F1, F2, F3)    , Expectations = new[] { F1(), F2(), F3()       } }
        : numArgs == 4 ? new { Actuals = MoreEnumerable.From(F1, F2, F3, F4), Expectations = new[] { F1(), F2(), F3(), F4() } }
        : throw new ArgumentOutOfRangeException(nameof(numArgs));

    r.Actuals.AssertSequenceEqual(r.Expectations);
}
```

Now if you do the assertions directly in the switch cases, then it makes more sense and perhaps reads equally simpler:

```c#
[TestCase(1)]
[TestCase(2)]
[TestCase(3)]
[TestCase(4)]
public void TestFromInvokesMethods(int numArgs)
{
    int F1() => -2;
    int F2() => 4;
    int F3() => int.MaxValue;
    int F4() => int.MinValue;

    switch (numArgs)
    {
        case 1: MoreEnumerable.From(F1).AssertSequenceEqual(F1()); break;
        case 2: MoreEnumerable.From(F1, F2).AssertSequenceEqual(F1(), F2()); break;
        case 3: MoreEnumerable.From(F1, F2, F3).AssertSequenceEqual(F1(), F2(), F3()); break;
        case 4: MoreEnumerable.From(F1, F2, F3, F4).AssertSequenceEqual(F1(), F2(), F3(), F4()); break;
        default: throw new ArgumentOutOfRangeException(nameof(numArgs));
    }
}
```

Perhaps the second version is a good compromise?

---
PS Note also use of local functions to simplify things further since.

fsateler:
> Generally, LINQ operators don't validate the elements of a sequence and pass-through any exceptions. However, if we want to make a special case here then it can get tricky. What's the error you would throw? Also, it could be surprising to see Evaluate not behaving in the same manner and if we want to render it consistent with From then it would end up scanning it source sequence before use. In light of all this, I'm inclined to say at the moment that null functions should be accepted and break when invoked.

OK, your argument makes sense to me. I have therefore removed those checks. However, now the null checks fail.

However now the NullArgumentTest fails because it breaks the convention... Should we add an explicit waiver to `From` somewhere in `NullArgumentTest`?

atifaziz:
> However now the NullArgumentTest fails because it breaks the convention... Should we add an explicit waiver to `From` somewhere in `NullArgumentTest`?

Aye!

fsateler:
Done.

MoreLinq.Test/NullArgumentTest.cs
```diff
@@ -112,7 +112,18 @@ static bool CanBeNull(ParameterInfo parameter)
...
                 select t.GetTypeInfo();
-            var nullableParameters = new[] { "Assert.errorSelector", "ToDataTable.expressions", "ToDelimitedString.delimiter", "Trace.format" };
+            var nullableParameters = new[]
+            {
+                "Assert.errorSelector",
+                "From.function",
+                "From.function1",
+                "From.function2",
+                "From.function3",
+                "From.function4",
```
atifaziz: Does this case exist?

fsateler: Ehm, no it doesn't. I got too carried away with copypaste, it seems.

MoreLinq.Test/FromTest.cs
```diff
@@ -0,0 +1,94 @@
...
+                case 3: results = MoreEnumerable.From(F1, F2, F3); break;
+                case 4: results = MoreEnumerable.From(F1, F2, F3, F4); break;
+                default: throw new ArgumentOutOfRangeException(nameof(numArgs));
+            }
+
+            results.Consume();
+            results.Consume();
+            results.Consume();
+
+            for (int i = 0; i < numArgs; i++)
```
atifaziz: Any reason to favour a loop here over the following?

```c#
evals.Take(numArgs).AssertSequenceEqual(Enumerable.Repeat(3, numArgs));
```

MoreLinq.Test/FromTest.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                Assert.That(evals[i], Is.EqualTo(3));
+            }
+            for (int i = numArgs; i < 4; i++)
```
atifaziz: Any reason to favour a loop here over the following?

```c#
evals.Skip(numArgs).AssertSequenceEqual(Enumerable.Repeat(0, evals.Length - numArgs));
```

MoreLinq.Test/FromTest.cs
fsateler: I find the explicit loop to be easier to read in this case. However, I can change it if you prefer the linq version.

atifaziz: Funny that I found the loop harder to read during the review. It slowed me down and I had to pause to mentally execute and make sure that all the moving parts of the loop are correct.

Setting aside our subjective views of the readability of each version, my (hopefully) objective argument would be that using LINQ makes the test safer and less prone to accidental edits. For example, if someone accidentally changed the `[i]` in `Assert.That(evals[i], Is.EqualTo(3))` to a `[0]` (reading `Assert.That(evals[0], Is.EqualTo(3))`)  then the test would still pass and yield a false positive. With the LINQ version, it would be a lot harder.

A loop inside the actual implementation (i.e. not the unit test) would be more acceptable for optimisation purposes.

fsateler: > Funny that I found the loop harder to read during the review.

Heh, I guess readability is in the eye of the beholder 😉 . It's not much of a difference to me, so I changed it.

PR #335, opened 2017-07-31
Added WhereIf (Implementation, xml-docs, tests, readme.md)

ysmoradi:


atifaziz:
@ysmoradi Thanks but I think what would be more generally useful is an _if-else_ operator instead of combining the _where_ and the _if-else_:

```c#
public static IEnumerable<TResult> If<T, TResult>(this IEnumerable<T> source,
    bool condition,
    Func<IEnumerable<T>, IEnumerable<TResult>> trueCase,
    Func<IEnumerable<T>, IEnumerable<TResult>> falseCase) =>
    condition ? trueCase(source) : falseCase(source);
```

The usage would look like this:

```c#
var sequence = Enumerable.Range(1, 10);
var result = sequence.If(true, xs => xs.Where(x => x % 2 == 0), xs => xs);
Console.WriteLine(string.Join(", ", result)); // 2, 4, 6, 8, 10
```

Here's another example of its general applicability:

```c#
var evens  = true;
var desc   = false;
var result =
    Enumerable.Range(1, 10)
              .If(evens, xs => xs.Where(x => x % 2 == 0),
                         xs => xs)
              .If(!desc, xs => xs.OrderBy(x => x),
                         xs => xs.OrderByDescending(x => x));
Console.WriteLine(string.Join(", ", result));
```

[System.Interactive](https://www.nuget.org/packages/System.Interactive/) already has an [`If`](https://github.com/Reactive-Extensions/Rx.NET/blob/v3.1.1/Ix.NET/Source/System.Interactive/If.cs) implementation available today so I don't think anything needs to be added to MoreLINQ.

ysmoradi:
@atifaziz Yeah, I'm agree with you. Next time I try to add more useful one (:

atifaziz:
> Next time I try to add more useful one (:

Would be also best to propose an enhancement by discussing it as an issue first instead of going straight to a PR. That way, you don't waste time coding up a whole implementation and tests or risk being disheartened if your contribution doesn't get merged in the end.

ysmoradi:
Yeah, we should do that. I had WhereIf and bunch of other codes with their unit tests in my own framework. So I copied one of them easily to see what happens next. I'm analyzing my next step and I'll discuss it using a new issue.

PR #340, opened 2017-08-08
FallbackIfEmpty: preserve source and fallback collections if possible

fsateler:


MoreLinq/FallbackIfEmpty.cs
```diff
@@ -162,45 +161,45 @@ public static IEnumerable<T> FallbackIfEmpty<T>(this IEnumerable<T> source, IEnu
...
+                        yield break;
                     }
                 }
+
+                foreach (var item in fallbackChooser())
+                    yield return item;
+            }
+            IEnumerable<T> fallbackChooser()
+            {
+                if (count > 0 && count <= 4) {
```
atifaziz: Why not make this as simple as predicating on `fallback` being null or not?

```c#
IEnumerable<T> fallbackChooser() => fallback ?? instancesFallback();
```

MoreLinq/FallbackIfEmpty.cs
```diff
@@ -162,45 +161,45 @@ public static IEnumerable<T> FallbackIfEmpty<T>(this IEnumerable<T> source, IEnu
...
+                    return instancesFallback();
+                }
+                else {
+                    return fallback;
+                }
+            }
+            IEnumerable<T> instancesFallback()
```
atifaziz: This could be further localized to your `fallbackChooser`.

Local functions, like methods, should still use the PascalCase naming convention.

MoreLinq/FallbackIfEmpty.cs
```diff
@@ -162,45 +161,45 @@ public static IEnumerable<T> FallbackIfEmpty<T>(this IEnumerable<T> source, IEnu
...
+                        do { yield return e.Current; }
+                        while (e.MoveNext());
+                        yield break;
                     }
                 }
+
+                foreach (var item in fallbackChooser())
+                    yield return item;
+            }
+            IEnumerable<T> fallbackChooser()
```
atifaziz: Local functions, like methods, should still use the PascalCase naming convention.

MoreLinq/FallbackIfEmpty.cs
```diff
@@ -162,45 +161,45 @@ public static IEnumerable<T> FallbackIfEmpty<T>(this IEnumerable<T> source, IEnu
...
-                        yield return fallback1;
-                        if (count > 1) yield return fallback2;
-                        if (count > 2) yield return fallback3;
-                        if (count > 3) yield return fallback4;
-                    }
-                    else
-                    {
-                        foreach (var item in fallback)
-                            yield return item;
+                using (var e = source.GetEnumerator()) {
```
fsateler: Arrgh. I really need to finish up the editorconfig so that this stops becoming an issue.

MoreLinq/FallbackIfEmpty.cs
fsateler: Because we don't check that `(count == 0 || count > 4) => (fallback == null)`. I didn't want to introduce a nullness requirement, when previously all that was required was that it be non-null when it was to be used.

MoreLinq.Test/FallbackIfEmptyTest.cs
```diff
@@ -39,6 +39,28 @@ public void FallbackIfEmptyWithEmptySequence()
         }
 
         [Test]
+        public void FallbackIfEmptyPreservesSourceCollectionIfPossible()
+        {
+            var source = new int[] { 1 };
+            // ReSharper disable PossibleMultipleEnumeration
+            Assert.AreSame(source.FallbackIfEmpty(12), source);
```
fsateler: I did this on ccca164 , but I'm not sure it ended up being better, as it is a bunch of extra code noise.

MoreLinq/FallbackIfEmpty.cs
atifaziz: Not sure I understand you here. Could you re-phrase?

MoreLinq.Test/FallbackIfEmptyTest.cs
atifaziz: I agree that ccca164c20d889716ad54e22b75275e6beb24c4b isn't any better. What I had more in mind was this:

```c#
[TestCase(new[] { 12                     })]
[TestCase(new[] { 12, 23                 })]
[TestCase(new[] { 12, 23, 34             })]
[TestCase(new[] { 12, 23, 34, 45         })]
[TestCase(new[] { 12, 23, 34, 45, 56     })]
[TestCase(new[] { 12, 23, 34, 45, 56, 67 })]
public void FallbackIfEmptyPreservesSourceCollectionIfPossible(int[] fallback)
{
    var source = new[] { 1 };
    Assert.AreSame(source.FallbackIfEmpty(fallback), source);
}
```

fsateler: But those test cases don't add much value, as they all test the same overload. The idea is to test that all overloads preserve the source collection.

atifaziz: Duh! My bad. All the overloading made it all very opaque at the call site. I guess in that case we can revert ccca164c20d889716ad54e22b75275e6beb24c4b and ignore having test cases.

MoreLinq/FallbackIfEmpty.cs
atifaziz: I see. Do we want to add the assertions? I guess what bothers me is having to worry about the range and seeing the 4 hard-coded. This is one way to get rid of that:

```c#
IEnumerable<T> FallbackChooser()
{
    return count == null ? fallback : InstancesFallback(count.Value);

    IEnumerable<T> InstancesFallback(int c)
    {
        if (c < 0) throw new ArgumentOutOfRangeException(nameof(count));
        if (c > 0) { c--; yield return fallback1; }
        if (c > 0) { c--; yield return fallback2; }
        if (c > 0) { c--; yield return fallback3; }
        if (c > 0) { c--; yield return fallback4; }
        if (c > 0) throw new ArgumentOutOfRangeException(nameof(count));
    }
}
```

The downside is that the last throw won't happen until after the last yield so yet another way is:

```c#
IEnumerable<T> FallbackChooser()
{
    switch (count)
    {
        case null: return fallback;
        case int n when n == 1: return Fallback1();
        case int n when n == 2: return Fallback2();
        case int n when n == 3: return Fallback3();
        case int n when n == 4: return Fallback4();
        default: throw new ArgumentOutOfRangeException(nameof(count));
    }

    IEnumerable<T> Fallback1() { yield return fallback1; }
    IEnumerable<T> Fallback2() => Fallback1().Concat(fallback2);
    IEnumerable<T> Fallback3() => Fallback2().Concat(fallback3);
    IEnumerable<T> Fallback4() => Fallback3().Concat(fallback4);
}
```

fsateler: I like the last version check. I made the switch is slightly different though as I preserved `InstancesFallback`. What do you think?

fsateler:
> What about reverting ccca164? Then I think we're done. 🏁

Awesome. Just pushed a revert of that commit.

atifaziz:
Did you forget to push the revert commit? I don't see it yet in the PR branch.

fsateler:
> Did you forget to push the revert commit? I don't see it yet in the PR branch.

Oops. The push was rejected due to me not having pulled your commits first. Now its there.

PR #347, opened 2017-09-07
putting - (minus) in front of original comparison does the job

otabekgb:
There is no need to check condition here


fsateler:
Hi, thanks for the PR. Unfortunately, it is incorrect. There is a need to check the condition. This is because `int.MaxValue` is 2147483647 and `int.MinValue` is -2147483648 . Note how they are not the same number. Should the given comparer return `int.MinValue`, just applying `-` would cause an integer overflow, as 2147483648 doesn't fit in an int.

I'm thus closing this PR.

Thanks!

Issue #238, opened 2017-01-14
Left, right and full outer joins


cknightdevelopment:
## Overview

I am thinking that it would be helpful to have `LeftJoin`, `RightJoin`, `InnerJoin`, and `FullOuterJoin` methods in MoreLINQ. I know that there is already `GroupJoin` and `FullGroupJoin` (from MoreLINQ), but I know sometimes I would like a way to have more control over my joins and have them be **flattened**.  

I am thinking these new methods could return an IEnumerable of an object such as this:

```
public class JoinResult<TLeft, TRight>
{
    public TLeft Left { get; set; }
    public TRight Right { get; set; }
}
```

Below is some sample data (to allow this to be a little more concrete) and examples of what I am thinking as far as calling these new methods and the results they would provide:

## Sample Data

```
var people = new List<Person>
{
    new Person { PersonId = 1, Name = "Chris" },
    new Person { PersonId = 2, Name = "Joe" },
    new Person { PersonId = 3, Name = "Mike" }
};
var addresses = new List<Address>
{
    new Address { AddressId = 1, PersonId = 2, Street = "Main St" },
    new Address { AddressId = 2, PersonId = 3, Street = "2nd Ave" },
    new Address { AddressId = 3, PersonId = 4, Street = "Broad St" }
};
```

#### Call

```
people.LeftJoin(
    addresses,
    p => p.PersonId,
    a => a.PersonId
);
```

#### Pseudo-Code Result

```
[
   {Left = Person1, Right = null},
   {Left = Person2, Right = Address1},
   {Left = Person3, Right = Address2}
] 
```

#### Call

```
people.RightJoin(
    addresses,
    p => p.PersonId,
    a => a.PersonId
);
```

#### Pseudo-Code Result

```
[
   {Left = Person2, Right = Address1},
   {Left = Person3, Right = Address2},
   {Left = null, Right = Address3}
] 
```

#### Call

```
people.InnerJoin(
    addresses,
    p => p.PersonId,
    a => a.PersonId
);
```

#### Pseudo-Code Result

```
[
   {Left = Person2, Right = Address1},
   {Left = Person3, Right = Address2}
] 
```

#### Call

```
people.FullOuterJoin(
    addresses,
    p => p.PersonId,
    a => a.PersonId
);
```

#### Pseudo-Code Result

```
[
   {Left = Person1, Right = null},
   {Left = Person2, Right = Address1},
   {Left = Person3, Right = Address2},
[...]

PR #350, opened 2017-09-18
Add left, right and full outer joins

atifaziz:
This PR adds operators for conducting left, right and full outer joins on two sequences, as discussed in #238.

/cc @cknightdevelopment


atifaziz:
@fsateler I had planned [the 2.8.0 release](https://github.com/morelinq/MoreLINQ/milestone/14) for today and I'm guessing you didn't find the time to review so far. I've gone ahead and merged this PR so as not to delay the release. Hopefully there's no glaring omissions but if you find something then let's just deal with it as separate issues.

fsateler:
> I'm guessing you didn't find the time to review so far.

Indeed, I did not. I just took a quick look now, and I wonder why did you not reuse the implementation of `FullGroupJoin`? I think it would have made this PR simpler.

atifaziz:
@fsateler To keep the order of the first sequence, although admittedly, it wasn't a strict requirement.

fsateler:
@atifaziz well, FullGroupJoin is supposed to preserve order of both sequences. We imported the Lookup class in order to guarantee that.

atifaziz:
@fsateler [You're absolutely right and I even used it](https://github.com/morelinq/MoreLINQ/blob/0832e05e46b68cda51c00a3c29f38b1a09e26585/MoreLinq/FullJoin.cs#L234). To be honest, I answered too soon and didn't recall the reason the correctly. I went back and looked at my attempts in local branches, which were cleaned-up and subsequently submitted as this PR.

My initial implementation was, in fact, based your `FullGroupJoin` but it suffered a lot from readability issues. The first attempt used value tuples to express optionality (because first or second element could be missing) and the joined groups. To then support build targets where value tuples are absent, I resorted to `KeyValuePair<,>` but that's where readability degraded considerably. I then stopped being lazy and implemented types to clearly represent optionality and couples. That version looked as follows:

```c#
public static IEnumerable<TResult> FullJoin<TFirst, TSecond, TKey, TResult>(
    this IEnumerable<TFirst> first,
    IEnumerable<TSecond> second,
    Func<TFirst, TKey> firstKeySelector,
    Func<TSecond, TKey> secondKeySelector,
    Func<TFirst, TResult> firstSelector,
    Func<TSecond, TResult> secondSelector,
    Func<TFirst, TSecond, TResult> bothSelector,
    IEqualityComparer<TKey> comparer) =>
    from j in first.Select(Option.Some)
                    .FullGroupJoin(second.Select(Option.Some),
                                    e => firstKeySelector(e.Value),
                                    e => secondKeySelector(e.Value),
                                    (k, fs, ss) => Couple.Create(fs, ss),
                                    comparer)
    from f in j.First.DefaultIfEmpty()
    from s in j.Second.DefaultIfEmpty()
    select f.HasValue && s.HasValue ? bothSelector(f.Value, s.Value)
            : f.HasValue ? firstSelector(f.Value)
            : s.HasValue ? secondSelector(s.Value)
            : throw new Exception("Implementation error.");

static class Option
{
    public static Option<T> Some<T>(T value) => new Option<T>(true, value);
}

struct Option<T> // TODO IEquatable<Option<T1, T2>>
{
    readonly T _value;

    public bool HasValue { get; }

    public T Value =>
        HasValue ? _value : throw new InvalidOperationException();

    public Option(bool hasValue, T value)
    {
        HasValue = hasValue;
        _value = value;
    }

    // TODO public override int GetHashCode()
    // TODO public override bool Equals(object obj)
    // TODO public override string ToString()
}

static class Couple
{
    public static Couple<T1, T2> Create<T1, T2>(T1 first, T2 second) =>
        new Couple<T1, T2>(first, second);
}

struct Couple<T1, T2> // TODO IEquatable<Couple<T1, T2>>
{
    public readonly T1 First;
    public readonly T2 Second;

    public Couple(T1 first, T2 second)
    {
        First = first;
        Second = second;
    }

    // TODO public override int GetHashCode()
    // TODO public override bool Equals(object obj)
    // TODO public override string ToString()
}
```

Besides the many to-do notes that didn't need to be addresses, speaking strictly, I felt that there was a lot of composition taking place (projections into options then a flat-map of the join along with calls into `DefaultIfEmpty`). Once I had the tests in place, I expanded it into a hand-rolled implementation to see if it would be a lot of work and duplication and I felt that the end result was reasonable enough to stand on its own.

atifaziz:
@fsateler Oh, man, I think some of it is coming back to me now and I think the order issue was a desired and intended effect. Bear with me while I try to recall my original thought process from memory. 😕 It's been over two weeks since I opened the PR and moved on to many other things. 

To highlight the difference, suppose the following data:

c#
```
var foo = (1, "foo");
var bar1 = (2, "bar");
var bar2 = (2, "Bar");
var bar3 = (2, "BAR");
var baz = (3, "baz");
var qux = (4, "qux");
var quux = (5, "quux");
var quuz = (6, "quuz");

var xs = new[] { foo, bar1, qux, bar1, foo };
var ys = new[] { quux, bar2, baz, bar3, quuz, quux };
```

Here's a C# Interactive (`csi.exe`) session showing that while both `FullJoin` and `FullGroupJoin` produce similar output, the former respects the order of the first and second sequence while the latter respects the order _within each group_:

```
Microsoft (R) Visual C# Interactive Compiler version 2.3.2.62019

> #r "MoreLinq.dll"
> #r "System.ValueTuple.dll"
> using MoreLinq;
>
> var foo = (1, "foo");
> var bar1 = (2, "bar");
> var bar2 = (2, "Bar");
> var bar3 = (2, "BAR");
> var baz = (3, "baz");
> var qux = (4, "qux");
> var quux = (5, "quux");
> var quuz = (6, "quuz");
>
> var xs = new[] { foo, bar1, qux, bar1, foo };
> var ys = new[] { quux, bar2, baz, bar3, quuz, quux };
>
> var nil = (default(int), default(string));
>
> var js =
.     xs.FullJoin(ys,
.         e => e.Item1,
.         f => new { First = f, Second = nil },
.         s => new { First = nil, Second = s },
.         (f, s) => new { First = f, Second = s });
>
> foreach (var e in js)
.     Console.WriteLine(e.ToString());
{ First = (1, foo), Second = (0, ) }
{ First = (2, bar), Second = (2, Bar) }
{ First = (2, bar), Second = (2, BAR) }
{ First = (4, qux), Second = (0, ) }
{ First = (2, bar), Second = (2, Bar) }
{ First = (2, bar), Second = (2, BAR) }
{ First = (1, foo), Second = (0, ) }
{ First = (0, ), Second = (5, quux) }
{ First = (0, ), Second = (3, baz) }
{ First = (0, ), Second = (6, quuz) }
{ First = (0, ), Second = (5, quux) }
>
> var jgs =
.     from j in xs.Select(e => (Defined: true, Item: e))
.                 .FullGroupJoin(ys.Select(e => (Defined: true, Item: e)),
.                                f => f.Item.Item1,
.                                s => s.Item.Item1)
.     from f in j.First.DefaultIfEmpty()
.     from s in j.Second.DefaultIfEmpty()
.     select f.Defined && s.Defined ? new { First = f.Item, Second = s.Item }
.          : f.Defined ? new { First = f.Item, Second = nil }
.          : new { First = nil, Second = s.Item };
>
> foreach (var e in jgs)
.     Console.WriteLine(e.ToString());
{ First = (1, foo), Second = (0, ) }
{ First = (1, foo), Second = (0, ) }
{ First = (2, bar), Second = (2, Bar) }
{ First = (2, bar), Second = (2, BAR) }
{ First = (2, bar), Second = (2, Bar) }
{ First = (2, bar), Second = (2, BAR) }
{ First = (4, qux), Second = (0, ) }
{ First = (0, ), Second = (5, quux) }
{ First = (0, ), Second = (5, quux) }
{ First = (0, ), Second = (3, baz) }
{ First = (0, ), Second = (6, quuz) }
```

Note also how `foo`, which appears in `xs` at the start and end, appears grouped together in the output from `FullGroupJoin`. `FullJoin`, on the other hand, preserves that. Likewise for wherever `bar1` appears in `xs`.

fsateler:
> Oh, man, I think some of it is coming back to me now and I think the order issue was a desired and intended effect. Bear with me while I try to recall my original thought process from memory. 😕 It's been over two weeks since I opened the PR and moved on to many other things.

Of course, it's the problem that happens when reviewers are not responsive 😞 . 

> To highlight the difference, suppose the following data
> <snip detailed example>

Oh, indeed! But it seems to me you may have uncovered a (design) bug in `FullGroupJoin`. I now tested that the upstream `GroupJoin` behaves as your current example: it may yield the same key multiple times, whereas `FullGroupJoin` creates a single group per key. Unfortunately we have documented the behavior, and thus tied our hands 😞 . Oh well. 

Anyway, thanks for the thorough explanation!

atifaziz:
> Of course, it's the problem that happens when reviewers are not responsive 😞 . 

I hope you didn't take that in wrong way. I get unresponsive all the time. A lot of this is volunteered work so completely understandable.

fsateler:
>>Of course, it's the problem that happens when reviewers are not responsive :disappointed: .
>
> I hope you didn't take that in wrong way. I get unresponsive all the time. A lot of this is volunteered work so completely understandable.

I did not. I was just mentioning that is also fully understandable if you forget some details if there's no response in two weeks.
Cheers!

Issue #351, opened 2017-09-19
Conflicts with new Enumerable members in .NET Core .20


atifaziz:
It looks like that .NET Core 2.0 has introduced a number of new methods on [`Enumerable`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable?view=netcore-2.0) that could potentially conflict with those in MoreLINQ. The problem is that the unit tests may be testing the LINQ extension methods instead of those from MoreLINQ and could either be giving false positives (passing when they should be failing) or false negatives (failing when they should be passing). This happened with, for example, 5eb30a2d70fbfaa294f2c23cb07085f9cd670441 in #304 where a fix still had [a `TaskLast` test failing mysteriously](https://github.com/morelinq/MoreLINQ/pull/304#pullrequestreview-61429113).

The following methods are suspects and need to be investigated:

- ~[`Concat`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.concat?view=netcore-2.0)~
- ~[`OrderBy`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.orderby?view=netcore-2.0)~
- [x] [`Prepend`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.prepend?view=netcore-2.0)
- ~[`Repeat`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.repeat?view=netcore-2.0)~
- [x] [`SkipLast`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.skiplast?view=netcore-2.0)
- [x] [`TakeLast`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.takelast?view=netcore-2.0)
- ~[`ThenBy`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.thenby?view=netcore-2.0)~
- ~[`ToDictionary`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.todictionary?view=netcore-2.0)~
- [x] [`ToHashSet`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tohashset?view=netcore-2.0)
- ~[`ToLookup`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tolookup?view=netcore-2.0)~

atifaziz:
[...]

PR #352, opened 2017-09-21
Prevent accidental testing of duplicate LINQ members 

atifaziz:
This PR addresses #351.

The strategy taken in the initial commit, 8871edb14fd7f5f1148fbde8c81b819c277c8fe8, is to have an `Enumerable` type act as a proxy for `System.Linq.Enumerable` and forwards all calls. This allows all LINQ extension methods to be used without importing `System.Linq`, which is now only done in one place (`Enumerable.cs` in the test project). It also enables a controlled way to manage conflicts by not having any of the duplicate members defined on the `Enumerable` proxy. For example, since it doesn't define and wrap [`TakeLast`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.takelast?view=netcore-2.0) that was introduced in .NET Core 2.0, we know for that the unit tests for MoreLINQ's `TakeLast` won't accidentally test the LINQ implementation (which happened in #304).

This is not a fool-proof approach because it requires being careful that no one imports `System.Linq` in a file with a test fixture. This is something that would have to be caught during PR reviews. A grep should also be quick to conduct and shouldn't reveal any other file than `MoreLinq.Test/Enumerable.cs`:

    $ egrep '^ *using +System\.Linq;' MoreLinq.Test/*.cs

An alternative approach would be to move the `System.Linq` imports in each unit test file outside the `MoreLinq.Test` namespace declaration, as was done in 88967853e005a9f20100948a70525bb33b3bc289 for #304. It's a little less involved but requires equal care during reviews.

### DIFF
--- a/MoreLinq.Test/Enumerable.cs
+++ b/MoreLinq.Test/Enumerable.cs
@@ -1,0 +1,538 @@
+namespace MoreLinq.Test
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using LinqEnumerable = System.Linq.Enumerable;
+
+    [DebuggerStepThrough]
+    static partial class Enumerable
+    {
+        public static TSource Aggregate<TSource>(this IEnumerable<TSource> source, Func<TSource, TSource, TSource> func) =>
+            LinqEnumerable.Aggregate(source, func);
+
+        public static TAccumulate Aggregate<TSource, TAccumulate>(this IEnumerable<TSource> source, TAccumulate seed, Func<TAccumulate, TSource, TAccumulate> func) =>
+            LinqEnumerable.Aggregate(source, seed, func);
+
+        public static TResult Aggregate<TSource, TAccumulate, TResult>(this IEnumerable<TSource> source, TAccumulate seed, Func<TAccumulate, TSource, TAccumulate> func, Func<TAccumulate, TResult> resultSelector) =>
+            LinqEnumerable.Aggregate(source, seed, func, resultSelector);
+
+        public static bool All<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate) =>
+            LinqEnumerable.All(source, predicate);
+
+        public static bool Any<TSource>(this IEnumerable<TSource> source) =>
+            LinqEnumerable.Any(source);
+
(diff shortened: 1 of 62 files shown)


MoreLinq.Test/Enumerable.cs
```diff
@@ -0,0 +1,538 @@
...
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using LinqEnumerable = System.Linq.Enumerable;
+
+    [DebuggerStepThrough]
+    static partial class Enumerable
```
fsateler: Do we really need all these methods? Or did you just put all non-conflicting methods here?

atifaziz: How do you define those as _we actually need_? Are all the methods in there used? No, so no they're not needed _today_? The idea, however, was to save trouble for a future contributor.

fsateler: It's not really a big deal, but I don't really like dead/unused code. It's not hard to add the forward either so they can be added as needed. I don't feel strongly about this so lets keep all forwards if you prefer that.

fsateler:
> An alternative approach would be to move the System.Linq imports in each unit test file outside the MoreLinq.Test namespace declaration

I prefer the current approach over this one, because import preference, while being well defined, is a rarely used[1] feature, so its importance is easy to forget/overlook. The current approach is easier to hook into the buildsystem too.

[1] At least by me. YMMV of course.

atifaziz:
> I would like the test you suggested be integrated into the CI runs

@fsateler Good idea & done. We good to merge & close this?

fsateler:
> Too bad a download is needed in order to user grep, but I guess that is not avoidable ATM.

Sorry, I just looked this up. It appears [msys is preinstalled](https://www.appveyor.com/docs/build-environment/#mingw-msys-cygwin) . Can't we use that?

atifaziz:
> Too bad a download is needed in order to user `grep`, but I guess that is not avoidable ATM.

It's not really needed because I could have using [`-match` from PowerShell](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-5.1) but in an initial cut I just wanted to get it working while maintaining parity between the Windows and non-Windows versions. We can improve with time.

Issue #247, opened 2017-02-06
Document project coding style and other guidelines


fsateler:
To make it easier for contributors to have the correct style, we should document it. This should hopefully save time for everyone involved.

Formatting could additionally made easier by using an [editorconfig](http://editorconfig.org/) file.

fsateler:
I've been toying with an editorconfig file[1] . A problem that I did not manage to solve with it is the indentation of `case` statements: there does not appear to be a editorconfig setting for that.


[1] Current file is:
```ini
root = true

[*]
end_of_line = crlf
insert_final_newline = true
trim_trailing_whitespace = true

[*.xml]
indent_style = space

[*.cs]
indent_style = space
indent_size = 4
csharp_new_line_before_open_brace = properties,control_blocks,types,methods,anonymous_types,object_collection_array_initalizers
csharp_new_line_before_else = true
csharp_new_line_before_catch = true
csharp_new_line_before_finally = true
csharp_new_line_before_members_in_object_initializers = true
csharp_new_line_before_members_in_anonymous_types = true
```

fsateler:
Another problem is constructs like this:

```csharp
public Constructor()
    this(null) {}
````

Editorconfig pushes out the `{}` to a separate line.

And another is constructs like this:
```csharp
return _(); IEnumerable<T> _() 
    {
        yield something;
    }
```

Since the `_()` method declaration is a separate statement, the indentation rule removes one level of indentation.

PR #353, opened 2017-09-26
Add editorconfig file

fsateler:
The first commit is the interesting one. The second one just applies the results willi-nilly to see how much difference do we get. It should be removed before merging.

Some cases cannot be expressed in the (simple) editorconfig language: table-like forms are lost. The extra indentation added to inner methods is not possible to express either AFAICS: to the editorconfig service, it is a new statement and thus should be indented at the method level.

The rest of the differences look very reasonable to me.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,11 +11,3 @@ indent_style = space
 [*.cs]
 indent_style = space
 indent_size = 4
-csharp_new_line_before_open_brace = properties,control_blocks,types,methods,anonymous_types,object_collection_array_initalizers
-csharp_new_line_before_else = true
-csharp_new_line_before_catch = true
-csharp_new_line_before_finally = true
-csharp_new_line_before_members_in_object_initializers = true
-csharp_new_line_before_members_in_anonymous_types = true
-csharp_indent_switch_labels = false
-csharp_preserve_single_line_blocks = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+root = true
+
+[*]
+end_of_line = crlf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.xml]
+indent_style = space
+
+[*.cs]
+indent_style = space
+indent_size = 4
+csharp_new_line_before_open_brace = properties,control_blocks,types,methods,anonymous_types,object_collection_array_initalizers
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_indent_switch_labels = false
+csharp_preserve_single_line_blocks = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,6 @@
 root = true
 
 [*]
-end_of_line = crlf
 insert_final_newline = true
 trim_trailing_whitespace = true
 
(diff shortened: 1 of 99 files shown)


MoreLinq.sln
```diff
@@ -1,9 +1,10 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.4
+VisualStudioVersion = 15.0.26403.7
```
atifaziz: This change is unrelated.

.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.xml]
+indent_style = space
+
+[*.cs]
+indent_style = space
+indent_size = 4
+csharp_new_line_before_open_brace = properties,control_blocks,types,methods,anonymous_types,object_collection_array_initalizers
```
atifaziz: I would drop everything `charp_*` as this is subjective and should be discussed and submitted as a separate PR.

.editorconfig
```diff
@@ -0,0 +1,21 @@
+root = true
+
+[*]
+end_of_line = crlf
```
atifaziz: This shouldn't be necessary with new-line normalization in `.gitattributes`. Also new-line terminators vary with platforms.

fsateler:
> Please revert e575e00 as it has destroyed a lot hand-formatting of code for readability. The problem with global application of styles is exactly this

Agreed. Not all styling can be applied automatically. The point of that commit is merely to show what sort of formatting would be automatically applied. As I said in the PR message, that commit should not be merged.

> I would drop everything `charp_*` as this is subjective and should be discussed and submitted as a separate PR.

Well, my primary goal was to reduce the number of style corrections (such as [this one](https://github.com/morelinq/MoreLINQ/pull/331#discussion_r125746564)). As such, I tried to preserve the current conventions as much as it is possible to express in the relatively simple set of editorconfig knobs. In other words, this is a simple first step towards fixing #247. If I missed a knob, or set one incorrectly, I'd like to see that fixed too.

MoreLinq.sln
fsateler: This was automatically inserted. Unfortunately VS encodes the version in the sln file, so at some point this number should be bumped? Why not now?

atifaziz:
> Well, my primary goal was to reduce the number of style corrections (such as this one). As such, I tried to preserve the current conventions as much as it is possible to express in the relatively simple set of editorconfig knobs. In other words, this is a simple first step towards fixing #247. If I missed a knob, or set one incorrectly, I'd like to see that fixed too.

Agreed for going forward provided we revert e575e00a277aadf6da97f687601b01a3b105535a on this PR so change net out on a squash-merge

MoreLinq.sln
atifaziz: I'm not going to pick on it but changes in a diff should be related to the commit message and anything else is a distraction. Something like this can be avoided with a `git add -p`.

atifaziz:
The PR branch now has a conflict with master that will disappear if e575e00a277aadf6da97f687601b01a3b105535a is reverted.

fsateler:
I have reverted the `csharp_*` directives, and added a few more changes on top (each with explanatory commit). PTAL.

fsateler:
> I don't think 0296bfa had the intended effect. It reverted the opposite change.

Doh! 🤦‍♂️  I got confused with the revert and `reset -p`

Issue #356, opened 2017-10-07
Flatten function


leandromoh:
I think would be useful a function that flatten nestead sequences. Not only `IEnumerable<IEnumerable<T>>`, or `IEnumerable<IEnumerable<IEnumerable<T>>>` but in any level of depth. 

So I propose the following function:

```
static IEnumerable<object> Flatten(this IEnumerable source)
{
	foreach (var e in source)
	{
		if (e is IEnumerable seq)
		{
			foreach (var i in Flatten(seq))
			{
				yield return i;
			}
		}
		else
		{
			yield return e;
		}
	}
}
```

So, we can flatten sequences in any deepth just with `Flatten(source).Cast<T>()`.
It also works with arbitrarily-nested sequence of different types. In the example below, 

`list.Flatten().OfType<int>() `returns `[1, 2, 3, 4, 5, 6]` and
`list.Flatten().OfType<string>()` returns `["foo", "bar"]`
`list.Flatten().OfType<bool>()` returns `[true]`
```
var list = new List<object>()
{
    1,
    2,
    new List<object>()
    {
        3,
        new ArrayList()
        {
            4,
            "foo"
        },
        5,
        true,
    },
   "bar",
    6
};
```

I would like to submit a PR for this.
What do you think?

atifaziz:
> I would like to submit a PR for this.

Go ahead & thanks!

atifaziz:
This is now available in release [2.10.0](https://github.com/morelinq/MoreLINQ/releases/tag/v2.10.0).

PR #357, opened 2017-10-12
Flatten  (Issue #356) 

leandromoh:
Implementation for #356.


MoreLinq.Test/NullArgumentTest.cs
```diff
@@ -142,7 +142,11 @@ static object CreateInstance(Type type)
             if (type.GetTypeInfo().IsValueType || HasDefaultConstructor(type)) return Activator.CreateInstance(type);
             if (typeof(Delegate).IsAssignableFrom(type)) return CreateDelegateInstance(type);
 
-            return CreateGenericInterfaceInstance(type.GetTypeInfo());
+            var typeInfo = type.GetTypeInfo();
+
+            return typeInfo.IsGenericType
+                    ? CreateGenericInterfaceInstance(typeInfo)
+                    : CreateNonGenericInterfaceInstance(typeInfo);
```
atifaziz: I don't understand the purpose of `CreateNonGenericInterfaceInstance`. Why not just use `new NonGenericArgs.Enumerable()`?

MoreLinq.Test/FlattenTest.cs
```diff
@@ -0,0 +1,196 @@
...
+                    9,
+                    10
+                },
+            };
+
+            var result = source.Flatten();
+
+            Assert.AreEqual(13, result.Count());
+
+            result.OfType<int>().AssertSequenceEqual(Enumerable.Range(1, 10));
```
atifaziz: These assertions are not very clear and seem like a shortcut. They don't assert the order in which elements are returned in `result`. It's much clearer to write the full expected result.

MoreLinq.Test/NullArgumentTest.cs
leandromoh: I tried to follow the pattern, but really it is not necessary here

MoreLinq/Flatten.cs
```diff
@@ -0,0 +1,110 @@
...
+                            }
+                            else
+                            {
+                                yield return e.Current;
+                            }
+                        }
+
+                        if (!next)
+                        {
+                            (e as IDisposable)?.Dispose();
```
atifaziz: Test to exercise this disposal is missing.

MoreLinq/Flatten.cs
```diff
@@ -0,0 +1,110 @@
...
+                        }
+                    }
+                }
+                finally
+                {
+                    stack.Prepend(e)
+                         .OfType<IDisposable>()
+                         .ForEach(x => x.Dispose());
```
atifaziz: Test to exercise these disposals in the event of an exception is missing.

leandromoh:
> I made some improvements directly to speed up things. There are just some tests for disposals missing now and then I believe we'll be done.

Thanks for the help! Surely it speed up.

Tests for dispose were delivery.

leandromoh:
@atifaziz I finished the development. I added the overload for map an object to a property. IMHO It is only missing the parameter documentation for "selector"

leandromoh:
@atifaziz sorry, I thought I had undertood your idea but I just did it now. I dont like too much about the idea of return `null `to indicate `predicate `(as you said, it is not an obvious signature). 

How about maintain `predicate` and `selector `separated but put `selector `as `Func <IEnumerable, bool>`? If we can not reach consensus on the final design soon, I suggest merge the PR with the MVP (`predicate `only) and think discuss about `selector` in a separate issue.

leandromoh:
@atifaziz what about the above question?

atifaziz:
> I suggest merge the PR with the MVP (`predicate` only) and discuss about `selector` in a separate thread (new issue).

Let's shoot for this. Doing an MVP is a good thing anyway.

leandromoh:
@atifaziz The PR is only with the MVP now. ready for merge. can we publish a new release before december finished?

leandromoh:
Any news?

atifaziz:
Did you rebase and remove commits because that's confused me a bit and we may have lost some work. The history of this PR seems to have diverged at d0297a5c4b19e73a1afe2aa46942523bd40b4b5d. I think it would have been simpler to `git revert` if you wanted to back out of changes. It also doesn't help if commit messages read “improvements” as it makes the review slower because then I need to go through the physical diff changes and figure out what they're about. All this means that a review that should have taken 5 minutes took 15 and then I had to abort and get back to other work until I have more time to spare.

leandromoh:
Did you rebase and remove commits 

I checkout to the last commit without selector overload. But before that, created a new branch with all work (this one for the possíble new PR)

Sorry about the commit messages, i will be more descritive next time.

atifaziz:
> I checkout to the last commit without selector overload. But before that, created a new branch with all work (this one for the possíble new PR)

Okay, that explains it and so the branch was rewound. As I said, a revert would have been the right call here.

> Sorry about the commit messages, i will be more descritive next time.

Cheers! :+1:

atifaziz:
@leandromoh I've put your changes on top of a revert (c6f59f6cf44c4fd6fa02899e31bb500540747191) just to maintain the “selector overload” as part of this PR's history and discussion, in case we need to pick it up again.

leandromoh:
Ok. Thanks.

What about the code itself? Are we ready to merge?

MoreLinq/Flatten.cs
```diff
@@ -0,0 +1,110 @@
...
+
+                        if (!next)
+                        {
+                            (e as IDisposable)?.Dispose();
+                        }
+                    }
+                }
+                finally
+                {
+                    stack.Prepend(e)
```
atifaziz: The `finally` block should be simple rather than use any fancy LINQ to avoid causing allocations under low memory conditions (the reason for which the `finally` code could be running in the first place). You can re-write this in the same lines of code:

```c#
(e as IDisposable)?.Dispose();
foreach (var se in stack)
    (se as IDisposable)?.Dispose();
```

Note that since [`Stack<>`'s enumerator](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.stack-1.enumerator?view=netcore-2.0) is a `struct`, there's no cost to iterating over it using `foreach`.

leandromoh:
@atifaziz code refactored! also I added flatten to readme and release notes.

PR #360, opened 2017-11-07
Add line break to improve readability

BKSpurgeon:


c9952594:
I think this is a repeated pattern throughout the tests.

Would it be worth updating the other tests as well to maintain consistency across the code base?

atifaziz:
Thanks, but this is completely intentional and a style shared throughout the code base (see #290) as also noted by @c9952594.

Putting the return statement and the local function declaration on a single line is designed to compensate for the fact that the language doesn't support _anonymous_ iterators succinctly. There is no extra information or context being provided on that line so there's no clarity gained by separating the two except that it might appear a little unorthodox in styling. Just consider how many things are immaterial:

- The name of the local iterator function so it is given the bogus name of `_`.
- The return type of the local function because it's redundant with the outer method's return type.
- The call to the local function never really executes because the iterator function becomes lazy so it's even somewhat misleading to highlight the call on its own.

What's in fact being returned is an iterator object with the body of its algorithm and so the style of putting it all on a single line is designed to make it appear just as that.

The origin and play on the styling come the idea that… 

> If you squint hard enough, you can almost believe that C# 7 now has [anonymous
iterators][why-no-anonymous-iterators]…
>
> &mdash; <cite>[Anonymous Iterators in C# 7, Almost][anonymous-iterators-almost]</cite>

See also some examples in #291.

[anonymous-iterators-almost]: https://gist.github.com/atifaziz/ce69bbb3dd2b64ec5fbe9c224d218198
[why-no-anonymous-iterators]: https://blogs.msdn.microsoft.com/ericlippert/2009/08/24/iterator-blocks-part-seven-why-no-anonymous-iterators/

Issue #328, opened 2017-07-01
Max/MinBy should return all matching elements


orthoxerox:
The docs say that "If more than one element has the maximal projected value, the first one encountered will be returned." Is there a method that returns all of them?

If not, could it be added? If yes, I think there should be a reference to it in the docs.

atifaziz:
I personally think that `MinBy` and `MaxBy` should themselves be changed to return a list. Since that would be a breaking change, it would have to be scheduled for [the next major release](https://github.com/morelinq/MoreLINQ/milestone/2).

orthoxerox:
Why not add `SingleMaxBy` and `EveryMaxBy` in the next minor release and obsolete `MaxBy` in the next major?

atifaziz:
@orthoxerox That would be a quick and simple fix but which would have be reverted anyhow when we could afford a breaking change. Perhaps it's getting time for a [major release that's been pending](https://github.com/morelinq/MoreLINQ/milestone/2).

TormodS:
Why......? Why would you do this? 
If you have a collection of integers, let's say [1,2,3,3] and you invoke "Max" on it. Would you return 3 or a collection [3,3]?
I assume you would return only 3. Why on earth would you return a collection from MaxBy then?
As vanilla LINQ in .NET 7 now includes MaxBy, then you see that it is properly implemented there.

We had a legacy library that was using old MoreLinq nuget. When ported to .NET standard, the dev didn't pay attention when bumping to new MoreLinq version and just added a ToString() when using MaxBy in an edge case of course resulting in a bug that took a while to discover.

What interpretation of MaxBy() would warrant breaking the symmetry with Max(), which is the case for all the "-by" operators such as DisctinctBy, OrderBy, IntersectBy, UnionBy etc?

viceroypenguin:
@TormodS If you update to v4.0.0, you'll find that the method has been renamed to `Minima`/`Maxima`, avoiding this discrepancy entirely.

TormodS:
@viceroypenguin 
Right, I see that the MaxBy and MinBy have been deprecated (not reverted to their proper implementation) in version 4.0. 
I suppose that this means that I can use Minima().First() construct as a future proof workaround that wont break if (when?) the proper implementation of MaxBy and MinBy returns.

I agree that this change avoids the convention discrepancy, and I hope that mainstream functions such as MaxBy and MinBy will return in the next major version. 

[...]

PR #369, opened 2018-01-25
Change MaxBy/MinBy to return all extrema elements (#328)

atifaziz:
This PR addresses #328.


fsateler:
Overall looks good to me. I have just one doubt: Now that this metod returns multiple elements, it could be deferred? Or in other words, why make execution immediate?

atifaziz:
@fsateler That's a fair point. I'll revise the methods to use deferred execution semantics.

MoreLinq/MaxBy.cs
```diff
@@ -81,29 +81,74 @@ static partial class MoreEnumerable
...
+                    while (e.MoveNext())
                     {
-                        extremum = candidate;
-                        key = candidateProjected;
+                        var item = e.Current;
+                        var key = selector(item);
+                        var comparison = comparer(key, extremaKey);
+                        if (comparison > 0)
+                        {
+                            extrema = new List<TSource> { item };
```
fsateler: Is there a reason for not using `extrema.Clear()` + `extrema.Add(item)` ? Using this could save some allocations and GC pressure, at the cost of potentially keeping more memory allocated than strictly necessary (ie, if there are many equal keys that appear before the true maximal key).

atifaziz: > at the cost of potentially keeping more memory allocated than strictly necessary

I think you answered the question yourself.

The cost is going to be different for everyone depending on what's going to be their most common case. If you've collected the wrong list of extrema and going to restart it then `Clear` is going to spend time zero-ing out the internal array to release the elements. You then also risk, under memory pressure conditions, promoting the list and the internal array to the next generation. If you use `TrimExcess` then you still end up with an allocation and so you're only saving the trivial cost of a `List<>` allocation. The simplest thing to do therefore is to let the system do the balancing act and start afresh with a new list each time it needs to be restarted. The clever thing to do would be to build an adaptive system, but I don't think we want to get into that level of complexity (YAGNI?). Even `OrderBy` is hard-wired to one sorting algorithm.

atifaziz:
> However, if the purpose of returning this lazy list is to allow downstream operators to optimize, why not return `IList<T>` directly?

How do you propose to do that _lazily_? Perhaps I'm not seeing something you are. An `IList<>` is what was being returned before it was made lazy. Are you suggesting to change the return type? The optimisation usually rely on the run-time type so they'll kick-in nonetheless.

leandromoh:
Why simple not follow the same approach as [CountBy ](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/CountBy.cs#L74) and [ScanRight](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/ScanRight.cs#L109):

atifaziz:
@leandromoh Good suggestion and addressed in f6586bcc76409d2bc898a8c7c95b018499c9b9f0.

leandromoh:
LGTM. When do you plan to merge?

Issue #128, opened 2016-01-14
Optimize collection/list-optimized methods for read-only family


Arithmomaniac:
[Edit by @atifaziz] This issue tracks the optimizations as applicable to the following methods:

- [x] [`IReadOnlyCollection<>`][IReadOnlyCollection`1]
  - [x] `AssertCount`
  - [x] Counting methods:
    - [x] `AtLeast`
    - [x] `AtMost`
    - [x] `Exactly`
    - [x] `CountBetween`
    - [x] `CompareCount`
  - [x] `EndsWith`
  - [x] `FallbackIfEmpty`
  - [x] `PadStart`
  - [x] `SkipLast`
  - [x] `TakeLast`
  - [x] `CountDown`
- [x] [`IReadOnlyList<>`][IReadOnlyList`1]
  - [x] `AggregateRight`
  - [x] ~~`SingleOrFallback`~~ (#388)
  - [x] `Slice`
  - [x] `ScanRight`
---
[Original issue comment as opened by @Arithmomaniac]

The current `AtLeast` has an optimization for `ICollection`. It should have one for `IReadOnlyCollection` as well.

[IReadOnlyCollection`1]: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlycollection-1?view=netstandard-1.0
[IReadOnlyList`1]: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlylist-1?view=netstandard-1.0

sholland1:
The `IReadOnlyCollection` interface was introduced in .NET 4.5.

I upgraded the test projects, so I could confirm that the current implementation isn't optimized for `IReadOnlyCollection`s. It isn't, so we would have to upgrade the projects to at least .NET 4.5 to implement this.

@atifaziz, what do you want to do?

atifaziz:
@sholland1 I think we should move to .NET 4.5 but post 2.0 release, which I hope to do any day (though unfortunately due to time constraints, that's been getting postponed a lot). It would be good to track the upgrade to 4.5 as a separate issue.

Arithmomaniac:
Now that MoreLinq is multi-targeted, it may be more compelling to add a new Framework45 target. It can certainly be added to the NetStandard targets; it's been [supported since 1.0](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlycollection-1?view=netframework-4.7.1#Applies_to).

[...]

PR #372, opened 2018-01-31
Optimise EndsWith for read-only collections

Arithmomaniac:
Another item in #128, with ❤️ from @Milyli.


MoreLinq.Test/ReadOnlyDecoratedList.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    /// <summary>
+    /// A decorator for a <see cref="IReadOnlyList{T}"/> that ensures it does not implement another interface.
+    /// </summary>
+    sealed class ReadOnlyDecoratedList<T> : IReadOnlyList<T>
```
atifaziz: If you're being strict then why not just have it implement `IReadOnlyCollection<>`?

MoreLinq.Test/ReadOnlyDecoratedList.cs
```diff
@@ -0,0 +1,25 @@
+namespace MoreLinq.Test
+{
+    using System;
```
atifaziz: This import is not used.

MoreLinq.Test/ReadOnlyDecoratedList.cs
```diff
@@ -0,0 +1,25 @@
...
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Linq;
```
atifaziz: This import is not used.

MoreLinq.Test/ReadOnlyDecoratedList.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    /// <summary>
+    /// A decorator for a <see cref="IReadOnlyList{T}"/> that ensures it does not implement another interface.
+    /// </summary>
+    sealed class ReadOnlyDecoratedList<T> : IReadOnlyList<T>
+    {
+        private readonly IReadOnlyList<T> _list;
```
atifaziz: `private` is redundant.

MoreLinq.Test/MoreLinq.Test.csproj
```diff
@@ -37,6 +37,7 @@
 
   <ItemGroup>
       <Compile Include="*Test.cs" />
+      <None Remove="ReadOnlyDecoratedList.cs" />
```
atifaziz: Is this needed?

MoreLinq.Test/EndsWithTest.cs
```diff
@@ -91,5 +91,15 @@ public void EndsWithUsesSpecifiedEqualityComparerOrDefault()
             Assert.False(first.EndsWith(second, EqualityComparer.Create<int>(delegate { return false; })));
             Assert.True(first.EndsWith(second, EqualityComparer.Create<int>(delegate { return true; })));
         }
+
+#if IREADONLY
+        [TestCase(new[] { 1, 2, 3 }, new[] { 2, 3 }, ExpectedResult = true)]
+        [TestCase(new[] { 1, 2, 3 }, new[] { 1, 2, 3 }, ExpectedResult = true)]
+        [TestCase(new[] { 1, 2, 3 }, new[] { 0, 1, 2, 3 }, ExpectedResult = false)]
+        public bool EndsWithWithReadOnlyIntegers(IReadOnlyList<int> first, IReadOnlyList<int> second)
```
atifaziz: Is this test asserting anything? If I comment out the optimisation for `IReadOnlyCollection<>`, it still passes and which doesn't seem right.

Arithmomaniac: It makes sure that the optimization for `IReadOnlyCollection<>` does not cause it to fail. Is that worth something?

MoreLinq.Test/MoreLinq.Test.csproj
Arithmomaniac: That's strange; it was supposed to be in the "if .NET Framework" section. I don't even get why it compiles. I'll take a look at that.

MoreLinq.Test/ReadOnlyDecoratedList.cs
Arithmomaniac: I can update the text. The meaning is that it doesn't implement `ICollection<>` or `IList<>`. so that any tests will use the `IReadOnlyCollection<>` or `IReadOnlyList<>` optimization.

MoreLinq.Test/EndsWithTest.cs
atifaziz: Yes but the first would be that it is optimised for `IReadOnlyCollection<>` in the first place. How do we know that's covered. As I [demonstrated](https://github.com/morelinq/MoreLINQ/pull/372#discussion_r165290418), someone could come along and comment out that bit or refactor the optimisation out accidentally and tests would let that slip.

MoreLinq.Test/ReadOnlyDecoratedList.cs
atifaziz: >  or `IReadOnlyList<>` optimization

Let's cross that bridge when we get there but not prepare in anticipation of a future that may never occur. Strictness is better here.

Arithmomaniac:
It seems all the problems were in the testing code, and since the tests didn't include a check on the `ICollection` optimization (though the arrays are `ICollection`s), I figured it's OK to not check for `IReadOnlyConnection` either.

MoreLinq/EndsWith.cs
```diff
@@ -71,11 +71,31 @@ public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second,
 
             comparer = comparer ?? EqualityComparer<T>.Default;
 
-            var secondCollection = second as ICollection<T> ?? second.ToList();
-            using (var firstIter = first.TakeLast(secondCollection.Count).GetEnumerator())
+            var secondCollection = second;
```
atifaziz: I find that this entire section up to the `using` can be improved and folded into a simple expression, using either a tuple:

```c#
            List<T> secondList;
            var (secondCollection, secondCount)
                = second is ICollection<T> collection ? (second, collection.Count)
#if IREADONLY
                : second is IReadOnlyCollection<T> readOnlyCollection ? (second, readOnlyCollection.Count)
#endif
                : (secondList = second.ToList(), secondList.Count);
```

or delegating to a local function:

```c#
            List<T> secondList;
            return second is ICollection<T> collection ? Impl(second, collection.Count)
#if IREADONLY
                 : second is IReadOnlyCollection<T> readOnlyCollection ? Impl(second, readOnlyCollection.Count)
#endif
                 : Impl(secondList = second.ToList(), secondList.Count);

            bool Impl(IEnumerable<T> snd, int count)
            {
                using (var firstIter = first.TakeLast(count).GetEnumerator())
                    return snd.All(item => firstIter.MoveNext() && comparer.Equals(firstIter.Current, item));
            }
```

The flow with both is much easier to follow and reason about.

I would have liked to say take your pick but the first one is a no-go for now due to a regression in Roslyn that unnecessarily forces the use of C# 7.1 (see dotnet/roslyn#21518).

Arithmomaniac:
Hopefully I'll soon learn these conventions so that you don't have to rewrite every commit for me 😉

Issue #229, opened 2017-01-04
Remove zero-arg ToDelimitedString overload


atifaziz:
The utility of `ToDelimitedString` overloads accepting zero arguments and using the list separator from the current culture is of very marginal utility and should be considered as a strong candidate for removal in the next major version/release as a breaking change. If so, then the [next release](https://github.com/morelinq/MoreLINQ/milestone/1) should mark those methods as being obsolete.

Should `null` then be allowed as a delimiter to keep consistency with [`String.Join`](https://msdn.microsoft.com/en-us/library/dd992421(v=vs.110).aspx) that does, and treats `null` effectively as an empty string?

This issue has been forked from #116.

fsateler:
> The utility of `ToDelimitedString` overloads accepting zero arguments and using the list separator from the current culture is of very marginal utility and should be considered as a strong candidate for removal in the next major version/release as a breaking change.

Agreed.

> If so, then the next release should mark those methods as being obsolete.

Also agreed

> Should null them be allowed as a delimiter to keep consistency with String.Join that does and treats as effectively an empty string?

I don't think so. If you want empty string, pass an empty string. Moreover, making a parameter nullable is a non-breaking change (and thus doable in a 3.x release), while making a parameter non-nullable is a breaking change (and thus requiring a 4.0 release).

atifaziz:
> If you want empty string, pass an empty string.

Fair enough.

> making a parameter nullable is a non-breaking change (and thus doable in a 3.x release), while making a parameter non-nullable is a breaking change (and thus requiring a 4.0 release).

I tried re-reading this a couple of time but can't make sense of it. How did you get from 3.x to 4.0?

atifaziz:
Right so if the zero-arg `ToDelimitedString` overload is dropped then that's a breaking change and is currently slated for [the next major version](https://github.com/morelinq/MoreLINQ/milestone/2). At the same time, one has to review what allowing `null` for the delimiter means. Changing its meaning to be the same as an empty string, to match `String.Join`, or disallowing it completely can be made in the same major version as it would also be a breaking change, no? That doesn't have to wait for a 4.0?

fsateler:
> Changing its meaning to be the same as an empty string, to match String.Join, or disallowing it completely can be made in the same major version as it would also be a breaking change, no? That doesn't have to wait for a 4.0?

[...]

Issue #284, opened 2017-05-04
Wiki explaning how to contribute


leandromoh:
I imagine it's common to ask which files I must update after contribute (e.g., readme.md, project.json, etc). If the way that project builds changes (e.g. port for .net core), the files I must update might change too (ie. project.json dont exists anymore).

I think it would be very cool to have a wiki explaning the basic steps to contribute.
Maybe it would turn the project more attractive for new contributors.

jnyrup:
To make it more visible on GitHub, I suggest putting it in a [CONTRIBUTING.md](https://help.github.com/articles/setting-guidelines-for-repository-contributors/).
Something like:
* All tests must pass
* Code coverage of tests?
* Code style?
* Necessary documentation
* Which branch to target

Eventually together with a [PR template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/). 
I like them as I've more than once forgot/overlooked guidelines before submitting a PR.

GitHub guides about [building a strong community](https://help.github.com/articles/helping-people-contribute-to-your-project/)

PR #373, opened 2018-02-01
Remove delimiter-less ToDelimitedString overloads

atifaziz:
This PR addresses #229.

The overloads have been obsolete since [version 2.1](https://github.com/morelinq/MoreLINQ/releases/tag/v2.1.0).

### DIFF
--- a/MoreLinq/ToDelimitedString.cs
+++ b/MoreLinq/ToDelimitedString.cs
@@ -26,36 +26,13 @@ namespace MoreLinq
     static partial class MoreEnumerable
     {
         /// <summary>
-        /// Creates a delimited string from a sequence of values. The
-        /// delimiter used depends on the current culture of the executing thread.
-        /// </summary>
-        /// <typeparam name="TSource">Type of element in the source sequence</typeparam>
-        /// <param name="source">The sequence of items to delimit. Each is converted to a string using the
-        /// simple ToString() conversion.</param>
-        /// <returns>
-        /// A string that consists of the elements in <paramref name="source"/>
-        /// delimited by <see cref="TextInfo.ListSeparator"/>. If the source
-        /// sequence is empty, the method returns an empty string.
-        /// </returns>
-        /// <remarks>
-        /// This operator uses immediate execution and effectively buffers the sequence.
-        /// </remarks>
-
-        [Obsolete]
-        public static string ToDelimitedString<TSource>(this IEnumerable<TSource> source)
-        {
-            return ToDelimitedString(source, null);
-        }
-
-        /// <summary>
         /// Creates a delimited string from a sequence of values and
         /// a given delimiter.
         /// </summary>
         /// <typeparam name="TSource">Type of element in the source sequence</typeparam>
         /// <param name="source">The sequence of items to delimit. Each is converted to a string using the
         /// simple ToString() conversion.</param>
-        /// <param name="delimiter">The delimiter to inject between elements. May be null, in which case
-        /// the executing thread's current culture's list separator is used.</param>
+        /// <param name="delimiter">The delimiter to inject between elements.</param>
         /// <returns>
         /// A string that consists of the elements in <paramref name="source"/>
         /// delimited by <paramref name="delimiter"/>. If the source sequence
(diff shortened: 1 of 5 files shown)


MoreLinq/ToDelimitedString.cs
```diff
@@ -25,37 +25,14 @@ namespace MoreLinq
...
         /// <summary>
         /// Creates a delimited string from a sequence of values and
         /// a given delimiter.
         /// </summary>
         /// <typeparam name="TSource">Type of element in the source sequence</typeparam>
         /// <param name="source">The sequence of items to delimit. Each is converted to a string using the
         /// simple ToString() conversion.</param>
-        /// <param name="delimiter">The delimiter to inject between elements. May be null, in which case
-        /// the executing thread's current culture's list separator is used.</param>
+        /// <param name="delimiter">The delimiter to inject between elements.</param>
```
fsateler: The new `<exception>` tag for the delimiter should be added.

fsateler:
> Looks good, except that the new exception should be documented, and these breaking changes should be documented somewhere as an upgrade guide.

Should we add a `CHANGES.md` file ?

MoreLinq/ToDelimitedString.cs
atifaziz: If you insist :) although we haven't been doing that consistently.

atifaziz:
> Should we add a `CHANGES.md` file ?

Unless someone can help automate that through a discipline & process (as I've seen employed in some projects) we use then I don't want to add more to my plate in terms of maintenance at this point. Having said, I'm currently relying on [milestones](https://github.com/morelinq/MoreLINQ/milestones) to help. Each [release](https://github.com/morelinq/MoreLINQ/releases) is linked to its milestone. [Issues that introduce breaking changes are marked so](https://github.com/morelinq/MoreLINQ/issues?q=is%3Aissue+is%3Aopen+label%3Abreaking-change). So for release 3.0, I'm expecting that people will be able to quickly consult [the breaking changes there](https://github.com/morelinq/MoreLINQ/issues?utf8=%E2%9C%93&q=is%3Aissue+milestone%3A3.0.0+label%3Abreaking-change+) and follow-up the rational that's been captured in the discussion of the issue that interests them. I admit this doesn't answer the “upgrade guide” bit but it's better than nothing. And as I said, if I can get another pair of hands then I'd vote for a `CHANGES.md` too. However, I feel that #284 (contribution guide) and #247 (style guide) might trump that in priority. What do you think?

fsateler:
> Unless someone can help automate that through a discipline & process (as I've seen employed in some projects) we use then I don't want to add more to my plate in terms of maintenance at this point. 

I can totally understand this.

> Having said, I'm currently relying on milestones to help. Each release is linked to its milestone. Issues that introduce breaking changes are marked so. So for release 3.0, I'm expecting that people will be able to quickly consult the breaking changes there and follow-up the rational that's been captured in the discussion of the issue that interests them. I admit this doesn't answer the “upgrade guide” bit but it's better than nothing. And as I said, if I can get another pair of hands then I'd vote for a CHANGES.md too. However, I feel that #284 (contribution guide) and #247 (style guide) might trump that in priority. What do you think?

I don't think the CHANGES.md file needs to be too high maintainance. I think simply reusing the content put in the github release page would be sufficient. This has the advantage that once this file is in existence, it can be filled by the PR submitter, thus overall reducing the release load (as then there's only a copy-paste into the github release description).

I think I used the wrong description with "upgrade guide". A thourough guide is not what I meant. I mean a simple bullet-point list of things to take care when upgrading. In this case, the entry could be as simple as:

- The `ToDelimitedString` overload with no delimiter has been removed. If you want to preserve the exact behavior as previously, pass `TextInfo.ListSeparator` as delimiter.
- `ToDelimitedString` no longer accepts `null` as its delimiter parameter. Pass `TextInfo.ListSeparator` as delimiter if you want to mimic the old behavior.

Does this sound reasonably low-overhead?

atifaziz:
> This has the advantage that once this file is in existence, it can be filled by the PR submitter

Right but this create a process that needs to be documented in a contribution guide. The advantage of release pages is that they can be changed, updated & improved _after_ the release whereas a `CHANGES.md` that's part of the repo kind of pollutes the project history and would need maintenance of release branches. Perhaps a compromise would be a wiki per release that can be community edited? In any case, I feel we should be having this discussion outside of this PR. Speaking of the PR, we good to merge this? I think I've addressed everything from [your review](https://github.com/morelinq/MoreLINQ/pull/373#pullrequestreview-93364600).

fsateler:
> Right but this create a process that needs to be documented in a contribution guide. 

Right.

> The advantage of release pages is that they can be changed, updated & improved after the release whereas a CHANGES.md that's part of the repo kind of pollutes the project history and would need maintenance of release branches. Perhaps a compromise would be a wiki per release that can be community edited? 

This might be a good idea.

> In any case, I feel we should be having this discussion outside of this PR. Speaking of the PR, we good to merge this? I think I've addressed everything from your review.

Yes, LGTM.

PR #375, opened 2018-02-03
WIP ToDataReader method

sholland1:
This method converts an `IEnumerable<TSource>` to an `IDataReader`. I use something like this to give data to `SqlBulkCopy`, and I thought it would fit nicely in MoreLINQ.

This blog post was the inspiration for this method: http://www.michaelbowersox.com/2011/12/22/using-a-custom-idatareader-to-stream-data-into-a-database/

I left a lot of the `IDataReader` methods unimplemented because they are necessary for usage with `SqlBulkCopy`.


atifaziz:
This is marked as work-in-progress (WIP) but there have been no updates since it was opened. I'm going to close this PR but happy to reconsider and re-open if there's a strong desire to see this through to a complete and robust implementation.

PR #379, opened 2018-02-06
Bump minimum target framework version to 4.5.1

fsateler:
As per the .Net framework lifecycle policy as of today, all versions below 4.6
are unsupported (except 3.5 but we don't support that anyway).

https://support.microsoft.com/eu-es/help/17455/lifecycle-faq-net-framework

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,4 +43,4 @@ script:
   - ./msbuild.sh /v:m /p:Configuration=$CONFIGURATION
   - dotnet exec MoreLinq.Test/bin/$CONFIGURATION/netcoreapp1.0/MoreLinq.Test.dll
   - dotnet exec MoreLinq.Test/bin/$CONFIGURATION/netcoreapp2.0/MoreLinq.Test.dll
-  - mono MoreLinq.Test/bin/$CONFIGURATION/net451/MoreLinq.Test.exe
+  - mono MoreLinq.Test/bin/$CONFIGURATION/net46/MoreLinq.Test.exe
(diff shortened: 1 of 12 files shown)


atifaziz:
> all versions below 4.6 are unsupported

Where do you see that? Per the policy, 4.5.2 is bound with the lifecycle policy of the parent operating system:

> **.NET Framework 4.5.2:** Support for .NET 4.5.2 follows the lifecycle policy of the parent operating system. It is supported as a Windows component on the latest required operating system update for Windows Vista SP2, Windows 7 SP1, Windows Server 2008 SP2, Windows Server 2008 R2 SP1, Windows 8.1 Update, Windows Server 2012, and Windows Server 2012 R2.

Considering the [product lifecycle of Windows 2012 R2 Standard](https://support.microsoft.com/eu-es/lifecycle/search?alpha=Windows%20Server%202012%20R2%20Standard), .NET 4.5.2 would be supported at least as far out as 2023, though mainstream support ends October of this year:

| Products Released                       | Lifecycle Start Date | Mainstream Support End Date | Extended Support End Date |
|-----------------------------------------|----------------------|-----------------------------|---------------------------|
| Windows Server 2012 R2 Standard         | 25/11/2013           | 09/10/2018                  | 10/10/2023                |
| Windows Storage Server 2012 R2 Standard | 25/11/2013           | 09/10/2018                  | 10/10/2023                |

Anyway, we don't work for Microsoft so we're not bound by those dates, just the threshold for pain to support a particular target. We can certainly drop 4.5.1:

> **.NET Framework 4.5.1:** Support for .NET Framework 4.5.1 on all operating systems ended on January 12, 2016.

Any reason to skip 4.5.2 and bump to 4.6 here?

fsateler:
>> all versions below 4.6 are unsupported
>
> Where do you see that? Per the policy, 4.5.2 is bound with the lifecycle policy of the parent operating system

Oh dear. Embarrassing. It appears I need glasses 👓 .

> Any reason to skip 4.5.2 and bump to 4.6 here?

No, not at all. I'll rework to use 4.5.2

> Anyway, we don't work for Microsoft so we're not bound by those dates, just the threshold for pain to support a particular target. 

We are not bound, I agree. Yet we don't need to take on the pain of supporting old stuff for too long. I find that the upstream support timeframes are a good [Schelling point](https://en.wikipedia.org/wiki/Focal_point_(game_theory)) for providing upper limits for target support.

PR #383, opened 2018-02-14
optimize TakeLast for read-only collections #128

Arithmomaniac:
@Milyli ❤️s fixing #128


MoreLinq/TakeLast.cs
```diff
@@ -49,9 +49,11 @@ public static IEnumerable<TSource> TakeLast<TSource>(this IEnumerable<TSource> s
         {
             if (source == null) throw new ArgumentNullException(nameof(source));
 
-            return 
+            return
                 source is ICollection<TSource> col
                 ? col.Slice(Math.Max(0, col.Count - count), int.MaxValue)
+                : source is IReadOnlyCollection<TSource> readOnlyCol
+                ? readOnlyCol.Slice(Math.Max(0, readOnlyCol.Count - count), int.MaxValue)
```
atifaziz: `col.Slice(Math.Max(0, col.Count - count), int.MaxValue)` reads identical to `readOnlyCol.Slice(Math.Max(0, readOnlyCol.Count - count), int.MaxValue)` and so the two should be consolidated via, perhaps, a local function.

leandromoh:
@atifaziz here is one example of usage for GetCheapCount, instead of create a local function.

atifaziz:
@leandromoh Sure, but how many more are there?

PR #384, opened 2018-02-14
optimize FallbackIfEmpty for IReadOnlyCollection

Arithmomaniac:
The tests here, as in #383, use `I(ReadOnly)List` in their collections despite being optimized at the collection level more generally. If that is a problem, I can update them both (though this was already a problem when using simple arrays for test parameters).

### DIFF
--- a/MoreLinq/FallbackIfEmpty.cs
+++ b/MoreLinq/FallbackIfEmpty.cs
@@ -163,14 +163,12 @@ namespace MoreLinq
         {
             if (source is ICollection<T> collection)
             {
-                if (collection.Count == 0)
-                {
-                    return Fallback();
-                }
-                else
-                {
-                    return collection;
-                }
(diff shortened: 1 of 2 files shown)


atifaziz:
📝 This PR adds to #128.

MoreLinq/FallbackIfEmpty.cs
```diff
@@ -163,14 +163,12 @@ public static IEnumerable<T> FallbackIfEmpty<T>(this IEnumerable<T> source, IEnu
...
-                    return Fallback();
-                }
-                else
-                {
-                    return collection;
-                }
+                return collection.Count == 0 ? Fallback() : collection;
+            }
+
+            if (source is IReadOnlyCollection<T> readOnlyCollection)
```
atifaziz: Like in #383, this can be consolidated with the above as both lead to the same results (`Fallback()` or `source`) predicated only on the collection count.

Issue #386, opened 2018-02-15
GetCheapCount internal method


leandromoh:
Almost always when we try to cast a sequence for collection is about its `Count` property. 

Now that we are also checking for `IReadOnlyCollection`, It is becomming a bit repetitive and with a lot of ternarious opetator. To avoid this I suggest To create the `GetCheapCount` internal method. This one tries all the possible casts and return the `count` If avaible.

Initial signature suggestion:

`Internal static int? GetCheapCount<T>(this IEnumerable<T> souce)`

atifaziz:
Sometimes you need the count _and_ the collection type you cast to.

> It is becomming a bit repetitive and with a lot of ternarious opetator.

Do you have a list of places where this would be make savings? If you want to demonstrate that through a PR then that's fine too (sometimes code can speak volumes) as long as you're not disheartened if we decide to abandon it at the end; that is, if the savings don't seem worthwhile.

Arithmomaniac:
Since an `IReadOnlyCollection` is just an `IEnumerable` with a count, you should be able to use it in any of the optimizations listed in #128 for `IReadOnlyCollection`.

Arithmomaniac:
You could similarly catch all of the `IReadOnlyList` cases, too. If I was to implement this, I would have the two following method signatures:

`bool IsCollectionType<T>(this IEnumerable<T> enumerable, out int count)`

`bool IsListType<T>(this IEnumerable<T> enumerable, out int count, out Indexer<T> indexer)`, using `delegate T Indexer<T>(IEnumerable<T> enumerable, int index)`

Using `out` works very nicely with the C#7 out conditions, as you can check for the Collection/List case with a ternary expression.

Arithmomaniac:
@leandromoh do you plan on working on this? If not, I'll do it this coming week.

atifaziz:
`IsCollectionType` should instead be called `TryGetCollectionCount` because it says more clearly what it's doing:

```c#
bool TryGetCollectionCount<T>(this IEnumerable<T> source, out int count) // can live with
int? TryGetCollectionCount<T>(this IEnumerable<T> source) // preferred and as effective
```

The second signature allows you to use pattern matching, as in `source.TryGetCollectionCount() is int count` and reads better.

For the list, I would open a separate issue/PR.

leandromoh:
@Arithmomaniac No, I am not planning to work on this. Go ahead =)

[...]

PR #391, opened 2018-02-21
Add TryGetCollectionCount

Arithmomaniac:
Implements #386.


Arithmomaniac:
#383 and #384 are on hold pending this one.

MoreLinq/InternalEnumerableExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    static class InternalEnumerableExtensions
+    {
+        /// <summary>
+        /// Gets the count of an enumerable if it is precalculated in <see cref="ICollection{T}"/>
+        /// or <see cref="IReadOnlyCollection{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of objects to enumerate.</typeparam>
+        /// <param name="source">The enumerable to get the count for.</param>
+        /// <returns>The count if it is available without enumeration, or null if it is not.</returns>
+        public static int? TryGetCollectionCount<T>(this IEnumerable<T> source)
```
atifaziz: We don't need to introduce a whole new type just to house this method. There's already `MoreEnumerable` so kindly move this to `MoreEnumerable.cs` and remove the `public` modifier so it becomes private. The effect will be the same but with less ceremony.

Arithmomaniac: Right. Forgot everything is in a single class (I need to get out more).

MoreLinq/MoreEnumerable.cs
```diff
@@ -27,5 +27,18 @@ namespace MoreLinq
     
     public static partial class MoreEnumerable
     {
+        /// <summary>
```
atifaziz: _Trivial_ private functions where the code speaks for itself means doc summaries feel redundant and add little value (even reads longer than the code) therefore best to remove it.

MoreLinq/MoreEnumerable.cs
```diff
@@ -27,5 +27,18 @@ namespace MoreLinq
...
+        /// Gets the count of an enumerable if it is precalculated in <see cref="ICollection{T}"/>
+        /// or <see cref="IReadOnlyCollection{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of objects to enumerate.</typeparam>
+        /// <param name="source">The enumerable to get the count for.</param>
+        /// <returns>The count if it is available without enumeration, or null if it is not.</returns>
+        static int? TryGetCollectionCount<T>(this IEnumerable<T> source)
```
atifaziz: Shouldn't we be validating that the `source` argument is not `null`?

Arithmomaniac: Yes.

MoreLinq/MoreEnumerable.cs
Arithmomaniac: I like having the Intellisense (I find the `out int` pattern to be more intuitive for this sort of thing, and others might as well), but I can do this.

MoreLinq/MoreEnumerable.cs
```diff
@@ -34,7 +34,7 @@ public static partial class MoreEnumerable
         /// <typeparam name="T">The type of objects to enumerate.</typeparam>
         /// <param name="source">The enumerable to get the count for.</param>
         /// <returns>The count if it is available without enumeration, or null if it is not.</returns>
-        private static int? TryGetCollectionCount<T>(this IEnumerable<T> source)
```
Arithmomaniac: This rule should probably be added to `.editorconfig`.

MoreLinq/MoreEnumerable.cs
atifaziz: > I find the `out int` pattern to be more intuitive

I'm not sure I understand your reasoning here. There's no `out` variable here.

> I find…to be more intuitive…others might as well

It's best to avoid being subjective here and I can't speak for others. The things we should challenge are:

- Are method and argument names intuitive?
- Does the return type reflect the method name as best & succinctly as possible?
- The value of the documentation on top of everything that the code is communicating.

The IntelliSense argument is fair for public members but this is private, internal and trivial.

Arithmomaniac: > I'm not sure I understand your reasoning here. There's no out variable here.

This had to do with my assertion that `out int` was more idiomatic, which we concluded it was not already; I shouldn't have brought it up again. Sorry.

> The IntelliSense argument is fair for public members but this is private, internal and trivial.

Fair.

PR #393, opened 2018-02-21
Fixes #392

### LINKED ISSUE #392: Scan iterates too eagerly
When doing Scan (regular or seeded), the source gets iterated once too many times.

Scan should yield the first element right after it has been read from the source, but it waits for the next element to be read instead.

I've put together a PR to fix this.

quassnoi:
Fixes #392

### DIFF
--- a/MoreLinq.Test/ScanTest.cs
+++ b/MoreLinq.Test/ScanTest.cs
@@ -1,13 +1,13 @@
 #region License and Terms
 // MoreLINQ - Extensions to LINQ to Objects
 // Copyright (c) 2008 Jonathan Skeet. All rights reserved.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


MoreLinq.Test/ScanTest.cs
```diff
@@ -43,6 +43,15 @@ public void ScanIsLazy()
             new BreakingSequence<object>().Scan(delegate { throw new NotImplementedException(); });
         }
 
+        [Test]
```
atifaziz: The `Enumerable.Range(1, 3).Concat(new BreakingSequence<int>()).Scan(SampleData.Plus)` is repeated twice. Can we consolidate such duplication? For example:

```c#
        [Test]
        public void ScanDoesNotIterateExtra()
        {
            var xs =
                Enumerable.Range(1, 3)
                          .Concat(new BreakingSequence<int>())
                          .Scan(SampleData.Plus);
            Assert.Throws<InvalidOperationException>(() => xs.Consume());
            xs.Take(3).AssertSequenceEqual(1, 3, 6);
        }
```

Ditto for `SeededScanDoesNotIterateExtra`.

MoreLinq.Test/ScanTest.cs
```diff
@@ -54,6 +63,8 @@ public void SeededScanSum()
         {
             var result = Enumerable.Range(1, 10).Scan(0, SampleData.Plus);
             var gold = new[] { 0, 1, 3, 6, 10, 15, 21, 28, 36, 45, 55 };
+            Assert.Throws<InvalidOperationException>(() =>
+                Enumerable.Range(1, 3).Concat(new BreakingSequence<int>()).Scan(SampleData.Plus).Consume());
```
atifaziz: What's the intention here? It doesn't use `result` and a different overload of `Scan` is being called. Seems like a copy & paste error from [line 51](https://github.com/morelinq/MoreLINQ/pull/393/files#diff-071fe005ca9d2e65c1d052c88657c410R51)?

MoreLinq.Test/ScanTest.cs
quassnoi: fixed

PR #398, opened 2018-03-02
Upgrade to C# 7.2

atifaziz:
~~This is to primarily help get rid of more syntactic redundancies.~~

The only benefit for now would be to use:

- [Non trailing named-arguments](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-2#non-trailing-named-arguments)


atifaziz:
Fails for non-Windows build:

```
CSC : error CS1617: Invalid option '7.2' for /langversion; must be ISO-1, ISO-2, Default, Latest or a valid version in range 1 to 7.1. [/home/travis/build/morelinq/MoreLINQ/MoreLinq/MoreLinq.csproj]
```

Need to investigate why.

atifaziz:
Most syntactic redundancies are addressed by upgrading to C# 7.1 in #399, so this is not urgent.

atifaziz:
Yay, I'm sure #420 had something to do with this as the merge that is 0cd0c8c97b2464b551a02d2d965cf3a4b3c0bb55, is passing. :tada: While this is ready to merge, let's delay it until we're absolutely itching to need something from C# 7.2 to force it as a baseline.

atifaziz:
Superseded by #566.

Issue #400, opened 2018-03-02
Optimize Batch to return lazy enumerables


Arithmomaniac:
This takes the point of #98 and runs it in an opposite direction: What if `Batch` really only returned lazy `IEnumerable<T>`s for each batch, without getting all of the items in the batch first?

This [StackOverflow answer](https://stackoverflow.com/questions/419019/split-list-into-sublists-with-linq/10425490#10425490) contains such an implementation.

atifaziz:
Best if @SamSaffron, whose implementation you cite, drops by himself and contributes the implementation along with tests. 😄 Short of that, I don't see it happening any time soon as it's a fairly involved implementation to undertake. On a quick run, I noticed it does multiple passes over the source (possibly by design). I didn't dig in to understand why because I don't have the bandwidth for it right now (involved in far too many projects for my own good, I'm afraid). I'm happy if you are volunteering to investigate. I reckon it's not an easy problem.

Arithmomaniac:
Not volunteering; I myself barely have enough time to keep up with #128 🙁

leandromoh:
If I understood right, my sketch bellow does that. As well the [stackoverflow asnwer](https://stackoverflow.com/questions/419019/split-list-into-sublists-with-linq/10425490#10425490), it is not thread safe.

```
public static IEnumerable<IEnumerable<T>> Batch<T>(this IEnumerable<T> source, int size)
{
    if (source == null) throw new Exception();

    return BatchImpl(source, size).TakeWhile(x => x.Any());
}

static IEnumerable<IEnumerable<T>> BatchImpl<T>(this IEnumerable<T> source, int size)
{
    var e = source.GetEnumerator();
    var list = new List<T>();
    var i = 1;

    while (true)
    {
        yield return Oi(e, list, i, size);
        i++;
    }
}

static IEnumerable<T> Oi<T>(IEnumerator<T> e, List<T> list, int group, int size)
{
    var min = (group - 1) * size + 1;
    var max = group * size;
    var hasValue = false;

    while (list.Count < min && e.MoveNext())
    {
        list.Add(e.Current);
    }

    for (var i = min; i <= max; i++)
    {
        if (i <= list.Count)
        {
            hasValue = true;
        }
        else if (hasValue = e.MoveNext())
        {
            list.Add(e.Current);
        }

        else
        {
            e.Dispose();
        }

        if (hasValue)
            yield return list[i - 1];
        else
            yield break;
    }
}
```

[...]

PR #402, opened 2018-03-03
Batch with lazy enumerables (issue #400)

leandromoh:
implements the requested of [Optimize Batch to return lazy enumerables](https://github.com/morelinq/MoreLINQ/issues/400).  

As requested [here](https://github.com/morelinq/MoreLINQ/issues/400#issuecomment-370135245), all batches are buffered in different lists. So when you move on to the next batch, the one before becomes fodder for GC.

### DIFF
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -41,13 +41,21 @@ namespace MoreLinq.Test
         public void BatchEvenlyDivisibleSequence()
         {
             var result = new[] { 1, 2, 3, 4, 5, 6, 7, 8, 9 }.Batch(3);
+            IEnumerable<IEnumerable<int>> batch2;
+
             using (var reader = result.Read())
             {
                 reader.Read().AssertSequenceEqual(1, 2, 3);
-                reader.Read().AssertSequenceEqual(4, 5, 6);
+
+                batch2 = reader.Read();
+
+                batch2.AssertSequenceEqual(4, 5, 6);
                 reader.Read().AssertSequenceEqual(7, 8, 9);
+
                 reader.ReadEnd();
             }
+
+            batch2.AssertSequenceEqual(4, 5, 6);
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


leandromoh:
@atifaziz @arithmomaniac
What do you think?

Arithmomaniac:
I have my hands full, so unless @atifaziz asks me to do otherwise, I'm not going to review this.

atifaziz:
@Arithmomaniac I've got my hand full here too so I doubted I'll get round to reviewing this any time soon. There's another reason…we never decided in #400 that there is indeed a problem that needs and what the solution would be so this PR jumps the gun for me. If @leandromoh is just sharing an experimenting with an implementation here then that's okay but this isn't going to be a priority for me.

@Arithmomaniac Perhaps #400 should be closed seeing that no one has the time and that `Batch` isn't burning down the house for anymore? 😉

Arithmomaniac:
@atifaziz I'm fine with this, but we should [have a tag for things that were closed due to lack of interest](https://github.com/moq/moq4/issues/505#issuecomment-354365114) if we don't already.

atifaziz:
@Arithmomaniac There's already [the won't-fix label][wontfix] for that but honestly, a closed issue that's not been tagged as either an [enhancement][enhancement], a [bug][bug] or a [task][task] is essentially something that's been abandoned. Perhaps won't-fix is too strong for #400 as I feel it doesn't have a problem statement?

[wontfix]: https://github.com/morelinq/MoreLINQ/issues?q=label%3Awontfix
[enhancement]: https://github.com/morelinq/MoreLINQ/issues?q=label%3Aenhancement
[bug]: https://github.com/morelinq/MoreLINQ/issues?q=label%3Abug
[task]: https://github.com/morelinq/MoreLINQ/issues?q=label%3Atask

Arithmomaniac:
Maybe "won't-implement"?

leandromoh:
I cant see why close the issue If It was not resolved.  

There is a lot of issues openned even without a PR. So why close this, that does?

atifaziz:
Closing since #400 has been closed too.

PR #414, opened 2018-03-17
Shared EquiZip, ZipLongest & ZipShortest implementation

leandromoh:
This PR refactored **EquiZip**, **ZipLongest** and **ZipShortest** to use a common Implamentation, as well adds tests that were missing.
So, when extend EquiZip, ZipLongest and ZipShortest to 16 sequences, the change will be made more easily. This new common implementation also fix some important points.

- current [ZipShortest](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/ZipShortest.cs#L154) always consume the first sequence entirely, even when second sequence already ended.

- current [ZipLongest](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/ZipLongest.cs#L81) dont dispose sequences as soon as possible. Dispose occurs just when all sequences finished.


leandromoh:
@atifaziz I would like to merge this before continue with the extensions of zip's to 16 sequences. Can you or @fsateler do the review?

MoreLinq/ZipImpl.cs
```diff
@@ -0,0 +1,96 @@
...
+                    if (disposed <= limit)
+                        yield return resultSelector(v1, v2, v3, v4);
+                    else
+                        yield break;
+                }
+            }
+            finally
+            {
+                e1?.Dispose();
+                e2?.Dispose();
```
atifaziz: `e1` and `e2` can never be `null` so don't defend against it.

MoreLinq/ZipImpl.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+                else
+                {
+                    e.Dispose();
+                    e = null;
+                    disposed++;
+                    return ValidateEquiZip(default);
+                }
+
+                T ValidateEquiZip(T value)
```
atifaziz: This seems like an abuse of local functions. Seems like you want common exit code, which is better expressed more classically as:

```c#
T GetValue<T>(ref IEnumerator<T> e)
{
    call++;
    if (e == null || disposed > limit)
    {
        return default;
    }

    T value;
    if (e.MoveNext())
    {
        value = e.Current;
    }
    else
    {
        e.Dispose();
        e = null;
        disposed++;
        value = default;
    }

    if (error != null && disposed > 0 && disposed < call)
        error(new IEnumerator[]{ e1, e2, e3, e4 });

    return value;
}
```

leandromoh:
@atifaziz all ajusts were done. about the new tests, all are related to the refactoring. the tests check about unnecessary computations, and fails for the current implementations.

MoreLinq/ZipImpl.cs
```diff
@@ -0,0 +1,100 @@
...
+            )
+        {
+            IEnumerator<T1> e1 = null;
+            IEnumerator<T2> e2 = null;
+            IEnumerator<T3> e3 = null;
+            IEnumerator<T4> e4 = null;
+            var disposals = 0;
+            int calls;
+
+            try
```
atifaziz: why not replace the try-finally block here with just `using` blocks, like below?

```c#
            IEnumerator<T1> e1;
            IEnumerator<T2> e2;
            IEnumerator<T3> e3;
            IEnumerator<T4> e4;
            var disposals = 0;
            int calls;

            using (e1 = s1 .GetEnumerator())
            using (e2 = s2 .GetEnumerator())
            using (e3 = s3?.GetEnumerator())
            using (e4 = s4?.GetEnumerator())
            {

                while (true)
                {
                    calls = 0;
                    var v1 = GetValue(ref e1);
                    var v2 = GetValue(ref e2);
                    var v3 = GetValue(ref e3);
                    var v4 = GetValue(ref e4);

                    if (disposals <= limit)
                        yield return resultSelector(v1, v2, v3, v4);
                    else
                        yield break;
                }
            }
```

Less code and reads much simpler.

MoreLinq/EquiZip.cs
```diff
@@ -152,38 +153,28 @@ static partial class MoreEnumerable
...
-                if (e2.MoveNext() || (e3 != null && e3.MoveNext())
-                                  || (e4 != null && e4.MoveNext()))
-                {
-                    throw new InvalidOperationException("First sequence too short.");
-                }
+                if (i == 1) return "First";
+                if (i == 2) return "Second";
+                if (i == 3) return "Third";
+                if (i == 4) return "Fourth";
+                return i + "th";
```
atifaziz: This is not a correct translation of the old code. If `i` is out of bounds then throw. If you use an array as suggested earlier then just index into it.

MoreLinq/EquiZip.cs
```diff
@@ -152,38 +153,28 @@ static partial class MoreEnumerable
...
+            var limit = new object[] { s1, s2, s3, s4 }.TakeWhile(s => s != null).Count() - 1;
+
+            return ZipImpl(s1, s2, s3, s4, resultSelector, limit, enumerators =>
+            {
+                int shortest = enumerators.Index().First(x => x.Value == null).Key;
+
+                return new InvalidOperationException(GetOrdinal(shortest + 1) + " sequence too short.");
+            });
+
+            string GetOrdinal(int i)
```
atifaziz: Use an array?

```c#
static readonly string[] OrdinalNumbers =
{
    "First",
    "Second",
    "Third",
    "Fourth",
};
```

It will help when extending the arity of zipping operators.

MoreLinq/EquiZip.cs
```diff
@@ -152,38 +153,28 @@ static partial class MoreEnumerable
...
         {
-            using (var e1 = first.GetEnumerator())
-            using (var e2 = second.GetEnumerator())
-            using (var e3 = third?.GetEnumerator())
-            using (var e4 = fourth?.GetEnumerator())
+            var limit = new object[] { s1, s2, s3, s4 }.TakeWhile(s => s != null).Count() - 1;
+
+            return ZipImpl(s1, s2, s3, s4, resultSelector, limit, enumerators =>
+            {
+                int shortest = enumerators.Index().First(x => x.Value == null).Key;
```
atifaziz: `var`

MoreLinq/EquiZip.cs
```diff
@@ -152,38 +153,28 @@ static partial class MoreEnumerable
...
+
+                return new InvalidOperationException(GetOrdinal(shortest + 1) + " sequence too short.");
```
atifaziz: You can use `Index(1)` and get rid of `+ 1` in `shortest + 1`.

MoreLinq/ZipImpl.cs
```diff
@@ -0,0 +1,100 @@
...
+                else
+                {
+                    e.Dispose();
+                    e = null;
+                    disposals++;
+                    value = default;
+                }
+
+                if (errorSelector != null && disposals > 0 && disposals < calls)
+                    throw errorSelector(new IEnumerator[]{ e1, e2, e3, e4 });
```
atifaziz: Space before `{`.

MoreLinq/ZipImpl.cs
```diff
@@ -0,0 +1,100 @@
...
+            int limit,
+            Func<IEnumerator[], Exception> errorSelector = null
+            )
+        {
+            IEnumerator<T1> e1 = null;
+            IEnumerator<T2> e2 = null;
+            IEnumerator<T3> e3 = null;
+            IEnumerator<T4> e4 = null;
+            var disposals = 0;
+            int calls;
```
atifaziz: This should be scoped to _inside_ the while-loop, which means the same for the `GetValue` local function.

leandromoh: This looks odd to me. I expect to see local functions always in the end of the outer function. 
Inside the while-loop the local function difficults the reading of the flow at all, because dont let me see if there is something beetween its end and the end of the loop.

MoreLinq/ZipImpl.cs
leandromoh: Really. I confused me about the "behind the scenes" of using statement.

leandromoh:
@atifaziz improviments done.

MoreLinq/ZipImpl.cs
atifaziz: I understand but that's a side-effect of your design. If you don't like local functions in the middle of your code then you need to pass more of the context as arguments, like `calls`. It's much harder to reason about the data flow when the scope of `calls` needs to be moved out of the loop for the artificial purpose of having the local function at the bottom of the enclosing method's body. For example, send `calls` by reference to `GetValue` and keep the local function at the bottom as you have now:

```c#
T GetValue<T>(ref IEnumerator<T> e, ref int calls)
{
    // implementation omitted.
}
```

Now you can scope `calls` to the loop:

```c#
while (true)
{
    var calls = 0;
    var v1 = GetValue(ref e1, ref calls);
    var v2 = GetValue(ref e2, ref calls);
    var v3 = GetValue(ref e3, ref calls);
    var v4 = GetValue(ref e4, ref calls);

    if (disposals <= limit)
        yield return resultSelector(v1, v2, v3, v4);
    else
        yield break;
}
```

I still find it strange though to see a variable in a loop where _it's never used_ in the body of the loop. You have to look _inside_ `GetValue` to understand what's really going on. As I said, before, the final design might become clearer with time. For the moment, I am looking for _mechanical_ clarity. In that light, I feel it's even better to just do this:

```c#
using (e1 = s1 .GetEnumerator())
using (e2 = s2 .GetEnumerator())
using (e3 = s3?.GetEnumerator())
using (e4 = s4?.GetEnumerator())
{
    while (true)
    {
        var calls = 0;
        var v1 = GetValue(ref e1, ++calls);
        var v2 = GetValue(ref e2, ++calls);
        var v3 = GetValue(ref e3, ++calls);
        var v4 = GetValue(ref e4, ++calls);

        if (disposals <= limit)
            yield return resultSelector(v1, v2, v3, v4);
        else
            yield break;
    }
}

T GetValue<T>(ref IEnumerator<T> e, /* ref */ int calls)
{
    // call++;
    // rest omitted
}
```

So `calls` doesn't need to be passed by-reference, it can remain in the scope of the loop and it's _partially_ clear how it's being used (I say partially because we see it's being incremented unconditionally but we don't see how `GetValue` uses it).

leandromoh: I see and agree that passing `calls `as argument to `GetValue` is the better we can do. you are right, it is better avoid side-effects for clarity.

atifaziz:
> Something more before merge ?

I'm still not satisfied with how `ZipImpl` reads as it's very hard to reason through the code. I've tried to improve the names but `limit` is still bothering me a lot. Otherwise the implementation is really solid in the sense that that it's really well-covered by the tests such that tests break no matter what little I change. 👍 

> I would like to work extending to 16 this weekend.

Have we talked about how you plan to go about this? Do you plan to use code generation?

leandromoh:
> I've tried to improve the names but limit is still bothering me a lot. 

what do you think about the `disposalsLimit` name?

> Do you plan to use code generation?

Yes, using C#. I will create a function that based in its argument, returns a string that is the function code. so we copy it and paste on the IDE. I know there is others ways to do it, but I did not learn about them, I prefered use C#, and turn the thing more programmatically. Here is an example of this approach on my Curryfy project: [This ](https://github.com/leandromoh/Curryfy/blob/master/Curry/Curry.cs#L84) function generate the code of the Curry' s functions in the same file.

leandromoh:
@atifaziz what do you think about [85d7ceb](https://github.com/morelinq/MoreLINQ/pull/414/commits/85d7ceb8d7a7f0d13012ba1bec578bbf45f01273) ? It help us avoid repetitive validation code.

atifaziz:
> what do you think about the `disposalsLimit` name?

I'm afraid it doesn't give me any more clarity than before. 😕 

>  I know there is others ways to do it, but I did not learn about them

May be this is an opportunity to learn. So far, T4 has been used for code generation, which is templating based on C# and [quite straightforward to learn](https://docs.microsoft.com/en-gb/visualstudio/modeling/writing-a-t4-text-template). See also `*.tt` files in the repo for help.

leandromoh:
I understood about the variable name. Since It is internal and local variable, we could merge the PR and rename It later when a better name appears without any breaking-change. What do you think? I would not like to stop our progress for this reason.

atifaziz:
> Since It is internal and local variable, we could merge the PR and rename It later when a better name appears without any breaking-change. …I would not like to stop our progress for this reason.

I agree.

leandromoh:
> which I am ready to do if you could revert 85d7ceb

done.

PR #406, opened 2018-03-07
Clean up Unenumerable(ReadOnly)List #385

Arithmomaniac:

Issue #385, opened 2018-02-14
Clean up special test collections/lists


Arithmomaniac:
Refactoring suggestions for MoreLinq.Test:

1) This pattern:
```c#
readOnly
    ? new UnenumerableReadOnlyList<int>(new[] { 1 })
    : (IEnumerable<int>)new UnenumerableList<int> { 1 };
```
Has come up often enough at this point that I think a future commit (or even this one) should consolidate it into a helper method of signature `IEnumerable<T> GetUnenumerableList<T>(bool readOnly, params T[] items)`.

2) `Unenumerable(ReadOnly)List` should inherit from `BreakingSequence`.

3) `Breaking(ReadOnly)Collection` should be replaced with `Unenumerable(ReadOnly)List`.

atifaziz:
Makes sense. The exact shape and form of the consolidation may become clearer and need debating when you submit a PR.

Arithmomaniac:
I'm currently leaning towards an extension method for 1): `IEnumerable<T> ToUnenumerable<T>(this T[] enumerable, bool readOnly)`.

PR #418, opened 2018-03-22
Make BreakingList inherit from BreakingCollection

Arithmomaniac:
As proposed in https://github.com/morelinq/MoreLINQ/pull/406#issuecomment-372410156. Rounds out the rest of #385 .

### DIFF
--- a/MoreLinq.Test/BreakingCollection.cs
+++ b/MoreLinq.Test/BreakingCollection.cs
@@ -22,20 +22,20 @@ namespace MoreLinq.Test
 
     class BreakingCollection<T> : BreakingSequence<T>, ICollection<T>
     {
-        protected readonly List<T> _list;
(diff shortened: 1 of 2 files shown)


MoreLinq.Test/BreakingCollection.cs
```diff
@@ -20,17 +20,22 @@ namespace MoreLinq.Test
     using System;
     using System.Collections.Generic;
 
-    sealed class BreakingCollection<T> : BreakingSequence<T>, ICollection<T>
+    class BreakingCollection<T> : BreakingSequence<T>, ICollection<T>
     {
-        public BreakingCollection(int count) => Count = count;
+        protected readonly List<T> _list;
```
atifaziz: Rename to `List` as it's not private.

PR #421, opened 2018-03-23
Update ForEach.cs

jslicer:
Added AggregateException much like Parallel.ForEach accumulates them.


Orace:
`ForEach` is a shortcut to a simple `foreach` call.
What you provide have a different behavior. Maybe you should call it `ForEachTry` but put a try overload on each Linq/MoreLinq method is a huge job.

atifaziz:
Closing for the same reasons #515 was abandoned. It is not reliable to continue to iterate a source once an exception has occurred.

PR #422, opened 2018-03-26
Avoid multiple enumerations on Repeat and Cartesian operators

leandromoh:


leandromoh:
@atifaziz How about this? Since memoize is internal, we can change the approach if something better appears latter, without any breaking-change.

leandromoh:
@atifaziz how about this?

atifaziz:
> I would like to review this after merging #211 and use that `Memoize` implementation here. 

Done! 🙌

atifaziz:
This needs to be two separate PRs, one for `Catersian` and another for `Repeat`, especially now that `Memoize` is in.

atifaziz:
@leandromoh I have submitted a PR for `Cartesian` (see #490). Unless you see an issue with it, perhaps you'd like to address #489 in a similar way?

PR #424, opened 2018-03-27
Add IReadOnlyCollection support to Counting methods 

leandromoh:
This is related to [#128](https://github.com/morelinq/MoreLINQ/issues/128).

/cc @atifaziz  @Arithmomaniac


Arithmomaniac:
Did you check whether any tests could be updated?

leandromoh:
Yes. Tests were updated too.

leandromoh:
>  I'm guessing the comment for 70cc3eb is wrong? It doesn't seem to have anything to do with PadStart.

Yes. I forgot to replace `PadStart` for `CountingMethods`. Sorry.

>  after the change to implementation then would the test break if I remove the changes. 

The current implementations already were optimized for `ICollections`, but tests did not cover that. So what I did on tests, was to add this coverage as improviment. But you are right, now, tests should check against `IReadOnlyCollections` too, and therefore fails. 

@atifaziz
I added a validation for `IReadOnlyCollection` on [5d2c03c](https://github.com/morelinq/MoreLINQ/pull/424/commits/5d2c03cc8995b7e9b14cd520a783862eaa4834e4), this is why tests were failing, as you expected.  

After this I revert your revert on [4d5c9df](https://github.com/morelinq/MoreLINQ/pull/424/commits/4d5c9df43124f9cd44b6c6a817e8c3b7547d72b7), changing the implementation to be optimized for `IReadOnlyCollection` again.

Arithmomaniac:
@atifaziz I should have time later today.

atifaziz:
> Can you add non-collaborators as formal reviewers?

As far as I understand, anyone with read access can submit a review.

atifaziz:
> But you are right, now, tests should check against `IReadOnlyCollections` too, and therefore fails. 
>
> Now test should correctly pass.

Thanks for breaking _then_ fixing! 👍

> [Optimize PadStart for read-only collections](https://github.com/morelinq/MoreLINQ/commit/d5c2c28ed17dd99c05f5ba3a3f823a7227338f36) should had added converage for `IReadOnlyCollection`  too, no?

I agree.

Arithmomaniac:
I considered optiomize

> Optimize PadStart for read-only collections should had added converage for IReadOnlyCollection too, no?

The tests were not explicitly optimized for `ICollection` either.
I considered adding the optimized tests for `ICollection` and `IReadOnlyCollection`, but unlike most optimized methods, the enumerator is still used in the `Concat` statement, so actually checking the full results using `Breaking(ReadOnly)List` like we usually do is not feasible.
What one _could_ do, in retrospect, is check that `PadStart` can at least be called on a `Breaking(ReadOnly)List`; it would fail even without enumeration without optimization.

MoreLinq.Test/BreakingCollection.cs
```diff
@@ -20,6 +20,14 @@ namespace MoreLinq.Test
...
 
+    partial class TestExtensions
+    {
+        internal static IEnumerable<T> ToBreakingCollection<T>(this IEnumerable<T> enumerable, bool readOnly) =>
+            readOnly
+            ? (IEnumerable<T>)new BreakingReadOnlyCollection<T>(enumerable.ToList())
+            : new BreakingCollection<T>(enumerable.ToList());
+    }
+
     class BreakingCollection<T> : BreakingSequence<T>, ICollection<T>
```
Arithmomaniac: `BreakingCollection` should inherit from `BreakingReadOnlyCollection`.

MoreLinq/CountMethods.cs
```diff
@@ -178,16 +178,16 @@ static bool QuantityIterator<T>(IEnumerable<T> source, int limit, int min, int m
             if (first == null) throw new ArgumentNullException(nameof(first));
             if (second == null) throw new ArgumentNullException(nameof(second));
 
-            if (first is ICollection<TFirst> firstCol)
+            if (first.TryGetCollectionCount() is int firstCount)
```
Arithmomaniac: The updates for `CompareCount` need expanded testing as well.

MoreLinq.Test/AtLeastTest.cs
```diff
@@ -18,6 +18,8 @@
 namespace MoreLinq.Test
 {
     using NUnit.Framework;
+    using System.Collections.Generic;
```
Arithmomaniac: Sort namespaces

MoreLinq/CountMethods.cs
```diff
@@ -178,16 +178,16 @@ static bool QuantityIterator<T>(IEnumerable<T> source, int limit, int min, int m
...
             {
-                return firstCol.Count.CompareTo(second is ICollection<TSecond> secondCol
-                                                ? secondCol.Count
-                                                : PartialCount(second, firstCol.Count + 1));
+                return firstCount.CompareTo(second.TryGetCollectionCount() is int secondCount
```
Arithmomaniac: You could use `??` here.

leandromoh:
> BreakingCollection should inherit from BreakingReadOnlyCollection.

Can you explain me why, please? 

@Arithmomaniac other changes were done, thanks for review.

MoreLinq.Test/BreakingReadOnlyCollection.cs
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq.Test
+{
+    using System;
+    using System.Collections.Generic;
+
+    class BreakingReadOnlyCollection<T> : BreakingSequence<T>, IReadOnlyCollection<T>
```
Arithmomaniac: Now that we have `BreakingReadOnlyCollection`, `BreakingReadOnlyList` should inherit form it (see `BreakingCollection`\`BreakingList` implementation in #418 )

Arithmomaniac:
> Can you explain me why, please?

You are correct; I added the suggestions I actually meant.

leandromoh:
@Arithmomaniac 

> Now that we have BreakingReadOnlyCollection, BreakingReadOnlyList should inherit form it 

Make sense now. Improved already done. Something more you would like to point in your review?

> What one could do, in retrospect, is check that PadStart can at least be called on a Breaking(ReadOnly)List; it would fail even without enumeration without optimization.

we can add a `Take(collectionCount)` on `source `before apply `Concat`, So `MoveNext `will be called exactly `collectionCount` times, not `collectionCount + 1` which is the case when it returns `false`.

Arithmomaniac:
I am on vacation with limited desktop access, but I'll get back to you when
I can.

Re PadStart, we should open a new bug.

-----

leandromoh:
@atifaziz I liked the idea of `ArrangeCollectionTestCases`. I implemented that with `foreach` in
[82116b0](https://github.com/morelinq/MoreLINQ/pull/424/commits/82116b0c28b7d5bc0544ef3143492569bd86e92d).

leandromoh:
@atifaziz  before resolve the conflict I would like to merge PR [#425](https://github.com/morelinq/MoreLINQ/pull/425), that is an "one commit PR".

Arithmomaniac:
> However, I'd also like to explore how this can be refactored into test case data.

Maybe this should be moved to a separate story, but [the new(ish) ability to add parameters to `TestCaseSource`](https://stackoverflow.com/a/35822379/1180926) should do the same thing as `ArrangeCollectionTestCases` while making it possible for each test case to fail independently.

atifaziz:
> Maybe this should be moved to a separate issue

Except will it ever get done then once all is green? 😉 

> while making it possible for each test case to fail independently.

Aye, that's the goal here.

leandromoh:
What do you think about revert the collection validation for all tests and Just create a new test that check It (the approach used by @Arithmomaniac in other PR)

atifaziz:
@leandromoh Any chance you can resolve the conflicts? We should merge this soon(-ish).

leandromoh:
@atifaziz I can do It today. But testing optmization in each test or in a separate one?

leandromoh:
@atifaziz resolved conflicts done!
thanks for [a3dbf2e ](https://github.com/morelinq/MoreLINQ/commit/a3dbf2ed7b6389737c1a91afa8e90ff1452bc9ac), it clarify alot! But I was wondering if this may go to master here or on the refactoring PR.

Arithmomaniac:
> We don't have to wait for “the new(ish) ability to add parameters to TestCaseSource"

Huh? It landed a while ago already.

That's not to the exclusion of the above being brilliant 😃 It's just that it may make a good alternative to code like https://github.com/morelinq/MoreLINQ/blob/8d4e87139888a916c6c9f499596a7460e23630e1/MoreLinq.Test/TakeLastTest.cs#L66-L73

atifaziz:
> It landed a while ago already.

Oh, I see. My bad. I read “which is about to be released” in the [linked SO answer](https://stackoverflow.com/a/35822379/1180926) and didn't even bother checking that we're already on 3.6.1 when it was referring to 3.2.

PR #425, opened 2018-03-30
Reuse PartialCount inside QuantityIterator

leandromoh:


MoreLinq/CountMethods.cs
```diff
@@ -205,22 +191,22 @@ static bool QuantityIterator<T>(IEnumerable<T> source, int limit, int min, int m
 
                 return Convert.ToInt32(firstHasNext).CompareTo(Convert.ToInt32(secondHasNext));
             }
+        }
 
-            int PartialCount<T>(IEnumerable<T> source, int limit)
-            {
-                var count = 0;
+        static int PartialCount<T>(IEnumerable<T> source, int limit)
```
atifaziz: How is this different from doing `source.Take(limit).Count()`?

leandromoh:
`source.Take(limit).Count()` execute the following operations:

1. call `MoveNext` on source `limit` times inside `Take` - complexity O(limit), 
2. `**Take` creates a lazy sequence in heap memory, Garbage Collector will have job here**
3. allocate a `int` variable in stack memory to store the quantity of elements, on `Count`
4. call `MoveNext` `limit` times inside `Count `on the lazy sequence created by `Take `- complexity O(limit), 
5. return the `int` variable

`PartialCount(source, limit)` execute the following operations:

1. call `MoveNext` on source `limit` times inside `PartialCount` - complexity O(limit), 
2. allocate a `int` variable in stack memory to store the quantity of elements, on `PartialCount`
3. return the `int` variable

The difference is that `PartialCount` has complexity O(limit) instead of O(limit) + O(limit) (therefore do less operations, therefore is faster) and does not allocate nothing is stack memory. This is to reduce allocations, as well [f00f5ca](https://github.com/morelinq/MoreLINQ/pull/414/commits/f00f5ca1d2eee11f77b269413c3de9cbfe99feda).

In fact, the `Take + Count` approach was how I saw the implementation of `AtLeast` at first time, the I said about its disadvantages [my first comment in morelinq](https://github.com/morelinq/MoreLINQ/pull/197#issue-86624066) and refactored it in counting methods PR.

atifaziz:
Thanks for the explanation. I was just wondering if there was a _logical difference_; looks like not. I understand the benefits of hand-rolling some LINQ.

leandromoh:
I see. There's no logical difference... In fact, this PR is about DRY principle.

MoreLinq/CountMethods.cs
atifaziz: - Convert `PartialCount` to an extension method so it's reusable but leave it private as it is now
- Rename `PartialCount` to `CountUpto`
- Like for all extension methods, check `source` is non-null
- Rename `limit` to `max`
- Change implementation so it behaves well for `max` ≤ 0 and perhaps even validate that `max` is never negative.

leandromoh:
Ok. If i try do these improviments tonight

leandromoh:
@atifaziz all improviments are done. I just wonder if `CountUpTo` is a better name than `PartialCount`. 
`PartialCount` is pretty fine to me. If you don't like it, maybe `CountUntil` is another option too.

MoreLinq/MoreEnumerable.cs
```diff
@@ -35,5 +35,24 @@ public static partial class MoreEnumerable
                  : source is IReadOnlyCollection<T> readOnlyCollection ? readOnlyCollection.Count
                  : (int?)null;
         }
+
+        static int CountUpTo<T>(this IEnumerable<T> source, int max)
+        {
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (max < 0) throw new ArgumentOutOfRangeException(nameof(max), "Max cannot be negative.");
```
atifaziz: If `max` is allowed to be 0 then this method will produce an incorrect result due to the `++count == max` as the loop breaking condition. It can be made to work with a simpler and more robust while-loop:

```c#
while (count < max && e.MoveNext())
    count++;
```

The exception message should read as follows:

> The maximum count argument cannot be negative.

Because “Max” is not a proper word in English.

leandromoh:
@atifaziz  Adjusts done!

> If max is allowed to be 0 then this method will produce an incorrect result

Good catch!

> Because “Max” is not a proper word in English.

I meant "Max **(parameter)** cannot be negative."

PR #426, opened 2018-03-30
Added Tests About Unnecessary Computation On Counting Methods

leandromoh:


leandromoh:
Glad to hear that. Thanks.

PR #427, opened 2018-03-30
Some improvements in AcquireTest tests

leandromoh:


leandromoh:
> There is a reason for using a custom exception (ApplicationException), that it can never be thrown by Acquire. It makes the test more resilient to future refactorings. There is no benefit to using an exception type from the framework.

I see. I reverted [3bf0d04](https://github.com/morelinq/MoreLINQ/commit/3bf0d046c615edbea9d2a3990347c78f6526a8a4).

Issue #381, opened 2018-02-10
T4 template generation for all IDEs & non-Windows platforms


atifaziz:
T4 is used for code generation, like in [`Fold.g.tt`][Fold.g.tt] and [`ToDelimitedString.g.tt`][ToDelimitedString.g.tt]. T4 is great when you have VS as your primary IDE and building on Windows but there should be a way to refresh the generated code for someone using another IDE and on another platform than Windows. It doesn't have to be as integrated an experience as T4 enjoys in VS so just a command-line alternative would do.

[Fold.g.tt]: https://github.com/morelinq/MoreLINQ/blob/812045328654177044e67eb93212efa3a7de20a8/MoreLinq/Fold.g.tt
[ToDelimitedString.g.tt]: https://github.com/morelinq/MoreLINQ/blob/812045328654177044e67eb93212efa3a7de20a8/MoreLinq/ToDelimitedString.g.tt

fsateler:
I found out today about Scripty: https://github.com/daveaglick/Scripty

atifaziz:
I am aware of Scripty but I have something else in mind for sake of compatibility: [T5](https://github.com/atifaziz/t5). It is my fork of [Mono.TextTemplating](https://www.nuget.org/packages/Mono.TextTemplating) that's been made to work on .NET Core.

If you are interested in some background, see [my comment](https://github.com/mono/t4/issues/8#issuecomment-363590843) in mono/t4#8.

atifaziz:
Closed via c418b1cb6aa3a6318179442374680e7269256f37.

PR #428, opened 2018-03-31
Cross-platform T4 template generation

atifaziz:
This PR addresses #381.

I don't think these should be invoked with every build so as not to slow down `build.cmd` and `build.sh` further. There should be a check during CI that the generated code is fresh, like was done in 8ffe07c060c7d8a51243a6a149c4ef8a43d366a3 for #235.

### DIFF
--- a/MoreLinq/tt.sh
+++ b/MoreLinq/tt.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+[[ -e tt.sh ]] || { echo >&2 "Please cd into the script location before running it."; exit 1; }
+set -e
+find . -name "*.tt" | xargs -t -n 1 dotnet tt
--- a/MoreLinq/MoreLinq.csproj
+++ b/MoreLinq/MoreLinq.csproj
@@ -185,6 +185,7 @@
 
   <ItemGroup>
     <Folder Include="Properties\" />
+    <DotNetCliToolReference Include="T5.TextTransform.Tool" Version="1.0.0" />
   </ItemGroup>
 
 </Project>
--- a/MoreLinq/tt.cmd
+++ b/MoreLinq/tt.cmd
@@ -1,0 +1,8 @@
+@echo off
+pushd "%~dp0"
+for /f %%f in ('dir /s /b *.tt') do (
+    echo>&2 dotnet tt %%f
+    dotnet tt %%f || goto :end
+)
+:end
+popd


MoreLinq/tt.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+[[ -e tt.sh ]] || { echo >&2 "Please cd into the script location before running it."; exit 1; }
+set -e
+find . -name "*.tt" | xargs -t -n 1 dotnet tt
```
fsateler: The canonical option is `-L 1` (`-n` has some special handling for the case the argument is larger than 1 but there are not enough arguments, and is not relevant for this particular case). I would use `-L` if only because it is more likely to be recognized already (I had to look up `-n`).

Also, on these pipelines, it is customary to pass `-print0` to `find` and `-0` to `xargs`, to prevent files with whitespace or other oddities to mess the `xargs` line parsing.

MoreLinq/tt.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+[[ -e tt.sh ]] || { echo >&2 "Please cd into the script location before running it."; exit 1; }
```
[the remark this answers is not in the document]
atifaziz: The reason I didn't go for `cd "$(dirname "$0")"` is to avoid the [so many ways you can skin this cat and still get it wrong by someone's book](https://stackoverflow.com/questions/59895/getting-the-source-directory-of-a-bash-script-from-within). I figured, leave the problem up to the caller when the invocation from the shell could be just `(cd MoreLinq; ./tt.sh)`. However, for all practical intent and purposes of this repo, I can imagine `cd "$(dirname "$0")"` is good enough and that we needn't worry ourselves about corner cases. But then this is something we should do for all `*.sh` shell scripts in the repo, don't you think?

MoreLinq/tt.sh
atifaziz: > Also, on these pipelines, it is customary to pass `-print0` to `find` and `-0` to `xargs`, to prevent files with whitespace or other oddities to mess the `xargs` line parsing.

Am aware but wanted to keep it consistent with the Windows script for now, as I'm not sure I designed that to work with whitespaces in names. I can test and revert.

atifaziz:
@fsateler You okay with my follow-up changes & [explanation](https://github.com/morelinq/MoreLINQ/pull/428#discussion_r178959394)? Hopefully it's a quick one to approve for merge. 😉

atifaziz:
> can fix that up in a later PR

Done too in 082382c74ceffd88a942652dbab9c62552994ceb.

PR #432, opened 2018-04-07
ToMultiDictionary mixin

wilversings:
I just think that this should belong here.


atifaziz:
How is this different from [`ToLookup`](https://docs.microsoft.com/en-gb/dotnet/api/system.linq.enumerable.tolookup) already present in LINQ?

wilversings:
@atifaziz Yes, you are right

atifaziz:
Thanks anyway for submitting what seems to be a good quality PR with unit tests and documentation. I wish you had opened an issue proposing this before doing all hard work.

PR #434, opened 2018-04-08
Updated release notes with CompareCount & Transpose

leandromoh:
reminding also that we changed signature of `MaxBy `and `MinBy`, and improved `Scan`.


atifaziz:
Thanks!

> reminding also that we changed signature of `MaxBy `and `MinBy`, and improved `Scan`.

I'm thinking that the release notes should just point to [the releases](https://github.com/morelinq/MoreLINQ/releases) where the full list of changes can be maintained.

leandromoh:
> I'm thinking that the release notes should just point to the releases where the full list of changes can be maintained.

Make sense. Will we need continue updating MoreLinq.csproj ?

atifaziz:
I don't know. I'm still thinking about this for the 3.0 release. You will know if it doesn't need updating.

Issue #405, opened 2018-03-04
new Acquire overload


leandromoh:
in a similar scenario that #404, sometimes some elements of a sequence can not be access, but we must do something with all avaible elements.

scenario :

```c#
foreach (var element in source) // if some element can not be reach, we must skip it
{ 
       // do something with element 
}
```
example of error :

```c#
Enumerable.Range(1, 10).Select(x =>
{
    if (x % 5 == 0)
        throw new InvalidOperationException();
    else
        return x;
}).ForEach(p => Console.WriteLine(p));
```

here we can not access elements 5 and 10, but the others can be printed.

So I propose a new operator that do this handling and return a sequence with all avaible elements.

Propose signature:

```c#
public static IEnumerable<T> Available<T>(this IEnumerable<T> source)
```

example of usage:

```c#
Enumerable.Range(1, 10).Select(x =>
{
    if (x % 5 == 0)
        throw new InvalidOperationException();
    else
        return x;
})
.Available()
.ForEach(Console.WriteLine);
```
now all elements except 5 and 10 are printed.

Draft: 

```c#
public static IEnumerable<T> Available<T>(this IEnumerable<T> source)
{
    if (source == null) throw new ArgumentNullException(nameof(source));

    var e = source.GetEnumerator();
    bool hasValue;
    bool dispose = false;

    while (true)
    {
        hasValue = true;
        try
        {
            if (!e.MoveNext())
            {
                e.Dispose();
                dispose = true;
            }
        }
        catch
        {
            hasValue = false;
        }

        if (dispose)
            yield break;

        if (hasValue)
            yield return e.Current;
    }
}
```

I would like to submit a PR for this. What do you think?

PS: the operator name is not defined yet;

atifaziz:
Generally speaking, I don't think exceptions should be dealt with exceptionally. They should be part of the query:

```c#
var xs =
    from x in Enumerable.Range(1, 10)
    select (Item: x, Error: x % 5 == 0 ? new InvalidOperationException() : null) into r
    where r.Error == null
    select r.Item;

q.ForEach(x => Console.WriteLine(x));
```

However, I can imagine that if the exception is thrown by some third-party code that's not under your control then such an operator can be useful.

I would call it `IgnoreErrors` but it should allow you to specify the specific errors to ignore, e.g.:

```c#
public static IEnumerable<T> IgnoreErrors<T, TException>(
    this IEnumerable<T> source,
[...]

PR #435, opened 2018-04-08
new Acquire permissive overload

leandromoh:
I would like to go ahead with the PR (add documentation and tests) if no one disagree the idea.

### DIFF
--- a/MoreLinq/Acquire.cs
+++ b/MoreLinq/Acquire.cs
@@ -19,6 +19,7 @@ namespace MoreLinq
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
 
     static partial class MoreEnumerable
     {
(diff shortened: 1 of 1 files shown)


leandromoh:
> But you closed #405.

My bad, I did it unintentionally. I reopened it now.

>  I can't commit time to review before we reach a reasonable consensus.

For sure. This PR is just sharing an experimenting of API design. Nothing to review currently.

> I'm guessing my proposal like IgnoreErrors hasn't connected with you

As you said [here](https://github.com/morelinq/MoreLINQ/issues/405#issuecomment-370417474): "The name does not match what it does; it's not just ignoring errors anymore".

> so just calling it XXXX means we're worse off than when the discussion started.

My bad. By XXXX I meant "whatever", that is, something not important that moment.

atifaziz:
@leandromoh Thanks for the clarifications.

PR #443, opened 2018-04-13
Refactored Batch implementation for clarity

leandromoh:


atifaziz:
I don't know what I am reviewing here because I don't know the basis for your justification that the code reads clearer now. If I just go on how I read it, it's worse. To remove subjectivity (what's readable to you versus what's readable to me), it's best lay down the _objectives_ of this refactoring.

PR #446, opened 2018-04-14
Added missing tests on PreScan

leandromoh:


MoreLinq.Test/PreScanTest.cs
```diff
@@ -17,11 +17,37 @@
...
+    using System;
     using NUnit.Framework;
 
     [TestFixture]
     public class PreScanTest
     {
+        [Test]
+        public void PreScanIsLazy()
+        {
+            new BreakingSequence<object>().PreScan(delegate { throw new NotImplementedException(); }, 0);
```
atifaziz: Use `BreakingFunc.Of<object, object, object>()` here since we already have it.

MoreLinq.Test/PreScanTest.cs
```diff
@@ -17,11 +17,37 @@
...
+
+            Assert.IsFalse(result.Any());
+        }
+
+        [Test]
+        public void PreScanWithSingleElement()
+        {
+            var source = new[] { 111 };
+            var result = source.PreScan(SampleData.Plus, 999);
+            var gold = new[] { 999 };
```
atifaziz: Inline `gold` such that next reads simply `result.AssertSequenceEqual(999);`.

MoreLinq.Test/PreScanTest.cs
```diff
@@ -22,6 +22,29 @@ namespace MoreLinq.Test
...
+            new BreakingSequence<int>().PreScan(BreakingFunc.Of<int, int, int>(), 0);
+        }
+
+        [Test]
+        public void PreScanWithEmptySequence()
+        {
+            var source = Enumerable.Empty<int>();
+            var result = source.PreScan(BreakingFunc.Of<int, int, int>(), 0);
+
+            Assert.IsFalse(result.Any());
```
atifaziz: This assertion is more clearly & explicitly expressed as: `Assert.That(result, Is.Empty);`

In case of failure, it will say something like this:

```
  Expected: <empty>
  But was:  < 2 >
```

instead of the following, which all `Assert.IsFalse` can say:

```
  Expected: False
  But was:  True
```

PR #447, opened 2018-04-14
PreScan iterates too eagerly

leandromoh:


MoreLinq/PreScan.cs
```diff
@@ -52,23 +52,32 @@ static partial class MoreEnumerable
         /// <param name="identity">Identity element (see remarks)</param>
         /// <returns>The scanned sequence</returns>
 
-        public static IEnumerable<TSource> PreScan<TSource>(this IEnumerable<TSource> source,
-            Func<TSource, TSource, TSource> transformation, TSource identity)
+        public static IEnumerable<TSource> PreScan<TSource>(this
+		    IEnumerable<TSource> source,
```
atifaziz: Move `this` from line above to here.

atifaziz: I think there's mixed use of tabs and spaces that's causing misalignment of code around here.

leandromoh: adjusts done.

PR #448, opened 2018-04-14
Optimize StartsWith false case for collections

leandromoh:


leandromoh:
@atifaziz adjusts done

PR #449, opened 2018-04-14
Optimize EndsWith false case for collections 

leandromoh:


MoreLinq/EndsWith.cs
```diff
@@ -71,8 +71,17 @@ public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second,
 
             comparer = comparer ?? EqualityComparer<T>.Default;
 
+            var nullableSecondCount = second.TryGetCollectionCount();
+
+            if (nullableSecondCount is int secondCount &&
```
atifaziz: This starting condition here is the same as the one on the `return` statement below so I don't think it has to be tested twice. What about collapsing it so?

```c#
            List<T> secondList;
            return second.TryGetCollectionCount() is int secondCount
                 ? first.TryGetCollectionCount() is int firstCount && secondCount > firstCount
                   ? false
                   : Impl(second, secondCount)
                 : Impl(secondList = second.ToList(), secondList.Count);
```

The benefit is that `nullableSecondCount` and `collectionCount` (which is the same as `secondCount`) are not needed.

leandromoh:
@atifaziz adjusts done

MoreLinq.Test/EndsWithTest.cs
```diff
@@ -92,11 +92,12 @@ public void EndsWithUsesSpecifiedEqualityComparerOrDefault()
...
 
-        [Test]
-        public void EndsWithUsesCollectionsCountToAvoidUnnecessaryIteration()
+        [TestCase(true)]
+        [TestCase(false)]
+        public void EndsWithUsesCollectionsCountToAvoidUnnecessaryIteration(bool readOnly)
         {
-            var first = new BreakingCollection<int>(1, 2);
-            var second = new BreakingCollection<int>(1, 2, 3);
+            var first = new[] { 1, 2 }.ToBreakingList(readOnly);
```
atifaziz: Use `ToBreakingCollection`, like you did in 1499147c92efbf92ef7c804bf4dbcc6756894448 for PR #448.

leandromoh: @atifaziz this fix was already done.

PR #450, opened 2018-04-14
Avoid XML escaping in doc code examples

leandromoh:


MoreLinq/Generate.cs
```diff
@@ -36,7 +36,7 @@ static partial class MoreEnumerable
         /// </remarks>
         /// <example>
         /// <code>
-        /// IEnumerable&lt;int&gt; result = Sequence.Generate(2, n => n * n).Take(5);
+        /// var result = MoreEnumerable.Generate(2, n => n * n).Take(5);
```
atifaziz: I fixed the name typo (`Sequence` &rarr; `MoreEnumerable`) in b43afc13f588f41dfb9e30182605a9fbbc75bbe2 as it's unrelated to this PR.

leandromoh:
> Thanks for this PR. While I'm fine with using var as a workaround, I think that the more robust way to solve this problem (while we are at it) would be to use a character data section (CDATA) for <code>

I see. I did not know about `CDATA`, good to know about it. 
In this case, even knowing about `CDATA`, I would like to keep using `var`, since `CDATA` is not necessary in this case. What do you think @atifaziz ?

atifaziz:
Sorry, [closed by mistake](https://github.com/morelinq/MoreLINQ/pull/450#event-1576892659).

leandromoh:
do you suggested merge the `CDATA` PR before this? in which summaries it would be added ?

atifaziz:
> do you suggested merge the `CDATA` PR before this?

I would say so because that's the proper solution to avoiding escaping. However, since that is what this PR intended to address, you could also revert all changes and make this the PR that adds `CDATA` instead. Then once merged, using `var`, which is a nice to have, can come in a second PR.

> in which summaries it would be added ? 

I would add it to all `<code>`.

leandromoh:
> I would add it to all <code>.

done.

leandromoh:
> Thanks for helping with this!

done

Issue #456, opened 2018-04-18
Backsert: insert a sequence from the back of another


leandromoh:
Similar to [Insert operator](https://github.com/morelinq/MoreLINQ/issues/306) (same signature) but the position where `second` will be inserted at `first` is count from end to start. That is, 

- 0 means last position, 
- 1 means second-last position, 
- 2 means third-last position
- etc

## Signature:
```c#
IEnumerable<T> InsertEnd<T>(this IEnumerable<T> first, IEnumerable<T> second, int index)
```

## Example:

```c#
[1,2].InsertEnd([3,4], -1); // ArgumentOutOfRangeException when index < 0
[1,2].InsertEnd([3,4], 0); // [1,2,3,4]
[1,2].InsertEnd([3,4], 1); // [1,3,4,2]
[1,2].InsertEnd([3,4], 2); // [3,4,1,2]
[1,2].InsertEnd([3,4], 3); // ArgumentOutOfRangeException when index > first.Count()
```

## workaround

```c#
first.SkipLast(index).Concat(second).Concat(first.TakeLast(index))
```

produces the same as 

```c#
first.InsertEnd(second, index)
```

except for exception cases with `index` parameter

## Motivation

1. `InsertEnd` iterates `first` once, 
2. performance, since workaround iterates `first` 2 times and uses 2 buffers (`SkipLast` + `TakeLast`)
3. `Insert` operator symmetry
4. more readable, workaround does not capture the intention as clearly as `InsertEnd`

atifaziz:
Interesting. 💭 Let me propose something a little more general first because I am seeing a pattern that's common among this and several other operators that we already have.

atifaziz:
> Let me propose something a little more general first

@leandromoh See #458. Here's an implementation of `InsertEnd` in terms of `CountDown`:

```c#
public static IEnumerable<T> InsertEnd<T>(this IEnumerable<T> first,
    int count, IEnumerable<T> second) =>
    from seq in
        first.Index()
             .Select(e => (Index: e.Key, Element: Enumerable.Repeat(e.Value, 1)))
             .CountDown(count + 1, (e, cd) => 
                 cd < count && e.Index == 0
                 ? second.Concat(e.Element)
                 : e.Element.Concat(cd == count ? second : Enumerable.Empty<T>()))
    from e in seq
    select e;
```

Thoughts?

atifaziz:
> I really can not understand your implementation =(

Annotated version:

```c#
[...]

PR #460, opened 2018-04-19
Backsert implementation (issue #456)

leandromoh:
just sharing an experimenting


atifaziz:
I see you have `CountDown` embedded. 😆

atifaziz:
📝 This is for #460.

leandromoh:
@atifaziz I adjust the documentation. IMHO this PR is ready for merge. Some more in your opinion?

leandromoh:
@atifaziz I adjusted the documentation, renamed operator to `Backsert` and updated README. 
Something before merge in your opinion?

leandromoh:
> All tests should use a TestingSequence to make sure Backsert disposes and doesn't iterate the source sequence more than once.

done

MoreLinq/Backsert.cs
```diff
@@ -0,0 +1,77 @@
...
+            if (second == null) throw new ArgumentNullException(nameof(second));
+            if (index < 0)  throw new ArgumentOutOfRangeException(nameof(index), "Index cannot be negative.");
+
+            return index == 0
+                   ? first.Concat(second)
+                   : _();
+
+            IEnumerable<T> _() =>
+                 first.CountDown(index, (e, cd) => (Element: e, Countdown: cd))
+                      .SelectMany((e, i) => i == 0
```
atifaziz: This `SelectMany` use should be rolled-out by hand otherwise this version (which was fine as a prototype) will cause a heap allocation for _each_ item (`Enumerable.Repeat(e.Element, 1)`) of the first sequence!

leandromoh: done!

atifaziz: I made some improvements in b9643c3e517ca922fbcbb87a92754a7fab28e634 to simplify the tuple usage that were easier & faster to do than describe.

MoreLinq.Test/BacksertTest.cs
```diff
@@ -0,0 +1,105 @@
...
+        public void Backsert(int count, int index)
+        {
+            var seq1 = Enumerable.Range(1, count);
+            var seq2 = new[] { 97, 98, 99 };
+
+            using (var test1 = seq1.AsTestingSequence())
+            using (var test2 = seq2.AsTestingSequence())
+            {
+                var result = test1.Backsert(test2, index);
+                var expectations = seq1.SkipLast(index).Concat(seq2).Concat(seq1.TakeLast(index));
```
atifaziz: I think this test (and others) rely too much on algorithmic expression of expectations that's hard to follow. It makes reviewing the test hard, that it's testing the right thing in the first place without doing the mental gymnastics. In other words, a literal expression is much clearer as to the inputs and what's going to happen:

```c#
[TestCase(new[] { 1, 2, 3 }, 1, new[] { 97, 98, 99 }, ExpectedResult = new[] { 1, 2, 97, 98, 99, 3 })]
[TestCase(new[] { 1, 2, 3 }, 2, new[] { 97, 98, 99 }, ExpectedResult = new[] { 1, 97, 98, 99, 2, 3 })]
public IEnumerable<int> Backsert(int[] seq1, int index, int[] seq2)
{
    using (var test1 = seq1.AsTestingSequence())
    using (var test2 = seq2.AsTestingSequence())
    {
        foreach (var item in test1.Backsert(test2, index))
            yield return item;
    }
}
```

leandromoh: I removed all "mental gymnastic"

leandromoh:
@atifaziz is it ready for merge?

MoreLinq.Test/BacksertTest.cs
atifaziz: > I removed all "mental gymnastic"

I'm afraid it's still there unless I'm missing something? For example:

```c#
var expectations = seq1.SkipLast(index).Concat(seq2).Concat(seq1.TakeLast(index));
```

Why not use `ExpectedResult` on `[TestCase]` instead? If you do, you can fold `BacksertWithIndexEqualsSourceLength` and `BacksertWithIndexZero` into the `Backset` test.

leandromoh: I know, but I prefer keep different tests for each scenario. I think explict tests improves clarity of what is being tested.

atifaziz: >  I think explict tests improves clarity of what is being tested.

It's not, unfortunately, which is why I am asking for the changes. I am still doing “mental gymnastics” to do the review so it's not helping. That means when I have 5 minutes and I try to make progress on a review for a _simple operator_ and 5 minutes become 15 and then I have to abort; that translates into the review staying open longer.

We have 3 tests with duplicate code that needs maintenance over time. The only explicit or unique part about each test is its name. If you wish to make a note of that, then I'd comment on top of the `[TestCase]` or even use its `Description` property.

leandromoh: I see. So basically "less is more". Okay. Adjust done.

MoreLinq.Test/BacksertTest.cs
```diff
@@ -0,0 +1,66 @@
...
+        [TestCase(new[] { 1, 2, 3 }, 0, new[] { 8, 9 }, ExpectedResult = new[] { 1, 2, 3, 8, 9 })]
+        [TestCase(new[] { 1, 2, 3 }, 1, new[] { 8, 9 }, ExpectedResult = new[] { 1, 2, 8, 9, 3 })]
+        [TestCase(new[] { 1, 2, 3 }, 2, new[] { 8, 9 }, ExpectedResult = new[] { 1, 8, 9, 2, 3 })]
+        [TestCase(new[] { 1, 2, 3 }, 3, new[] { 8, 9 }, ExpectedResult = new[] { 8, 9, 1, 2, 3 })]
+        public IEnumerable<int> Backsert(int[] seq1, int index, int[] seq2)
+        {
+            using (var test1 = seq1.AsTestingSequence())
+            using (var test2 = seq2.AsTestingSequence())
+            {
+                foreach (var item in test1.Backsert(test2, index))
```
atifaziz: I think this loop can be replaced by simply a call to `ToArray`:

```c#
using (var test1 = seq1.AsTestingSequence())
using (var test2 = seq2.AsTestingSequence())
    test1.Backsert(test2, index).ToArray();
```

I am not sure there's anything to be gained by yielding the results in a streamed fashion.

leandromoh: done.

PR #465, opened 2018-04-25
MoveSource in MoveTest bug

leandromoh:
[Method](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq.Test/MoveTest.cs#L69) `MoveSource` in `MoveTest.cs` always returns `fromIndex` equals `toIndex`, that is the [optmized case](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/Move.cs#L57) (test was created before optmization), therefore did not tests the expected scenario.

### DIFF
--- a/MoreLinq.Test/MoveTest.cs
+++ b/MoreLinq.Test/MoveTest.cs
@@ -17,6 +17,7 @@
 
 namespace MoreLinq.Test
 {
+    using System;
     using System.Collections.Generic;
     using NUnit.Framework;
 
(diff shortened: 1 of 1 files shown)


MoreLinq.Test/MoveTest.cs
```diff
@@ -70,9 +71,14 @@ public static IEnumerable<object> MoveSource()
         {
             const int length = 10;
 
-            return from index in Enumerable.Range(0, length)
-                   from count in Enumerable.Range(0, length + 1)
-                   select new TestCaseData(length, index, count, index);
+            foreach (var index in Enumerable.Range(0, length))
```
atifaziz: You can keep this as a query (instead of turning it into an iterator):

```c#
return from index in Enumerable.Range(0, length)
       from count in Enumerable.Range(0, length + 1)
       from tcd in new[]
       {
           new TestCaseData(length, index, count, Math.Max(0, index - 1)),
           new TestCaseData(length, index, count, index + 1),
       }
       select tcd;
```

This is more powerful because (in a later refactoring, after this PR) you can refactor the test data into a field:

```c#
public static IEnumerable<TestCaseData> MoveSource =>
    from length in new[] { 10 }
    from index in Enumerable.Range(0, length)
    from count in Enumerable.Range(0, length + 1)
    from tcd in new[]
    {
        new TestCaseData(length, index, count, Math.Max(0, index - 1)),
        new TestCaseData(length, index, count, index + 1),
    }
    select tcd;
```

leandromoh:
@atifaziz kept source as a query.

PR #467, opened 2018-04-25
Reuse CountDown on TakeLast

leandromoh:


MoreLinq/TakeLast.cs
```diff
@@ -52,23 +53,9 @@ public static IEnumerable<TSource> TakeLast<TSource>(this IEnumerable<TSource> s
...
-                    {
-                        if (q.Count == count)
-                            q.Dequeue();
-                        q.Enqueue(item);
-                    }
-
-                    foreach (var item in q)
-                        yield return item;
-                }
+                : source.CountDown(count, (x, cd) => (Element: x, CountDown: cd ))
```
atifaziz: The element name for the second tuple item should be `Countdown` instead of `CountDown`. In English, _countdown_ is a noun for the counter but _count-down_ is the verb for or action of counting-down. You use Pascal-casing of `CountDown` due to the hyphenation; devil in the details. :wink:

leandromoh:
> It degrades in allocation (and therefore GC pressure & performance)

I think the main point about performance is kept: iterates `source` once and uses a single buffer

> second tuple item should be Countdown instead of CountDown

done

PR #468, opened 2018-04-25
TakeLast optmization for when count <= 0

leandromoh:
As well we have for [SkipLast](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/SkipLast.cs#L40).


leandromoh:
> I think this needs re-basing as TakeLast implementation got completely refactored in dc89b81?

done

Issue #191, opened 2016-09-01
Force enumerable


dzmitry-lahoda:
One time I created `ThreadStatic` context for creating enumerable. I applied it to entities for lazy collection of these. So this context was lost in thread which popup lower in stack. I could change layers beneath which start new threads to do stuff, but too much work. It is easier to force evaluation. If it is list or array already - no need for evaluation. I cannot use empty `ForEach` as is to costly to invoke delegate. I could use `Count()` but need to enforce it with some code and ensure compiled will not throw this code away. So expecting method like `Force()` which returns `IEnumerable` but which is sure have been counted.

See related http://stackoverflow.com/questions/314100/something-better-than-toarray-to-force-enumeration-of-linq-output

asd-and-Rizzo:
Like next:
```
       /// <summary>
        /// Forces the specified enumeration to be evaluated, if it is not yet.
        /// </summary>
        /// <param name="self">The self.</param>
        /// <returns>Same enumeration.</returns>
        [System.Runtime.CompilerServices.MethodImpl(System.Runtime.CompilerServices.MethodImplOptions.NoInlining | System.Runtime.CompilerServices.MethodImplOptions.NoOptimization)]
        public static IEnumerable<T> Force<T>(this IEnumerable<T> self)
        {
            var count = self.Count();
            return self;
        }
```

asd-and-Rizzo:
Seems taking count and use enumration again will evaluate delegates again. So next:
```

        /// <summary>
        /// Forces the specified enumeration to be evaluated, if it is not yet. On secon pass no evaliation will happen.
        /// </summary>
        /// <param name="self">The self.</param>
        /// <returns>Same enumeration.</returns>
        [System.Runtime.CompilerServices.MethodImpl(System.Runtime.CompilerServices.MethodImplOptions.NoInlining | System.Runtime.CompilerServices.MethodImplOptions.NoOptimization)]
        public static IEnumerable<T> Force<T>(this IEnumerable<T> self)
        {
            if (self is ICollection<T> || self is T[] || self is IReadOnlyCollection<T>)
            {
                return self;
            }
            else
            {
                return self.ToArray();
            }            
        }
```

JamesFaix:
I believe this is the same idea as the existing `Consume` operator.

```
/// <summary>
/// Completely consumes the given sequence. This method uses immediate execution,
/// and doesn't store any data during execution.
/// </summary>
[...]

PR #469, opened 2018-04-25
Materialise RankBy source to list if not one already

leandromoh:


leandromoh:
I think this looks a good opportunity to add [Force](https://github.com/morelinq/MoreLINQ/issues/191) operator to library and use it here.

atifaziz:
> I think this looks a good opportunity to add a MVP of [Force](https://github.com/morelinq/MoreLINQ/issues/191) operator to library and use it here.
I would be happy in submit a PR with my [proposed implemetation](https://github.com/morelinq/MoreLINQ/issues/191#issuecomment-377742192).

Small steps first. 🐾 We can add it but it would have to be internal. We are, or perhaps I should say I am, not ready to commit to a public `Force` with the proposed implementation.

leandromoh:
> Small steps first. feet We can add it but it would have to be internal.

I think it is more worth create a public `Force` first, then refactoring `RankBy` and possible others operators.

> We are, or perhaps I should say I am, not ready to commit to a public Force with the proposed implementation.

We can discuss about a public `Force` implementation on the issue #191 . I would enjoy your comments there.

PR #472, opened 2018-04-27
List-like union as struct to save allocation + virtual dispatch

atifaziz:
This PR converts the [list-like](https://github.com/morelinq/MoreLINQ/blob/7eede03c6a8cdb6ca32ca8d68f3bf866fd552ef8/MoreLinq/ListLike.cs) union via a class hierarchy to a `struct` for the purpose of saving GC allocations and additional virtual hops. I don't expect there will ever be more than one allocation saved per list and per invocation of an operator so the savings there are really pathetic. The implementation also feels a bit clumsy but there is not much choice there.


MoreLinq/ListLike.cs
```diff
@@ -21,34 +21,35 @@ namespace MoreLinq
...
 
     /// <summary>
-    /// Represents an list-like (indexable) data structure.
+    /// Represents a union over list types implementing either
+    /// <see cref="IList{T}"/> or <see cref="IReadOnlyList{T}"/>, allowing
+    /// both to be treated the same.
     /// </summary>
 
-    interface IListLike<out T>
+    struct ListLike<T> // TODO Can be readonly when using C# 7.2
```
fsateler: I believe this TODO is now actionable as 7.3 is the minimum version https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

MoreLinq/ListLike.cs
```diff
@@ -21,34 +21,35 @@ namespace MoreLinq
...
     {
-        int Count { get; }
-        T this[int index] { get; }
-    }
+        readonly IList<T> _l;
+        readonly IReadOnlyList<T> _r;
```
fsateler: These member names are quite cryptic.

MoreLinq/ListLike.cs
```diff
@@ -21,34 +21,35 @@ namespace MoreLinq
     using System.Collections.Generic;
 
     /// <summary>
-    /// Represents an list-like (indexable) data structure.
+    /// Represents a union over list types implementing either
```
fsateler: Instead of a union, what about a delegating tuple?

```csharp
static class ListLike 
{
  public static (Func<int> GetCount, Func<int, T> GetItem) => (() => list.Count, (i) => list[i]);
  public static (Func<int> GetCount, Func<int, T> GetItem) => (() => list.Count, (i) => list[i]);
}
```

MoreLinq/ListLike.cs
atifaziz: Addressed with 659a56b40061de6496260b6f17062d1ec3e2a160.

MoreLinq/ListLike.cs
atifaziz: Tried to address this with 064fb6151586c2e467920662678c1dc8e0231f47. Perhaps `rw` and `rx` are still cryptic, but wanted to keep them the same length (trying to hard?) so these are *nix inspired (`rw` = read-write, `rx` = read-only).

MoreLinq/ListLike.cs
atifaziz: I don't this is syntactically correct, but I think I get your idea nonetheless. This would cause two closures and indirect calls so have my doubts it's as optimal.

atifaziz:
> I believe the performance benefit has not been demonstrated to offset the more awkward code.

Benchmark code:

```c#
[ParamsSource(nameof(Sources))]
public IEnumerable<int> Source { get; set; }

public IEnumerable<IEnumerable<int>> Sources
{
    get
    {
        var xs = Enumerable.Range(1, 1_000_000);
        yield return xs;
        yield return xs.ToArray();
    }
}

[Benchmark]
public void CountDown()
{
    Source.CountDown(1, (_, _) => 0).Consume();
}
```

Benchmarks for `CountDown` using [last release (3.3.2)](https://github.com/morelinq/MoreLINQ/releases/tag/v3.3.2):

    BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22621.1105)
    Intel Core i7-1065G7 CPU 1.30GHz, 1 CPU, 8 logical and 4 physical cores
    .NET SDK=7.0.102
      [Host] : .NET 7.0.2 (7.0.222.60605), X86 RyuJIT AVX2

|    Method |               Source |     Mean |    Error |   StdDev | Allocated |
|---------- |--------------------- |---------:|---------:|---------:|----------:|
| CountDown |       Int32[1000000] | 22.20 ms | 0.507 ms | 1.488 ms |     119 B |
| CountDown | Syste(...)rator [36] | 20.09 ms | 0.479 ms | 1.413 ms |     179 B |

Notice that the run with an array as a source is slower.

Following is the benchmark for changes in this PR up to 064fb6151586c2e467920662678c1dc8e0231f47:

    BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22621.1105)
    Intel Core i7-1065G7 CPU 1.30GHz, 1 CPU, 8 logical and 4 physical cores
    .NET SDK=7.0.102
      [Host] : .NET 7.0.2 (7.0.222.60605), X86 RyuJIT AVX2

|    Method |               Source |     Mean |    Error |   StdDev |   Median | Allocated |
|---------- |--------------------- |---------:|---------:|---------:|---------:|----------:|
| CountDown |       Int32[1000000] | 18.76 ms | 0.455 ms | 1.334 ms | 18.45 ms |     115 B |
| CountDown | Syste(...)rator [36] | 19.77 ms | 0.477 ms | 1.383 ms | 19.20 ms |     179 B |

The run with an array as a source is slightly faster.

---
- 📎 [`benchmark-3.3.2.log.txt`](https://github.com/morelinq/MoreLINQ/files/10475674/benchmark-3.3.2.log.txt)
- 📎 [`benchmark-pr-064fb615.log.txt`](https://github.com/morelinq/MoreLINQ/files/10475675/benchmark-pr-064fb615.log.txt)

viceroypenguin:
@atifaziz If you would, please compare your latest results to SuperLinq 4.6.0? I did a comparison against latest public release (3.3.2), and I show that SuperLinq is 40% faster on array and essentially equivalent (4% faster) on enumerables. 

For `CountDown` specifically, it appears to be faster to use the corelib `Reverse()` operator than to do any allocation of our own. 

```
// * Summary *

BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22623.1180)
Intel Core i7-1065G7 CPU 1.30GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK=7.0.102
  [Host] : .NET 7.0.2 (7.0.222.60605), X64 RyuJIT AVX2


|             Method |               Source |     Mean |    Error |   StdDev | Allocated |
|------------------- |--------------------- |---------:|---------:|---------:|----------:|
| SuperLinqCountDown |       Int32[1000000] | 10.30 ms | 0.187 ms | 0.166 ms |     138 B |
|  MoreLinqCountDown |       Int32[1000000] | 17.24 ms | 0.305 ms | 0.339 ms |     186 B |
| SuperLinqCountDown | Syste(...)rator [36] | 13.02 ms | 0.250 ms | 0.234 ms |     229 B |
|  MoreLinqCountDown | Syste(...)rator [36] | 13.58 ms | 0.240 ms | 0.224 ms |     237 B |
```

viceroypenguin:
I had already removed `ListLike<>` from SuperLinq code, which allows for the possibility of comparison. I did a further comparison for all three methods (AggregateRight, CountDown, ScanRight). I did find a [performance improvement](https://github.com/viceroypenguin/SuperLinq/pull/144) for SuperLinq in `ScanRight`; after fixing `ScanRight`, I found that in only one case does the `ListLike<>` behavior improve performance:

```
// * Summary *

BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22623.1180)
Intel Core i7-1065G7 CPU 1.30GHz, 1 CPU, 8 logical and 4 physical cores
.NET SDK=7.0.102
  [Host] : .NET 7.0.2 (7.0.222.60605), X64 RyuJIT AVX2
```

|                  Method |               Source |      Mean |     Error |    StdDev |     Gen0 |     Gen1 |     Gen2 | Allocated |
|------------------------ |--------------------- |----------:|----------:|----------:|---------:|---------:|---------:|----------:|
|      SuperLinqCountDown |       Int32[1000000] | 10.055 ms | 0.1166 ms | 0.1090 ms |        - |        - |        - |     138 B |
|       MoreLinqCountDown |       Int32[1000000] | 16.673 ms | 0.2415 ms | 0.2259 ms |        - |        - |        - |     186 B |
|      SuperLinqScanRight |       Int32[1000000] | 12.884 ms | 0.1779 ms | 0.1577 ms | 281.2500 | 281.2500 | 281.2500 | 4000642 B |
|       MoreLinqScanRight |       Int32[1000000] | 13.821 ms | 0.1599 ms | 0.1335 ms | 375.0000 | 375.0000 | 375.0000 | 4000912 B |
| SuperLinqAggregateRight |       Int32[1000000] |  6.466 ms | 0.0895 ms | 0.0837 ms | 265.6250 | 265.6250 | 265.6250 | 4000500 B |
|  MoreLinqAggregateRight |       Int32[1000000] |  5.326 ms | 0.0596 ms | 0.0557 ms |        - |        - |        - |      38 B |
|      SuperLinqCountDown | Syste(...)rator [36] | 10.958 ms | 0.2174 ms | 0.2232 ms |        - |        - |        - |     149 B |
|       MoreLinqCountDown | Syste(...)rator [36] | 13.263 ms | 0.2488 ms | 0.2206 ms |        - |        - |        - |     237 B |
|      SuperLinqScanRight | Syste(...)rator [36] | 15.848 ms | 0.3051 ms | 0.3858 ms | 375.0000 | 375.0000 | 375.0000 | 8000878 B |
|       MoreLinqScanRight | Syste(...)rator [36] | 16.365 ms | 0.3096 ms | 0.3180 ms | 343.7500 | 343.7500 | 343.7500 | 8000949 B |
| SuperLinqAggregateRight | Syste(...)rator [36] |  6.637 ms | 0.0931 ms | 0.0871 ms | 289.0625 | 289.0625 | 289.0625 | 4000552 B |
|  MoreLinqAggregateRight | Syste(...)rator [36] |  7.955 ms | 0.1582 ms | 0.1693 ms | 328.1250 | 328.1250 | 328.1250 | 4000636 B |

Data: https://gist.github.com/viceroypenguin/ec71e7adb8124c455e8a2925bee1b987

atifaziz:
> @atifaziz If you would, please compare your latest results to [SuperLinq 4.6.0](https://github.com/viceroypenguin/SuperLinq/blob/master/Source/SuperLinq/CountDown.cs)? I did a comparison against latest public release (3.3.2), and I show that SuperLinq is 40% faster on array and essentially equivalent (4% faster) on enumerables.
> 
> For `CountDown` specifically, it appears to be faster to use the collection foreach rather than use the list-like structure.

I shared those benchmarks to get back to [the comment](https://github.com/morelinq/MoreLINQ/pull/472#pullrequestreview-347043382) from @fsateler, where he rightly challenged that the performance & allocation benefits were not demonstrated (trade-off being code legibility). So this PR is not about `CountDown`, but since it uses `ListLike<T>` and scans the entire source, I wanted to use it as the basis of the demonstration. The (separate) problem with `CountDown` is that its loop is poorly written with respect to performance. It accesses the list `Count` property 3 times and twice during each iteration of the loop. Each access is a virtual dispatch so it probably adds up. If the list count is taken into a local variable like so:

```diff
diff --git a/MoreLinq/CountDown.cs b/MoreLinq/CountDown.cs
index f7b8ade1..16797781 100644
--- a/MoreLinq/CountDown.cs
+++ b/MoreLinq/CountDown.cs
@@ -65,15 +65,11 @@ static partial class MoreEnumerable
 
             IEnumerable<TResult> IterateList(ListLike<T> list)
             {
-                var countdown = Math.Min(count, list.Count);
+                var count = list.Count;
+                var countdown = Math.Min(count, count);
 
-                for (var i = 0; i < list.Count; i++)
-                {
-                    var cd = list.Count - i <= count
-                           ? --countdown
-                           : (int?) null;
-                    yield return resultSelector(list[i], cd);
-                }
+                for (var i = 0; i < count; i++)
+                    yield return resultSelector(list[i], count - i <= count ? --countdown : null);
             }
 
             IEnumerable<TResult> IterateCollection(int i)
```

then it makes a stark improvement:

|             Method |               Source |     Mean |    Error |   StdDev | Allocated |
|------------------- |--------------------- |---------:|---------:|---------:|----------:|
|  MoreLinqCountDown |       Int32[1000000] | 11.30 ms | 0.244 ms | 0.712 ms |      96 B |
| SuperLinqCountDown |       Int32[1000000] | 13.24 ms | 0.288 ms | 0.849 ms |      93 B |
|  MoreLinqCountDown | Syste(...)rator [36] | 19.77 ms | 0.491 ms | 1.433 ms |     179 B |
| SuperLinqCountDown | Syste(...)rator [36] | 18.70 ms | 0.537 ms | 1.582 ms |     179 B |

In fact, it's almost 15% faster than SuperLinq (which I'm guessing is just [iterating the list via the enumerator](https://github.com/viceroypenguin/SuperLinq/blob/4b872270b52ea8f7642ef1b75c04d8f61b7a8fdd/Source/SuperLinq/CountDown.cs#L84-L94)).

> For `CountDown` specifically, it appears to be faster to use the collection foreach rather than use the list-like structure.

So perhaps not? I haven't investigated in detail as to why, but if I were to take a somewhat educated guess, then I'd say iterating a sequence makes two virtual dispatches per loop iteration (`MoveNext` + `Current`) whereas with a list, there's just one, the indexer. Moreover, even if `ListLike<T>` indexer is coded for maximum probability of getting in-lined, it makes no discernible difference.

---
- 📎[`UserQuery-20230123-003056.log.txt`](https://github.com/morelinq/MoreLINQ/files/10475850/UserQuery-20230123-003056.log.txt)

atifaziz (commented):
@fsateler I'm going to assume you don't have the time to review, so I will go ahead and merge this since there's been at least another pair of eyes (@viceroypenguin) over it. If there's any strong objection (at some future point) to some aspect or the whole refactoring then feel free to open a follow-up issue. Thanks!

fsateler (commented):
I had the review but I forgot to actually submit it :(

Nothing critical but here it is anyway.

MoreLinq/ListLike.cs
```diff
@@ -22,43 +22,51 @@ namespace MoreLinq
...
+            _rx = null;
+        }
+
+        public ListLike(IReadOnlyList<T> list)
+        {
+            _rw = null;
+            _rx = list ?? throw new ArgumentNullException(nameof(list));
+        }
+
+        public int Count => _rw?.Count ?? _rx?.Count ?? 0;
```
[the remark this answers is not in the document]
atifaziz: If a `ListLike`, being a `struct`, is instantiated in its default state then it's treated like an empty list, which is why the 0 is there. The indexer that throws instead is not any different, but there an `index` for an empty like would an invalid argument.

MoreLinq/ListLike.cs
```diff
@@ -22,43 +22,51 @@ namespace MoreLinq
...
+    /// both to be treated the same.
     /// </summary>
 
-    interface IListLike<out T>
+    readonly struct ListLike<T>
     {
-        int Count { get; }
-        T this[int index] { get; }
+        readonly IList<T>? _rw;
+        readonly IReadOnlyList<T>? _rx;
```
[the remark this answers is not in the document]
atifaziz: I realize that, but I didn't want to use `_r_` with the second `_` meaning a `-`. The `x` was more to say _not_. Anyway, I've renamed to `_ro` with PR #948.

Thanks for the review!

PR #475, opened 2018-04-27
Add `UsingTestingSequence` to enforce simpler & correct usage

atifaziz:
This PR obsoletes the use of `AsTestingSequence` in favour of the new `UsingTestingSequence`. This is done to enforces simpler & more correct usage.

`TestingSequence` is primarily designed to help test two things in an operator:

1. It doesn't enumerate the sequence more than once
2. It disposes the sequence

However, most uses seemed to be exercising the first but not the second. The latter was probably often skipped/missed because it required more non-obvious setup; an explicit `using` block (non-obvious usage) as well as eager evaluation before assertion of results. Take the following example:

```c#
var result = sequence.AsTestingSequence().Rank();
```

This will only ensure that `Rank` doesn't iterate `sequence` more than once when it's iterated during testing. To also test that it disposes the source iterator when done, the above has to be written as:

```c#
IEnumerablt<int> result;
using (var ts = sequence.AsTestingSequence())
    result = ts.Rank().ToArray();
```

With `UsingTestingSequence`, you just do:

```c#
var result = sequence.UsingTestingSequence(ts => ts.Rank());
```

## `TestingSequence` Uses Needing Review

- [ ] `CompareCountTest`
- [x] `CountDownTest`
- [x] `EndsWithTest`
- [ ] `EquiZipTest`
- [ ] `FlattenTest`
- [x] `FoldTest`
- [ ] `InterleaveTest`
- [ ] `MoveTest`
- [ ] `NestedLoopTest`
- [x] `RankTest`
- [ ] `SortedMergeTest`
- [ ] `StartsWithTest`
- [ ] `TakeLastTest`
- [ ] `TestingSequence`
- [ ] `TransposeTest`
- [ ] `ZipLongestTest`
- [ ] `ZipShortestTest`


leandromoh:
I think all the explanation and operator list should be in a issue. Só que can create PR for each operator.

leandromoh (approved):

PR #476, opened 2018-04-27
Optimize ScanRight for IReadOnlyList #128

Arithmomaniac:


atifaziz:
:memo: This PR is related #128.

MoreLinq.Test/ScanRightTest.cs
```diff
@@ -42,11 +42,14 @@ public void ScanRightFuncIsNotInvokedOnSingleElementSequence()
             Assert.That(result, Is.EqualTo(new[] { value }));
         }
 
-        [Test]
-        public void ScanRight()
+        [TestCase(SourceKind.List)]
```
atifaziz: I don't think there is a way to prove that this test case can be broken? `ScanRight` internally skips materialisation to a list if the source is already one. Ideally, if that optimisation is removed then the test should break but detecting whether a copy was made or not is brittle. I'd hate to rely on current internal implementation details like it eventually ends up calling [`ICollection<>`'s `CopyTo`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icollection-1.copyto?), which [our `BreakingCollection<>` implements](https://github.com/morelinq/MoreLINQ/blob/bab96ace806efba59c444f8bf31faa36b6aefe57/MoreLinq.Test/BreakingCollection.cs#L45) (and why the test doesn't break).

Arithmomaniac: Something really hacky we could do is add a `ThrowOnCopy` property to `Breaking(ReadOnly)Collection` (which would toggle `CopyTo`'s behavior), and then do something like this:

```
var enumerable = Enumerable.Range(1, 5).Select(x => x.ToString()).ToSourceKind(sourceKind);
switch (enumerable)
{
case BreakingReadOnlyCollection readOnlyCollection: 
    readOnlyCollection.ThrowOnCopy = true; break;
case BreakingCollection collection:
    collection.ThrowOnCopy = true; break;
}
```

atifaziz: As I said, I'd hate to rely on current internal implementation details that can change (and [already have I believe](https://github.com/dotnet/corefx/pull/3429)) so I think anything we'll come up with will seem like a hack at best. I'm not sure it's worth the trouble. The best option might be to comment out that test case and make a note of why.

atifaziz:
f856d0df72ad2ca8d48145c4039f11e481ecad77 fixes the [broken](https://ci.appveyor.com/project/raboof/morelinq/build/921) [builds](https://travis-ci.org/morelinq/MoreLINQ/builds/374549423?utm_source=github_status&utm_medium=notification) due to a trailing whitespace in 36fafac8607d3ce7f48773cea9b5a740d2d5328b.

MoreLinq.Test/ScanRightTest.cs
```diff
@@ -42,11 +42,15 @@ public void ScanRightFuncIsNotInvokedOnSingleElementSequence()
             Assert.That(result, Is.EqualTo(new[] { value }));
         }
 
-        [Test]
-        public void ScanRight()
+        // Cannot test that ListLike optimizations would be used, so
```
atifaziz: Actually the `ReadOnlyList` test case _is_ okay because it does break if the optimisation within `ScanRight` is removed. I had [called out the comment](https://github.com/morelinq/MoreLINQ/pull/476#discussion_r185191063) on the `List` case only [earlier](https://github.com/morelinq/MoreLINQ/pull/476#discussion_r185191063) but perhaps should have made that more clear.

I would leave also leave the test case in there and either comment it out:

```c#
// [TestCase(SourceKind.List)]
```

or mark it to be ignored:

```c#
[TestCase(SourceKind.List, IgnoreReason = "Can't be broken; see https://github.com/morelinq/MoreLINQ/pull/476#discussion_r185191063")]
```

I'd also like to see the comment fleshed out a bit more. Here's a suggestion:

```c#
//
// There isn't a way to prove that the List test case can be broken
// so that case is ignored below. ScanRight internally skips
// materialisation to a list if the source is already one. Ideally,
// if that optimisation is removed then the test would break but
// detecting that a copy was made or not is tricky, hack-ish and
// brittle at best. It would mean relying on current and internal
// implementation details of Enumerable.ToList that can and have
// changed.
//
[TestCase(SourceKind.List, IgnoreReason = "Can't be broken; see https://github.com/morelinq/MoreLINQ/pull/476#discussion_r185191063")]
```

Happy if you want to review and fix my English and rephrase parts if they read unclear.

Arithmomaniac: > Actually the ReadOnlyList test case is okay because it does break if the optimisation within ScanRight is removed

This test passing also relies on the internal implementation details of `ToList` (namely, that it's not optimized for IReadOnlyList); it could become an unbreakable test at any time. I'd be more comfortable commenting both out and saying they are both unreliable because we can't force a check on `ToList` being skipped. I'd rather mark both as ignored.

atifaziz: [Does it?](https://github.com/Arithmomaniac/MoreLINQ/blob/f856d0df72ad2ca8d48145c4039f11e481ecad77/MoreLinq/ScanRight.cs#L93-L95) Isn't this relying on the following optimization in `ScanRight`?

```c#
            var list = source is IReadOnlyList<TSource> readOnlyList
                     ? readOnlyList.AsListLike()
                     : (source as IList<TSource> ?? source.ToList()).AsListLike();
```

Point is, if I remove that optimization so that it's down to this:

```c#
            var list = (source as IList<TSource> ?? source.ToList()).AsListLike();
```

Then the test fails. The same cannot be said for removing the above optimization too.

Had these tests been developed using [the red-green-refactor approach from TDD](https://www.google.ch/search?q=tdd+red+green+refactor) where a breaking test is written before the implementation to make it go green then a test case (like the list one) that never went red to begin with would have never sneaked in.

Arithmomaniac: `IReadOnlyList` fails without the optimization _now_, but if `Enumerable.ToList` ever gets optimized for it, then that would no longer be true. It's only failing now because, again, of the internal implementation of `ToList`. The underlying problem is the same; we're counting on internal implementation details.

atifaziz: I think I'm being a little daft here so let's go slower. As far as I can see in [the conditions at the start of `ScanRight`](https://github.com/Arithmomaniac/MoreLINQ/blob/f856d0df72ad2ca8d48145c4039f11e481ecad77/MoreLinq/ScanRight.cs#L93-L95
), `ToList` is not called in the `IReadOnlyList<>` case:

```c#
            var list = source is IReadOnlyList<TSource> readOnlyList
                     ? readOnlyList.AsListLike()
                     : (source as IList<TSource> ?? source.ToList()).AsListLike();
```

In fact, it's only when the source is neither type of list. Am I missing something?

Arithmomaniac: That is correct. `ToList` will only get called if it is not list-like.

atifaziz: Right, so then it makes sense to keep `[TestCase(SourceKind.ReadOnlyList)]`?

Arithmomaniac: I'll try going slower too:

- We would like to test that `IReadOnlyList` is not calling `ToList`.
- We can currently check that because if it was calling `ToList`, it would fail. This is because `ToList` enumerates over the input, and the enumerator for a `BreakingReadOnlyList` throws an exception.
- However, just as checking that `CopyTo` by `BreakingList` was called would be relying on an implementation detail (as you have noted above), so would assuming that `ToList` enumerates an `IReadOnlyList`.
  - At some future point, a `CopyTo`-like optimization could be created for `IReadOnlyList`, and the test would pass even if `ToList` _was_ being called.

The underlying issue is that there is no way to directly check whether `ToList` was called, and this is true as much of `IReadOnlyList` as `IList` (notwithstanding that they currently behave differently when the optimization is removed).

atifaziz: Thanks for going slower. 😄 I see your rationale, which is that it's futile to test for something that could be passing accidentally in the future even if that's not the case today. It's a false positive either way. Right? If so, I guess it's a fair point and sorry for not seeing it earlier and making you repeat yourself in so many ways.

So at this stage, we could:

1. Leave them in as harmless
2. Leave them in but mark them ignored
3. Leave them in but commented out
4. Leave them out altogether

What's your take? I could settle for either 2 or 3.

Finally, we good with expanding the comment [as I suggested earlier](https://github.com/morelinq/MoreLINQ/pull/476#discussion_r186014387)?

Arithmomaniac: It's all good :-)
I think I'll take 3, and expand the comment

PR #483, opened 2018-05-12
Test that RandomSubset returns original sequence elements

leandromoh:
add a test that ensure random subset returns exactly the same elements of original sequence

### DIFF
--- a/MoreLinq.Test/RandomSubsetTest.cs
+++ b/MoreLinq.Test/RandomSubsetTest.cs
@@ -206,6 +206,22 @@ namespace MoreLinq.Test
             Assert.That(sequence, Is.EqualTo(sequenceClone));
         }
 
+        /// <summary>
+        /// Verify that RandomSubset produces subset where all elements belongs to original sequence.
+        /// </summary>
+        [Test]
+        public void TestRandomSubsetReturnsSameElementsThatSequence()
+        {
+            const int count = 100;
+            var sequence = Enumerable.Range(1, count);
+            var resultA = sequence.RandomSubset(count);
+            var resultB = sequence.RandomSubset(count, new Random(12345));
+
+            // ensure random subset returns exactly the same elements of original sequence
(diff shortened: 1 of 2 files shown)


MoreLinq.Test/RandomSubsetTest.cs
```diff
@@ -206,6 +206,25 @@ public void TestRandomSubsetIsIdempotent()
...
 
+        /// <summary>
+        /// Verify that RandomSubset produces subset where all elements belongs to original sequence.
+        /// </summary>
+        [Test]
+        public void TestRandomSubsetReturnsSameElementsThatSequence()
+        {
+            const int count = 100;
+            var sequence = Enumerable.Range(1, count);
+            var resultA = sequence.RandomSubset(count);
```
fsateler: It should be rare, but possible for RandomSubset to return the same source sequence. Then this test would occasionally fail, without any change in the sources or the underlying libraries

leandromoh:
I thought that, but since source has 100 elements, the probability of return the same source sequence is 1 in 100 factorial.

MoreLinq.Test/RandomSubsetTest.cs
atifaziz: A simple workaround would be to send a custom `Random` implementation as the second argument, one that will never allow for this to happen.

leandromoh: I thought that, but since source has 100 elements, the probability of return the same source sequence is 1 in 100 factorial.

Without this validation, one can change implementation to always return source (what is obvious wrong) and all tests will pass.

fsateler: Note that `resultB` does not suffer from this problem, so I figure that you could just keep that one?

leandromoh: I removed resultA from test. (y)

Issue #470, opened 2018-04-26
Shuffle: RandomSubset overload without subset size


leandromoh:
I saw #409 by @MkazemAkhgary and looking into `RandomSubset` implementation I noted that case user wants to shuffle a sequence entirely, one can use `RandomSubset` operator, but it forces user evaluate a lazy sequence twice, for example.

```c#
// lazy sequence, that may be expensive
var source = Enumerable.Range(1, 10);

// first iteration to get the length of sequence
var size = source.Count(); 

// second iteration (there is a ToArray() inside RandomSubset to avoid change original source)
var shuffle = source.RandomSubset(size); 
```

I propose the new Shuffle operator:
```c#
IEnumerable<T> Shuffle<T>(IEnumerable<T> sequence)
```

This have the following advantages:

1. semantically correct (user wants shuffle sequence entirely, not a subset)
2. avoid unnecessary computations (iterates source once)
3. less code for user write

leandromoh:
@atifaziz?

fsateler:
I think this is useful. About the naming, what do other languages/toolkits/libraries have for equivalent functionality?

atifaziz:
@leandromoh Makes sense. If there was a `Shuffle` in MoreLINQ then `RandomSubset` would have been called `PartialShuffle` but it's a bit late in the day for that. I think `Shuffle` is a good name because a `RandomSubset` overload without the subset size parameter will look weird since there won't be any sub-setting. Question then is if we want to have `RandomSubset` as an alias for `PartialShuffle` or leave it as-is. I'd rather mark the former obsolete if it's going to be an alias but not drop it for 3.0. This can be discussed separately. Meanwhile, `Shuffle` is a good compromise.

> what do other languages/toolkits/libraries have for equivalent functionality

I think [_shuffle_ is a safe bet](http://rosettacode.org/wiki/Knuth_shuffle).

leandromoh:
Good. I will submit a PR.

PR #484, opened 2018-05-12
Shuffle implementation (issue 470)

leandromoh:


atifaziz:
:memo: This PR is for #470

leandromoh:
I will add a overload that receive Random

leandromoh:
> I will add a overload that receive Random

done

MoreLinq/RandomSubset.cs
```diff
@@ -23,6 +23,45 @@ namespace MoreLinq
...
     public static partial class MoreEnumerable
     {
+        /// <summary>
+        /// Returns a sequence of random elements from the original sequence.
+        /// </summary>
+        /// <typeparam name="T">The type of source sequence elements.</typeparam>
+        /// <param name="sequence">
+        /// The sequence from which to return random elements.</param>
+        /// <returns>
+        /// A random sequence of elements in random order from the original
```
atifaziz: Random is mentioned twice which confuses the mind. Rephrase to:

> A sequence of elements in random order from the original…

Is it worth mentioning that there is no chance of duplication, unless of course, the original has duplicates in the first place?

atifaziz: Random is mentioned twice which confuses the mind. Rephrase to:

> A sequence of elements in random order from the original…

Is it worth mentioning that there is no chance of duplication, unless of course, the original has duplicates in the first place?

MoreLinq/RandomSubset.cs
```diff
@@ -23,6 +23,45 @@ namespace MoreLinq
...
+        /// <typeparam name="T">The type of source sequence elements.</typeparam>
+        /// <param name="sequence">
+        /// The sequence from which to return random elements.</param>
+        /// <param name="rand">
+        /// A random generator used as part of the selection algorithm.</param>
+        /// <returns>
+        /// A random sequence of elements in random order from the original
+        /// sequence.
+        /// </returns>
+        public static IEnumerable<T> Shuffle<T>(this IEnumerable<T> sequence, Random rand)
```
atifaziz: Add blank line for clarity.

Also, even though this builds on top of `RandomSubset` is really one thing and the same, can we move this (along with its overloads) to another source file called `Shuffle.cs`?

MoreLinq/RandomSubset.cs
```diff
@@ -23,6 +23,45 @@ namespace MoreLinq
...
+        /// Returns a sequence of random elements from the original sequence.
+        /// </summary>
+        /// <typeparam name="T">The type of source sequence elements.</typeparam>
+        /// <param name="sequence">
+        /// The sequence from which to return random elements.</param>
+        /// <returns>
+        /// A random sequence of elements in random order from the original
+        /// sequence.
+        /// </returns>
+        public static IEnumerable<T> Shuffle<T>(this IEnumerable<T> sequence)
```
atifaziz: Add blank line for clarity.

---
PS I understand we need to do this consistently elsewhere in the source code, and we'll get round to it eventually, but meanwhile this also renders it consistent with this source file.

MoreLinq/RandomSubset.cs
```diff
@@ -23,6 +23,45 @@ namespace MoreLinq
...
+        {
+            return Shuffle(sequence, new Random());
+        }
+
+        /// <summary>
+        /// Returns a sequence of random elements from the original sequence.
```
atifaziz: This summary is a duplicate of the other. It can be rendered unique by saying that, “An additional parameter specifies a randomizer”.

MoreLinq/RandomSubset.cs
```diff
@@ -61,40 +100,43 @@ public static IEnumerable<T> RandomSubset<T>(this IEnumerable<T> sequence, int s
...
-            {
-                // The simplest and most efficient way to return a random subet is to perform
-                // an in-place, partial Fisher-Yates shuffle of the sequence. While we could do
-                // a full shuffle, it would be wasteful in the cases where subsetSize is shorter
-                // than the length of the sequence.
-                // See: http://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
+            return RandomSubsetImpl(sequence, rand, source => (source.ToArray(), subsetSize));
+        }
+
+        static IEnumerable<T> RandomSubsetImpl<T>(IEnumerable<T> sequence, Random rand, Func<IEnumerable<T>, (T[] array, int subsetSize)> seeder)
```
atifaziz: The element names on the tuple are unused so let's skip them. This is not a public API so no one will clear. The usage is obvious from the decomposition in the implementation:

```c#
var (array, subsetSize) = seeder(sequence);
```

MoreLinq.Test/ShuffleTest.cs
```diff
@@ -0,0 +1,86 @@
...
+        public void ShuffleWithEmptySequence()
+        {
+            var source = Enumerable.Empty<int>();
+            var result = source.Shuffle();
+
+            Assert.That(result, Is.Empty);
+        }
+
+        [Test]
+        public void ShuffleIsIdempotent()
```
atifaziz: I'm not sure I quite understand what you are trying to test here and with `ShuffleSeedWithEmptySequence`. Could you explain? Is there a way the implementation could change that would cause this test to break?

leandromoh: This test check that original source is not change (by side-effect) case It implements IList<T>. RandomSubset has the same test.

The test fails if we do the bellow

array = source as IList<TSource> ?? source.ToList().

atifaziz: > `RandomSubset` has the same test.

I didn't see that and was looking at `Shuffle` on its own but thanks for clarifying. I now see from the comment in the same test for `RandomSubset`…

https://github.com/morelinq/MoreLINQ/blob/d49025012812436039805196a63663a258ded06d/MoreLinq.Test/RandomSubsetTest.cs#L186-L190

…what it's trying to test. I'm not sure if _idempotent_ is the right adjective here but that's a different story.

PR #497, opened 2018-05-23
Added ToListLike to avoid repetitive code

leandromoh:


MoreLinq/AggregateRight.cs
```diff
@@ -47,10 +47,7 @@ public static TSource AggregateRight<TSource>(this IEnumerable<TSource> source,
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (func == null) throw new ArgumentNullException(nameof(func));
 
-            var list
```
atifaziz: Why not also in the other overload?

https://github.com/morelinq/MoreLINQ/blob/5048831832e3928fa1682e510fb3f330e2710f47/MoreLinq/AggregateRight.cs#L89-L91

MoreLinq/ListLike.cs
```diff
@@ -35,6 +36,17 @@ static class ListLike
         public static IListLike<T> AsListLike<T>(this IList<T> list) => new List<T>(list);
         public static IListLike<T> AsListLike<T>(this IReadOnlyList<T> list) => new ReadOnlyList<T>(list);
 
+        public static IListLike<T> ToListLike<T>(this IEnumerable<T> source)
```
atifaziz: `source` argument should be validated to be non-null.

To avoid too much mixing of styles, let's convert this to a simple expression, [like you did also in PR #499](https://github.com/morelinq/MoreLINQ/pull/499/files#diff-d5ae966de6f42beabe6e1199cd4c4bf3R60):

```c#
public static IListLike<T> ToListLike<T>(this IEnumerable<T> source)
    => source is null ? throw new ArgumentNullException(nameof(source))
     : source is IList<T> list ? new List<T>(list)
     : source is IReadOnlyList<T> readOnlyList ? (IListLike<T>) new ReadOnlyList<T>(readOnlyList)
     : new List<T>(source.ToList());
```

leandromoh: update done. the idea is after PR #499 be merged refactore this one to 

```c#
public static IListLike<T> ToListLike<T>(this IEnumerable<T> source)
    => source is null ? throw new ArgumentNullException(nameof(source))
          : source.TryAsListLike() is IListLike<T> listLike
               ? listLike
               : new List<T>(source.ToList());
```

PR #503, opened 2018-05-27
SkipErroneous implementation (issue #515)

leandromoh:
`IgnoreErrors `operator, as proposed in this [comment](https://github.com/morelinq/MoreLINQ/issues/405#issuecomment-370264244)


atifaziz:
How is this related to #405 and #435?

Issue #405 has a lot of discussion that could confuse future visitors. If your idea is to go forward with just [`IgnoreErrors` suggestion](https://github.com/morelinq/MoreLINQ/issues/405#issuecomment-370264244) then we should close #405 and either open a new issue or add the details here with example usages.

leandromoh:
> If your idea is to go forward with just IgnoreErrors suggestion then we should close #405 and either open a new issue or add the details here with example usages.

created [#515](https://github.com/morelinq/MoreLINQ/issues/515).

> What do you think about adding simpler overloads without the predicates where you want to ignore errors just be their type?

I think better keep the predicate, since it represents the `when` keyword in the `catch` block. Without it, the user can not simulate a true `catch` block.

atifaziz:
> I think better keep the predicate, since it represents the `when` keyword in the `catch` block.

Yes, of course it does, but the `when` bit in a `catch` statement is optional. When all you have to do is filter on the exception type then requiring a predicate seems a little verbose. In your own tests, you have three cases.

> Without it, user can not simulate a true `catch` block.

What's a _true_ `catch` block?

MoreLinq/IgnoreErrors.cs
```diff
@@ -0,0 +1,159 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
```
atifaziz: This import is not used/needed.

MoreLinq.Test/IgnoreErrorsTest.cs
```diff
@@ -0,0 +1,92 @@
...
+            var expectations = Enumerable.Range(1, 5);
+
+            Assert.That(result, Is.EqualTo(expectations));
+        }
+
+        [Test]
+        public void IgnoreErrorsWithParserFunction()
+        {
+            var source = "O,l,2,3,4,S,6,7,B,9".Split(',').Select(int.Parse);
+            var result = source.IgnoreErrors((FormatException ex) => true);
```
atifaziz: Introduce a simpler overload that allows saying `source.IgnoreErrors<FormatException>()` when predicate will unconditionally return `true`?

MoreLinq/IgnoreErrors.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// A sequence of elements <paramref name="source"/> that ignore the exceptions of type
+        /// <typeparamref name="TException1"/> or <typeparamref name="TException2"/>
+        /// in accordance with predicates.
+        /// </returns>
+        /// <remarks>
+        /// This operator uses deferred execution and streams its results.
+        /// Exceptions are caught in order which predicates were passed, as well in a catch block.
+        /// </remarks>
+
+        public static IEnumerable<T> IgnoreErrors<T, TException1, TException2>(
```
atifaziz: This overload is not tested.

MoreLinq.Test/IgnoreErrorsTest.cs
```diff
@@ -0,0 +1,92 @@
...
+        [Test]
+        public void IgnoreErrors()
+        {
+            var source = MoreEnumerable.From(() => 1,
+                                             () => throw new TestException("True"),
+                                             () => 2,
+                                             () => throw new TestException("False"),
+                                             () => 3);
+
+            var result = source.IgnoreErrors((TestException e) => bool.Parse(e.Message));
```
atifaziz: This predicate looks strange. Why not simply use [`Exception.Data`](https://docs.microsoft.com/en-us/dotnet/api/system.exception.data#System_Exception_Data) instead, like below?

```c#
var key = new object();
var source = MoreEnumerable.From(() => 1,
                                 () => throw new TestException { Data = { [key] = true } },
                                 () => 2,
                                 () => throw new TestException { Data = { [key] = false } },
                                 () => 3);

var result = source.IgnoreErrors((TestException e) => (bool) e.Data[key]);
```

MoreLinq.Test/TestException.cs
```diff
@@ -21,5 +21,9 @@ namespace MoreLinq.Test
     /// Reserved for use within tests.
     /// </summary>
 
-    sealed class TestException : System.Exception {}
+    sealed class TestException : System.Exception
```
atifaziz: Revert these changes because they're not needed (based on my earlier suggestion to use `Exception.Data` instead).

MoreLinq.Test/IgnoreErrorsTest.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            Assert.That(result, Is.EqualTo(expectations));
+        }
+
+        [Test]
+        public void IgnoreErrorsWithParserFunction()
+        {
+            var source = "O,l,2,3,4,S,6,7,B,9".Split(',').Select(int.Parse);
+            var result = source.IgnoreErrors((FormatException ex) => true);
+            var expectations = new[] { 2, 3, 4, 6, 7, 9 };
```
atifaziz: The blank line in-between `expectations` and the assertion confuses things. They belong together:

```c#
var expectations = Enumerable.Range(1, 5);
Assert.That(result, Is.EqualTo(expectations));
```

If the idea is to delineate the [AAA](http://wiki.c2.com/?ArrangeActAssert) parts of the test then the blank line should at least come between the _act_ and _assert_ parts, as in:

```c#
var result = source.IgnoreErrors((Exception e) => true);

var expectations = Enumerable.Range(1, 5);
Assert.That(result, Is.EqualTo(expectations));
```

Do we even need `expectations`? I would even simplify to just:

```c#
result.AssertSequenceEqual(1, 2, 3, 4, 5);
```

and alternatively:

```c#
source.IgnoreErrors((Exception e) => true)
      .AssertSequenceEqual(1, 2, 3, 4, 5);;
```

Review other tests for the same.

MoreLinq.Test/IgnoreErrorsTest.cs
```diff
@@ -0,0 +1,92 @@
...
+                                             () => 5);
+
+            var result = source.IgnoreErrors((Exception e) => true);
+            var expectations = Enumerable.Range(1, 5);
+
+            Assert.That(result, Is.EqualTo(expectations));
+        }
+
+        [Test]
+        public void IgnoreErrorsWithParserFunction()
```
atifaziz: Rename to `IgnoreErrorsInParsing`, `IgnoreErrorsInIntegerParsing` or `IgnoreErrorsInInt32Parsing`.

MoreLinq.Test/IgnoreErrorsTest.cs
```diff
@@ -0,0 +1,92 @@
...
+// limitations under the License.
+#endregion
+
+namespace MoreLinq.Test
+{
+    using NUnit.Framework;
+    using System;
+
+    [TestFixture]
+    public class IgnoreErrorsTest
```
atifaziz: All tests should use `TestingSequence` to ensure:

- The source is not enumerated more than once.
- The source enumerator is disposed.

MoreLinq.Test/IgnoreErrorsTest.cs
```diff
@@ -0,0 +1,92 @@
...
+            var result = source.IgnoreErrors((Exception e) => true);
+            var expectations = Enumerable.Range(1, 5);
+
+            Assert.That(result, Is.EqualTo(expectations));
+        }
+
+        [Test]
+        public void IgnoreErrorsWithParserFunction()
+        {
+            var source = "O,l,2,3,4,S,6,7,B,9".Split(',').Select(int.Parse);
```
atifaziz: Use the invariant culture with `int.Parse`.

MoreLinq/IgnoreErrors.cs
```diff
@@ -0,0 +1,244 @@
...
+#endregion
+
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+
+    static partial class MoreEnumerable
+    {
+   //   static Func<Exception, bool> defaultErrorPredicate = (Exception ex) => false;
```
atifaziz: Remove

README.md
```diff
@@ -246,6 +246,12 @@ selector function.
 
 This method has 4 overloads.
 
+### IgnoreErrors
+
+Ignore some exceptions that can occurs during the iteration of the sequence.
+
+This method has 3 overloads.
```
atifaziz: Isn't this 6 overloads now?

MoreLinq/IgnoreErrors.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            return _(); IEnumerable<T> _()
+            {
+                using (var e = source.GetEnumerator())
+                {
+                    while (true)
+                    {
+                        try
+                        {
+                            if (!e.MoveNext()) break;
```
atifaziz: `break` on next line for consistency. It would have been okay on the same line as you have it now if you styled the `catch` blocks likewise, as in:

```c#
try { if (!e.MoveNext()) break; }
catch (TException1 ex) when (error1Predicate?.Invoke(ex) ?? true) { continue; }
catch (TException2 ex) when (error2Predicate?.Invoke(ex) ?? true) { continue; }
catch (TException3 ex) when (error3Predicate?.Invoke(ex) ?? true) { continue; }
```

leandromoh:
all adjusts done

atifaziz:
Closing this since as the linked issue #515 was abandoned.

Issue #98, opened 2015-08-21
`Batch` should `IEnumerable<T[]>`


atifaziz:
Since batches are about splitting an `IEnumerable` into fixed sizes, every batch should know how many items it has, in addition to the original order of the source enumerable. Size + order preservation = `IList`.

This should break little (if any) existing code, because `IList` inherits `IEnumerable`. Asides from the signature, the only code change that needs to be made is for the underlying bucket store to be a `List` instead of an `Array` (which is the more LINQy approach anyways).

---

Originally reported on Google Code with ID 98

Reported by @Arithmomaniac on 2015-03-09 14:19:50

fsateler:
This looks reasonable to me (after all the length is known beforehand). However, if done it should be done for 2.0, or else for the next major version as that signature change would break the ABI.

atifaziz:
I would argue that even returning a sequence of arrays might be best here, which is required, could then be cast to an `IList<T>` or [wrapped into a read-only collection](https://msdn.microsoft.com/en-us/library/53kysx7b(v=vs.110).aspx). And while we're on the topic of buffered views, the same could apply to [`Windowed`][Windowed] and [`Partition`][Partition].

  [Partition]: https://github.com/morelinq/MoreLINQ/blob/6608a4b932f6e33ceea0ba6d64b811fad471c1ae/MoreLinq/Partition.cs
  [Windowed]: https://github.com/morelinq/MoreLINQ/blob/6608a4b932f6e33ceea0ba6d64b811fad471c1ae/MoreLinq/Windowed.cs

atifaziz:
> every batch should know how many items it has

Every batch will be the same size except the last one if the source sequence doesn't have a total count of elements that's a multiple of the requested size.

> Size + order preservation = `IList`.

`Batch` does not mandate order preservation & the signature currently reflects that. If batching is tomorrow implemented in a parallel scatter-gather fashion then it just needs to stream batches but requiring order preservation could pin implementations to being less efficient than they could be.

> in addition to the original order of the source enumerable. 

This is almost the case with most LINQ operators yet they don't return `IList<T>`. You'd think, for example, that something as simple as [`Select`](https://msdn.microsoft.com/en-us/library/bb548891(v=vs.110).aspx) would preserve order but it doesn't when using Parallel LINQ (see [“Order Preservation in PLINQ”](https://msdn.microsoft.com/en-us/library/dd460677(v=vs.110).aspx)):

[...]

PR #508, opened 2018-05-30
Changed Windowed to use a circular queue

allenbenz:
Changes Windowed to use a circular queue instead of copying and allocation per window.

Not sure if this conflicts with the desire to change the return type to an `IEnumerable<IList<T>>` (#98)

### DIFF
--- a/MoreLinq/Windowed.cs
+++ b/MoreLinq/Windowed.cs
@@ -55,18 +55,31 @@ namespace MoreLinq
                     // return the first window (whatever size it may be)
                     yield return window;
 
-                    // generate the next window by shifting forward by one item
+                    // generate the next window by using a circular queue
+                    // shifting forward by one item after each pass
+                    int j = 0;
                     while (iter.MoveNext())
                     {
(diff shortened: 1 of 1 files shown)


atifaziz:
Thanks for your contribution but I am afraid that as 4cb2846f6adb78c6281e7e437186165ceaa96afe demonstrates, it breaks previously yielded windows since the underlying array is modified. Thanks to this PR, I added an additional test in 99aa8c394f390074ea76b2b808f7f7d2f3789130 to ensure that the implementation doesn't do that:

https://github.com/morelinq/MoreLINQ/blob/99aa8c394f390074ea76b2b808f7f7d2f3789130/MoreLinq.Test/WindowedTest.cs#L104-L116

That [test doesn't fail for the current implementation](https://ci.appveyor.com/project/raboof/morelinq/build/1021) but [does for on this PR](https://ci.appveyor.com/project/raboof/morelinq/build/1022).

allenbenz:
Ah, yep, no way to get that behavior without copying or restarting the enumerable.

Thanks for the review, closing.

Issue #496, opened 2018-05-22
Rename Concat overload to Append


leandromoh:
how about rename `Concat `to `Append `in the signature` Concat<T>(this IEnumerable<T> head, T tail)`  ?

I know it is a breaking-change, but release 3.0 is a good opportunity.

This have the following pros:

- if we have `prepend`, make a lot o sense have `append`.
- avoid confusion between our `Concat `and Linq's one

atifaziz:
> - if we have `prepend`, make a lot o sense have `append`.

The value is very marginal.

> - avoid confusion between our `Concat `and Linq's one

What's so confusing about it? We have other such overloads sharing the same name as the base operator in LINQ. It'll be equally confusing if renamed to `Append` since [`Append` is now part of LINQ][Append]:


[Append]: https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.append?view=netcore-2.0#System_Linq_Enumerable_Append__1_System_Collections_Generic_IEnumerable___0____0_

leandromoh:
I see. At least the operator would be semantically correct.  Since Concat is about sequencies, and what We are doing is append a value in the end of the sequêncie.

PR #511, opened 2018-05-31
Rename Concat to Append, marking Concat obsolete

atifaziz:
This PR addresses #496.

`Concat` will still ship but is marked obsolete and calls `Append` behind the scenes. It will then be removed in a future version.


atifaziz:
> Why not drop It together with other overload (extending T) and NestedLoops?

I imagine it's more commonly used than the other two. This extended `IEnumerable<>` whereas neither of the other two did.

> otherwise, when It will happen? Release 4.0?

3.1 would be good enough as long as people got the warning in 3.0.

leandromoh:
I see. About code, LGTM

Issue #380, opened 2018-02-10
Support returning some count of Max/MinBy


atifaziz:
Since #328 changes [`MaxBy`][maxby] and [`MinBy`][minby] to return a list, I propose to add `PartialMaxBy` and `PartialMinBy` that take an extra parameter indicating the maximum count of extrema elements to return. It avoids wastage building a list of all extrema elements ~and scanning the entire source sequence~ when one may be satisfied with, for example, any of the first ones.

[maxby]: http://morelinq.github.io/1.x/ref/api/html/Overload_MoreLinq_MoreEnumerable_MaxBy.htm
[minby]: http://morelinq.github.io/1.x/ref/api/html/Overload_MoreLinq_MoreEnumerable_MinBy.htm

leandromoh:
It does not make sense, because we always have to Scan The entire source, since the extreme element may be the last one.

atifaziz:
@leandromoh Of course 🤦‍♂️ and thanks for setting me straight. However, it is still useful to not have to waste building a list of all extrema if one is happy with less.

leandromoh:
@atifaziz How common is this cenario (source with **great** number of extrema elements)? 
Does it justify a new operator?

atifaziz:
@leandromoh I can't speak for the world so it's hard for me to say whether it's common or not. The one time I did need `MinBy` and `MaxBy`, I know I needed a list as a result. Asking for a single item (as they've been implemented up to now) means you'll get any one of the extrema and that's rarely useful.

The real reason for proposing partial versions is to provide an option for those people who have been relying on `MinBy` and `MaxBy` returning a single item as we've made a breaking change and both methods have been part of MoreLINQ's history for an exceptionally long time; in fact, since as early as [the second commit](https://github.com/morelinq/MoreLINQ/commit/76ae6927b347e3ee66c07dff7c6e1bec2e99f889)! They have even been [mentioned](https://books.google.ch/books?id=3HqnBQAAQBAJ&lpg=PA122&ots=79WaRJwsJc&dq=morelinq%20minby%20maxby&pg=PA122#v=onepage&q=morelinq%20minby%20maxby&f=false) in the book [Thinking in LINQ](https://www.apress.com/us/book/9781430268451)!

`PartialMinBy` and `PartialMaxBy` with an upper limit are interesting if you're want some sample of an extremum, like 10 files of the largest files size in a directory. For small samples, I agree you can compose `MinBy` or `MaxBy` with `Take`.

PR #513, opened 2018-06-01
Add support for partial Min/MaxBy

atifaziz:
This PR addresses #380.

It adds support for partial query of `MinBy` and `MaxBy` without introducing any new methods. “It just works” with natural methods of LINQ like `Take` and `First`. It does this by introducing a new `IExtremaEnumerable<>`, which is now returned by `MinBy` and `MaxBy`:

```c#
public interface IExtremaEnumerable<out T> : IEnumerable<T>
{
    IEnumerable<T> Take(int count);
    IEnumerable<T> TakeLast(int count);
}
```

By composing `MaxBy` or `MinBy` with `Take` or `TakeLast`, the internal iteration is optimised to only keep the requested count of elements in memory at any given time. What's more, if `count` is 1 for either then yet another optimised version ensures that only a single element is maintained (i.e. no cost of a collection type is incurred). There are also extensions defined for `IExtremaEnumerable<>`:

- `First`
- `FirstOrDefault`
- `Last`
- `LastOrDefault`
- `Single`
- `SingleOrDefault`

These extensions internally call `Take` or `TakeLast` with a count of 1 (with the exception of `Single` and `SingleOrDefault` that use a count of 2), which effectively gives back the single-element version of `MaxBy` and `MinBy` up to the breaking change in #328 and without much additional cost of building a list only to throw away everything except one element.

## Examples

```c#
var strings = new[] { "ax", "hello", "world", "aa", "ab", "ay", "az" };
Console.WriteLine(strings.MinBy(s => s.Length).Take(3).ToDelimitedString(", "));
// output: ax, aa, ab
```

```c#
var strings = new[] { "ax", "hello", "world", "aa", "ab", "ay", "az" };
Console.WriteLine(strings.MaxBy(s => s.Length).First());
// output: hello
```

```c#
var strings = new[] { "ax", "hello", "world", "aa", "ab", "ay", "az" };
Console.WriteLine(strings.MaxBy(s => s.Length).Last());
// output: world
```

### DIFF
--- a/MoreLinq.Test/MaxByTest.cs
+++ b/MoreLinq.Test/MaxByTest.cs
@@ -59,5 +59,57 @@ namespace MoreLinq.Test
         {
             Assert.AreEqual(new[] { "aa" }, SampleData.Strings.MaxBy(x => x[1], SampleData.ReverseCharComparer));
         }
+
+        [TestCase(0, ExpectedResult = new string[0]             )]
+        [TestCase(1, ExpectedResult = new[] { "hello"          })]
+        [TestCase(2, ExpectedResult = new[] { "hello", "world" })]
+        [TestCase(3, ExpectedResult = new[] { "hello", "world" })]
+        public string[] MaxByTakeReturnsMaxima(int count)
+        {
+            using (var strings = SampleData.Strings.AsTestingSequence())
+                return strings.MaxBy(s => s.Length).Take(count).ToArray();
+        }
+
+        [TestCase(0, ExpectedResult = new string[0]             )]
+        [TestCase(1, ExpectedResult = new[] { "world"          })]
+        [TestCase(2, ExpectedResult = new[] { "hello", "world" })]
+        [TestCase(3, ExpectedResult = new[] { "hello", "world" })]
+        public string[] MaxByTakeLastReturnsMaxima(int count)
+        {
+            using (var strings = SampleData.Strings.AsTestingSequence())
+                return strings.MaxBy(s => s.Length).TakeLast(count).ToArray();
+        }
+
+        [TestCase(0, 0, ExpectedResult = new string[0]                         )]
+        [TestCase(3, 1, ExpectedResult = new[] { "aa"                         })]
+        [TestCase(1, 0, ExpectedResult = new[] { "ax"                         })]
+        [TestCase(2, 0, ExpectedResult = new[] { "ax", "aa"                   })]
+        [TestCase(3, 0, ExpectedResult = new[] { "ax", "aa", "ab"             })]
+        [TestCase(4, 0, ExpectedResult = new[] { "ax", "aa", "ab", "ay"       })]
+        [TestCase(5, 0, ExpectedResult = new[] { "ax", "aa", "ab", "ay", "az" })]
+        [TestCase(6, 0, ExpectedResult = new[] { "ax", "aa", "ab", "ay", "az" })]
+        public string[] MaxByTakeWithComparerReturnsMaxima(int count, int index)
+        {
+            using (var strings = SampleData.Strings.AsTestingSequence())
+                return strings.MaxBy(s => s[index], SampleData.ReverseCharComparer)
+                              .Take(count)
+                              .ToArray();
+        }
+
+        [TestCase(0, 0, ExpectedResult = new string[0]                         )]
+        [TestCase(3, 1, ExpectedResult = new[] { "aa"                         })]
+        [TestCase(1, 0, ExpectedResult = new[] { "az"                         })]
+        [TestCase(2, 0, ExpectedResult = new[] { "ay", "az"                   })]
+        [TestCase(3, 0, ExpectedResult = new[] { "ab", "ay", "az"             })]
+        [TestCase(4, 0, ExpectedResult = new[] { "aa", "ab", "ay", "az"       })]
+        [TestCase(5, 0, ExpectedResult = new[] { "ax", "aa", "ab", "ay", "az" })]
+        [TestCase(6, 0, ExpectedResult = new[] { "ax", "aa", "ab", "ay", "az" })]
+        public string[] MaxByTakeLastWithComparerReturnsMaxima(int count, int index)
+        {
+            using (var strings = SampleData.Strings.AsTestingSequence())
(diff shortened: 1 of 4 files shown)


leandromoh:
Just to clarify: are we adding a new public interface, and exposing it instead of IEnumerable in signatures?

atifaziz:
> are we adding a new public interface, and exposing it instead of IEnumerable in signatures?

Yes.

atifaziz:
@leandromoh I think it gives the smoothest API without any new learning. As I said in the opening comment, it'll do the right thing if you combine `MaxBy` or `MinBy` with optimised extensions and which share the same name as extensions on `IEnumerable<>`.

leandromoh:
> I think it gives the smoothest API without any new learning

User knows nothing about `IExtremaEnumerable<T>`, so one will have to learn about it.

Why not simply maintain as `internal` and return `IEnumerable<T>`? Personally I don't think that optimizations is a strong case to introduce a new public interface.

atifaziz:
> User knows nothing about `IExtremaEnumerable<T>`, so one will have to learn about it.

What does the have to learn about it? It's completely transparent to the flow of the queries. The only time the user has to be aware is during a break at the point of `MaxBy` or `MinBy` and continuation of the query, where the type would bleed. Nobody should ever need to expose `IExtremaEnumerable<>` in their public API.

Let me ask you, when did you last care about [`IOrderedEnumerable <>`](https://docs.microsoft.com/en-gb/dotnet/api/system.linq.iorderedenumerable-1)? Did you have to _learn_ anything about it?

I do agree that the abstraction may leak at times because the optimisation wouldn't take effect.

> Why not simply maintain `IExtremaEnumerable<T>` as `internal` and return `IEnumerable<T>`? 

There's no need for it to exist internally. Either it's public or it doesn't exist at all. It's sole purpose is to _trap_ composition with `Take` or `TakeLast` so as to detect the intention to use only part of the result!

> Personally I don't think that optimizations is a strong case to introduce a new public interface.

That's a fair point we can debate on. Are you saying the optimisations are not worth it at all or that the public interface is not worth it but that the optimisation should be done another way?

I went through quite some thought process to come up with this API shape. The other alternatives end up a bit somewhat ugly.  One is to have specific methods:

```c#
public enum Extremity { First, Last }

public static IEnumerable<T> PartialMaxBy<T, K>(this IEnumerable<T> source,
    Extremity extremity, int count, Func<T, K> selector);
public static IEnumerable<T> PartialMinBy<T, K>(this IEnumerable<T> source,
    Extremity extremity, int count, Func<T, K> selector);
```

If we don't want to introduce a new enumeration then we have to go with 4 methods:

```c#
public static IEnumerable<T> PartialFirstMaxBy<T, K>(this IEnumerable<T> source,
    int count, Func<T, K> selector);
public static IEnumerable<T> PartialLastMaxBy<T, K>(this IEnumerable<T> source,
    int count, Func<T, K> selector);

public static IEnumerable<T> PartialFirstMinBy<T, K>(this IEnumerable<T> source,
    Extremity extremity, int count, Func<T, K> selector);
public static IEnumerable<T> PartialLastMinBy<T, K>(this IEnumerable<T> source,
    Extremity extremity, int count, Func<T, K> selector);
```

So, yes, the overloads approach avoids a new interface but now you have to _embed_ details of what's going on in the method names and parameters. The `Extremity` enumeration removes some of that at the cost of introducing a new type.

Appreciate your review, feedback & comments.

atifaziz:
One way to avoid `IExtremaEnumerable<>` would be to introduce duplicates of `Take` and `TakeLast`, and `First`, `Last`, and `Single` _and_ their `*OrDefault` versions, in MoreLINQ. These duplicates could then optimise on an internal type coming out of `MinBy` and `MaxBy`. However, that approach is quite fragile and easy to get wrong since it depends on the user always being aware of using the MoreLINQ version.

atifaziz:
One way to avoid `IExtremaEnumerable<>` would be to introduce duplicates of `Take` and `TakeLast`, and `First`, `Last`, and `Single` (_and_ their `*OrDefault` versions) in MoreLINQ. These duplicates could then optimise on an internal type coming out of `MinBy` and `MaxBy`. However, that approach is quite fragile and easy to get wrong since it depends on the user always being aware of using the MoreLINQ version.

leandromoh:
> Let me ask you, when did you last care about IOrderedEnumerable <>? Did you have to learn anything about it?

Never, I shall admit.

> Are you saying the optimisations are not worth it at all or that the public interface is not worth it but that the optimisation should be done another way?

Honestly I do not think these optimisations are worth, since it requires time for implementation, review, documentation and maintenance, and sincerely users usually do not care about optmizations at all.

atifaziz:
@leandromoh:

> Honestly I do not think these optimizations are worth, since…

> …it requires implementation time,

Already done!

> …review,

It's too bad no one took the time to review thoroughly since this PR has been open so I'll agree with you on this point assuming lack of motivation rather than time.

> …documentation

Already done!

> …and maintenance, 

The implementation is minimal since they are simple wrappers to the base methods so hopefully there won't be much maintenance.

> …and sincerely users usually do not care about optimizations if it does not hurt the performance of the application.

How can you say that? 😕 If users do not care about optimisations then we can remove all optimisations we have worked on in MoreLINQ. :wink: It's one thing to start with an initial implementation and leave optimisations as a later enhancement (for sake of YAGNI) but a whole different approach to say “users do not care about optimisations” so let's not do any.

Whether the design in this PR is a good one or not is something time will tell. As I [already explained](https://github.com/morelinq/MoreLINQ/pull/513#issuecomment-393952551), I went through several design iterations before landing with the one proposed in this PR and I would like to go forward fully conscious of the fact that it could be a mistake. Let's take Nelson Mandela's stance on this, “[We] never lose. [We] either win or learn.” 😆

AsValeO:
Nice work!

atifaziz:
@AsValeO Thanks for the feedback.

akqira:
Just for information that maybe can be useful, i'm using **MaxBy(...)** in my code and after upgrading  to version **3.0.0** it raised an complication error saying 
`Cannot implicitly convert type MoreLinq.IExtremaEnumerable<myCustomType>  to MyCustomType...`
Then i've done this to fix it:
`MaxBy(...).FirstOrDefault();`

Hope it will help and good job @atifaziz

PR #529, opened 2018-08-07
Set initial capacity on buffer queues

dferretti-fig:
Lag.cs included offset when creating its buffer queue, but not Lead or CountDown.

### DIFF
--- a/MoreLinq/CountDown.cs
+++ b/MoreLinq/CountDown.cs
@@ -84,7 +84,7 @@ namespace MoreLinq
 
             IEnumerable<TResult> IterateSequence()
             {
-                var queue = new Queue<T>();
+                var queue = new Queue<T>(count + 1);
 
                 foreach (var item in source)
                 {
--- a/MoreLinq/Lead.cs
+++ b/MoreLinq/Lead.cs
@@ -64,7 +64,7 @@ namespace MoreLinq
 
             return _(); IEnumerable<TResult> _()
             {
-                var leadQueue = new Queue<TSource>();
+                var leadQueue = new Queue<TSource>(offset);
                 using (var iter = source.GetEnumerator())
                 {
                     bool hasMore;


MoreLinq/CountDown.cs
```diff
@@ -84,7 +84,7 @@ IEnumerable<TResult> IterateCollection(int i)
 
             IEnumerable<TResult> IterateSequence()
             {
-                var queue = new Queue<T>();
+                var queue = new Queue<T>(count + 1);
```
atifaziz: This is problematic when count is negative (more specifically, less than -1) because `CountDown` permits a negative count. In fact, this highlighted that we're missing a test and that I've added in bb05a54e4fd1d00aef48ac830481e312702555be. It now demonstrates that this change [breaks](https://ci.appveyor.com/project/raboof/morelinq/build/1076#L596):

```
1) Error : MoreLinq.Test.CountDownTest.WithNegativeCount
System.ArgumentOutOfRangeException : Non-negative number required.
Parameter name: capacity
Actual value was -999.
   at System.Collections.Generic.Queue`1..ctor(Int32 capacity)
   at MoreLinq.MoreEnumerable.<>c__DisplayClass28_0`2.<<CountDown>g__IterateSequence|2>d.MoveNext()
   at NUnit.Framework.Constraints.NUnitEqualityComparer.EnumerablesEqual(IEnumerable x, IEnumerable y, Tolerance& tolerance)
   at NUnit.Framework.Constraints.NUnitEqualityComparer.AreEqual(Object x, Object y, Tolerance& tolerance)
   at NUnit.Framework.Constraints.EqualConstraint.ApplyTo[TActual](TActual actual)
   at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression, String message, Object[] args)
   at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression)
   at MoreLinq.Test.TestExtensions.AssertSequenceEqual[T](IEnumerable`1 actual, IEnumerable`1 expected)
   at MoreLinq.Test.CountDownTest.WithNegativeCount()
```

dferretti-fig:
Thanks for taking a look at it! I've pushed a change to guard against a negative count, using either `count + 1` or `1`, whichever is greater.

PR #537, opened 2018-10-09
Fix CountBy to handle null keys

### LINKED ISSUE #536: CountBy does not work with null keys
`new string[] { "a", null, "a", null }.CountBy(x=>x)`

expected:
"a" - 2,
null - 2

Why I need it to work with nulls: to do `source.CountBy(x=>x).All(x => x.Value == 1)` instead of `source.GrounBy(x=>x).Count() == source.Count()`.

atifaziz:
This PR fixes #536.

### DIFF
--- a/MoreLinq/CountBy.cs
+++ b/MoreLinq/CountBy.cs
@@ -84,23 +84,31 @@ namespace MoreLinq
                     var havePrevKey = false;
                     var prevKey = default(TKey);
                     var index = 0;
+                    var nullIndex = (int?) null;
+
+                    int? IndexOf(TKey key) =>
+                        key == null ? nullIndex is int ni ? ni : (int?) null
+                                    : dic.TryGetValue(key, out var i) ? i : (int?) null;
 
                     foreach (var item in source)
                     {
                         var key = keySelector(item);
 
                         if (// key same as the previous? then re-use the index
-                            (havePrevKey && cmp.GetHashCode(prevKey) == cmp.GetHashCode(key)
-                                          && cmp.Equals(prevKey, key))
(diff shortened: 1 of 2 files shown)


MoreLinq/CountBy.cs
```diff
@@ -84,23 +84,31 @@ void Loop(IEqualityComparer<TKey> cmp)
...
                         var key = keySelector(item);
 
                         if (// key same as the previous? then re-use the index
-                            (havePrevKey && cmp.GetHashCode(prevKey) == cmp.GetHashCode(key)
-                                          && cmp.Equals(prevKey, key))
+                            havePrevKey && cmp.GetHashCode(prevKey) == cmp.GetHashCode(key)
+                                         && cmp.Equals(prevKey, key)
                             // otherwise try & find index of the key
-                            || dic.TryGetValue(key, out index))
+                            || IndexOf(key) is int i && (index = i) >= 0 /* always true */)
```
fsateler: I think I prefer this to be folded into `IndexOf`: 

```csharp
bool TryGetIndex(TKey key, out int index) {
  if (key == null) {
    index = nullIndex.GetValueOrDefault();
    return nullIndex.HasValue;
  }
  else {
    return dic.TryGetValue(key, out index);
  }
}
```

Assignments within `if` statements always feel icky to me (perhaps this is C-caused "trauma"?)

atifaziz:
@fsateler I've addressed your comments. We good to merge this?

atifaziz:
> since we need this only here for now, let's apply the rule of three and refactor when the time comes.

Agree.

PR #541, opened 2018-10-31
Update ForEach.

hetaoos:
Take source as the return value.


atifaziz:
See [`Pipe`](https://morelinq.github.io/3.0/ref/api/html/M_MoreLinq_MoreEnumerable_Pipe__1.htm), which already does this.

PR #548, opened 2018-11-15
Reworked GroupAdjacent and Batch to perform fully deferred execution

thibault-reigner:
#### What's this PR do?
Simplify the code of GroupAdjacent and Batch to remove internal partial enumeration relying on instantiation of arrays and lists, hence performing fully deferred enumeration of the sources.
In my tests it improves the performances by avoiding allocation of large objects which often end up in the generation 2 part of the heap.

### DIFF
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -27,14 +27,14 @@ namespace MoreLinq.Test
         public void BatchZeroSize()
         {
             AssertThrowsArgument.OutOfRangeException("size",() =>
-                new object[0].Batch(0));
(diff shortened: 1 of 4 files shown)


leandromoh:
your `Batch `implementation return `[3,  4,  5,  6,  7,  8,  9,  10,  11,  12,] `for `Enumerable.Range(1,30).Batch(10).ElementAt(2)` while it must return `[21,  22,  23,  24,  25,  26,  27,  28,  29,  30]`. I already had this bug (and resolve it :D ) in [my PR for a lazy Batch](https://github.com/morelinq/MoreLINQ/pull/402)

thibault-reigner:
@leandromoh nice catch ! Indeed skipping the enumeration of a batch will generate incorrect results for the next one.

PR #551, opened 2018-12-20
Refactored Scan implementation

leandromoh:
Refactored Scan implementation to avoid repetitive code.


leandromoh:
Title and description updated @atifaziz

PR #553, opened 2018-12-24
Refactored Traverse implementation 

leandromoh:
Refactored Traverse implementation to avoid repetitive code.

### DIFF
--- a/MoreLinq/Traverse.cs
+++ b/MoreLinq/Traverse.cs
@@ -18,6 +18,7 @@
 namespace MoreLinq
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.Linq;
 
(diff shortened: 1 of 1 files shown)


MoreLinq/Traverse.cs
```diff
@@ -88,19 +78,30 @@ public static IEnumerable<T> TraverseDepthFirst<T>(T root, Func<T, IEnumerable<T
         {
             if (childrenSelector == null) throw new ArgumentNullException(nameof(childrenSelector));
 
+            // because a stack pops the elements out in LIFO order, we need to push them in reverse
+            // if we want to traverse the returned list in the same order as was returned to us
+
+            var stack = new Stack<T>();
+            return TraverseImpl(root, x => childrenSelector(x).Reverse(), stack.Push, stack.Pop, stack);
```
Orace: We can avoid the reverse by using a stack (or it is a queue), of children enumerator. We should take care of the dispose.

MoreLinq/Traverse.cs
```diff
@@ -88,19 +78,30 @@ public static IEnumerable<T> TraverseDepthFirst<T>(T root, Func<T, IEnumerable<T
...
+        }
+
+        static IEnumerable<T> TraverseImpl<T>(
+            T root,
+            Func<T, IEnumerable<T>> childrenSelector,
+            Action<T> Push,
+            Func<T> Pop,
```
fsateler, line 92: I don't think we have any other method naming arguments in Capital case.

MoreLinq/Traverse.cs
fsateler: Maybe that can be a separate change.

fsateler (changes requested):
Honestly I'm not sure the deduplicated code really makes things more readable. There is a level on indirection on methods that are quite small. @atifaziz WDYT?

PR #559, opened 2019-02-10
Pairwise refactored

leandromoh:


leandromoh:
@atifaziz how about this refactoring?

fsateler:
I'd say it depends on the performance impact. Do we have any measurement? Normally I wouldn't put  much emphasis on optimizations, but given MoreLINQ is a fairly basic (as in, basic building block) library, we have to cater to somewhat weird uses (like long lists to be consumed).

Moreover, I'd say that if this refactoring goes in, `Pairwise` is not adding much value over Scan. Why not deprecate the method?

atifaziz:
> I'd say it depends on the performance impact. Do we have any measurement?

So before we get to this…

> Moreover, I'd say that if this refactoring goes in, `Pairwise` is not adding much value over Scan. Why not deprecate the method?

I don't think we should be deprecating existing methods that we have had for a long time on the sole basis that they could have been expressed by composing other methods. @leandromoh's refactoring composes on `Scan`, `Skip` and `Select`. At 3 compositions, I consider the chain long enough to deserve being encapsulated. You could argue that the last `Select` would go away if expressed in-lined so we are really down to two compositions. But that's not all. You have consider the call-site arguments too because the cognitive tax of writing and reading a seed for `Scan`, even if made up of defaults, and remembering to skip exactly 2 elements seems a bit much to ask. On top of that, the seed cannot be expressed when your source elements are of an anonymous type, which is the real deal breaker for me.

But `Pairwise` can be re-written even more simply using `Lead` & `SkipLast`:

```c#
public static IEnumerable<TResult> Pairwise<TSource, TResult>(this IEnumerable<TSource> source, Func<TSource, TSource, TResult> resultSelector) =>
    source.Lead(1, resultSelector).SkipLast(1);
```

This reads more natural, you don't have to come up with a seed, the 1 for `Lead` and `SkipLast` shouldn't be all that surprising (also much easier to commit to memory than 2) and it works with anonymous types when expressed in-line!

Yet another reason to not deprecate `Pairwise` is that [it's there in F# too](https://msdn.microsoft.com/visualfsharpdocs/conceptual/seq.pairwise%5b%27t%5d-function-%5bfsharp%5d), and in addition to [`Scan`](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/seq.scan%5b't,'state%5d-function-%5bfsharp%5d), so for folks working in both languages, it provides good parity.

Coming back to the question of performance, I measured the 3 approaches:

- `PairwiseDecomposed` is `Pairwise` as-is today (decomposed to its most basic form and loop)
- `PairwiseUsingScan` is the refactored version proposed by @leandromoh in this PR
- `PairwiseUsingLead` is my alternate version from above

The benchmark code was very simple:

```c#
Enumerable.Range(1, 1000).Pairwise(ValueTuple.Create).Consume();
```

The results are:

|             Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
| `PairwiseDecomposed` | 13.37 us | 0.1043 us | 0.0871 us | 0.0305 |     - |     - |     184 B |
|  `PairwiseUsingScan` | 33.84 us | 0.4678 us | 0.4147 us | 0.0610 |     - |     - |     488 B |
|  `PairwiseUsingLead` | 57.67 us | 0.3519 us | 0.3292 us | 0.1221 |     - |     - |     608 B |

The benchmark configuration & host details were:

``` 
BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17763.437 (1809/October2018Update/Redstone5)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=2.1.500
  [Host]     : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.9 (CoreCLR 4.6.27414.06, CoreFX 4.6.27415.01), 64bit RyuJIT
```

So my version, while simpler, is the slowest by over a factor of 4 😱 and @leandromoh is over twice as slow. I'm not surprised, frankly, since `Lead` & `SkipLast` involve more work. Given these numbers, I'd like to return to and re-iterate my original remark:

> while the refactoring is intellectually interesting, I wonder if we should leave this one as-is because 3 compositions means performance is worse now

In other words, we are not gaining anything except having more internal reuse. While it makes the code more readable, it's not reducing our maintenance (which has been practically zero since in 7 years since its addition).

@fsateler said:

> we have to cater to somewhat weird uses (like long lists to be consumed).

How is consuming long lists weird? 😕 You seem to be saying MoreLINQ should be fast because it's a basic library but that working with long lists is exceptional? Not sure I follow your argument.

fsateler:
> >we have to cater to somewhat weird uses (like long lists to be consumed).
>
>How is consuming long lists weird? confused You seem to be saying MoreLINQ should be fast because it's a basic library but that working with long lists is exceptional? Not sure I follow your argument.

Sorry, I was not very clear. I meant less common, instead of strange. Here is a more expanded version that hopefully makes more sense:

In my dayjob, I mostly work with user-facing line of business applications. Within that context, (micro) optimizations rarely matter: it is quite rare for such an application to work with large enough data that these optimizations have any effect. I mean, both 13µs and 57µs are imperceptible. If execution time scales linearly, lists 20.000 times larger are still "instantaenously" processed as far as the human in front of the app is concerned [(source)](https://www.pubnub.com/blog/how-fast-is-realtime-human-perception-and-technology/).

Within that background, I would usually advocate for the "easier to read" version within any of my apps, because code readability is paramount, and optimizations are very rarely relevant.

That said, there are cases where you need to process large enough data. Maybe your app is a number-crunching, data-science type of app. Or maybe your dataset is occasionally large; this has even [happened to me](https://github.com/morelinq/MoreLINQ/pull/129#issuecomment-173192111)!

Because MoreLINQ is a relatively basic building block of many apps, we don't know if we are faced with my line-of-business app, or a number crunching app. This would place a relatively higher weight on performant code, as surely someone, somewhere is using this for large datasets.

I think the approach outlined by @atifaziz  in https://github.com/morelinq/MoreLINQ/pull/129#issuecomment-173226421 , is the correct one: correct first, optimize later (and when needed). If this was adding the method, I would not complain. But this PR seems to go back from optimized to non-optimized.

> Coming back to the question of performance, I measured the 3 approaches:

Thanks for doing the measurements. With those results, I think it is not worth refactoring the method, since the original method is not unreadable. It is imperative style, but it is a relatively short method and easy to follow.

atifaziz:
> Sorry, I was not very clear. I meant less common, instead of strange. Here is a more expanded version that hopefully makes more sense:

Thanks for clarifying and sharing all your thoughts.

> But this PR seems to go back from optimized to non-optimized.

Right.

> I think it is not worth refactoring the method, since the original method is not unreadable. It is imperative style, but it is a relatively short method and easy to follow.

Agree.

@leandromoh Thanks for this PR but I am afraid we won't be merging this based on the comments from @fsateler and I. Hope this doesn't discourage you from future contributions.

Issue #560, opened 2019-02-18
New IndexBy operator


leandromoh:
Allow us to know the index of the elements inside their groups created by a `keySelecton` function.

### Prototype

```c#
static IEnumerable<TResult> IndexBy<T, TKey, TResult>(this
    IEnumerable<T> source,
    Func<T, TKey> keySelector,
    Func<T, TKey, int, TResult> resultSelector)
{
    return source
        .Select((element, index) => new { element, index })
        .GroupBy(pair => keySelector(pair.element))
        .SelectMany(grup => grup.Select((pair, innerIndex) => new { pair, grup.Key, innerIndex }))
        .OrderBy(x => x.pair.index)
        .Select(x => resultSelector(x.pair.element, x.Key, x.innerIndex));
}
```

### Example

```c#
var nomes = new[] { "ana", "beatriz", "carla", "bob", "davi", "adriano", "angelo", "carla", };

var resultado = nomes.IndexBy(x => x.First(), (nome, key, index) => new { nome, key, index });
```

outputs

```c#
{ nome = ana, key = a, index = 0 }
{ nome = beatriz, key = b, index = 0 }
{ nome = carla, key = c, index = 0 }
{ nome = bob, key = b, index = 1 }
{ nome = davi, key = d, index = 0 }
{ nome = adriano, key = a, index = 1 }
{ nome = angelo, key = a, index = 2 }
{ nome = carla, key = c, index = 1 }

```

atifaziz:
This is interesting and I trust you'll be sending a PR in the near future?

PR #562, opened 2019-03-01
Added IndexBy operator

leandromoh:
need help with method summary/readme, please


leandromoh:
@atifaziz on AppVeyor build I am receiving "New code was generated during build that's not been committed." error. I belive that it is about generated extension wrappers. Can you help me please?

atifaziz:
> AppVeyor build I am receiving "New code was generated during build that's not been committed." error. I belive that it is about generated extension wrappers. Can you help me please?

That's correct. You just need to run `build.cmd` or `build.sh` locally before doing a commit (which is what I did with c215552d016d0e2499dbe2f0e7541503a8382f82). Those scripts will always make sure that the generated code is up to date. Doing a build in an IDE like Visual Studio or Visual Studio Code isn't enough.

leandromoh:
Nice. When I tried to run these cmd I got an error and the console closed before I could read. May we follow with the review?

atifaziz:
This PR addresses #560.

MoreLinq.Test/IndexByTest.cs
```diff
@@ -0,0 +1,100 @@
...
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class IndexByTest
+    {
+        [Test]
+        public void IndexBySimpleTest()
+        {
+            var source = new[] { "ana", "beatriz", "carla", "bob", "davi", "adriano", "angelo", "carlos" };
+            var result = source.IndexBy(x => x.First(), (item, key, index) => (item, key, index));
```
atifaziz: `(item, key, index) => (item, key, index)` can be shortened to `ValueTuple.Create` but if the goal here is to name the elements, then please use PascalCase for consistency or just an anonymous-typed object like `(e, k, i) => new { Item = e, Key = k, Index = i }`.

MoreLinq.Test/IndexByTest.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            result.AssertSequenceEqual(
+                ( item: "ana",     key: 'a', index: 0 ),
```
atifaziz: For tuples, avoid space after `(` and before `)`.

You don't need to name the elements of the tuples at all, but if it's for readability, then choose between one of the following:

- Tuples with no element names, e.g. `("ana", 'a', 0)` is acceptable (here)
- Tuples with PascalCase element names
- Anonymous-type objects with PascalCase property names (GC allocations and speed concerns are negligible for these tests)

MoreLinq/CountBy.cs
```diff
@@ -55,76 +55,129 @@ static partial class MoreEnumerable
...
+                }
 
-                for (var i = 0; i < keys.Count; i++)
-                    yield return new KeyValuePair<TKey, int>(keys[i], counts[i]);
+                return dic.TryGetValue(key, out i);
+            }
+
+            var keys = new List<TKey>();
+            var counts = new List<int>();
+            var elements = new List<TSource>();
```
atifaziz: We do we need `elements`? It's never used.

leandromoh:
@atifaziz adjusts done. Except by `elements`. I did it for consistency, `IndexBy `need elements, `CountBy `does not, but `CountByImpl `does not know them, so it always provides.

MoreLinq/IndexBy.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <typeparam name="TKey">Type of the projected key.</typeparam>
+        /// <typeparam name="TResult">Type of the projected element.</typeparam>
+        /// <param name="source">Source sequence.</param>
+        /// <param name="keySelector">Function that transforms each item of source sequence into a key to be compared against the others.</param>
+        /// <param name="resultSelector">The projection to each element, its key and index.</param>
+        /// <param name="comparer">The equality comparer to use to determine whether or not keys are equal.
+        /// If null, the default equality comparer for <typeparamref name="TSource"/> is used.</param>
+        /// <returns>A sequence of unique keys and their number of occurrences in the original sequence.</returns>
+
+        public static IEnumerable<TResult> IndexBy<TSource, TKey, TResult>(
```
fsateler: I feel like `IndexBy` is not very descriptive of what this does. Other languages like Ruby have `index_by` as what C# knows as `ToDictionary`. I'm having a hard time coming up with a better name. Since this is mostly `CountBy`, where the last parameter to `resultSelector` is decremented once, does this really make sense to add?

atifaziz: > I feel like `IndexBy` is not very descriptive of what this does. Other languages like Ruby have `index_by` as what C# knows as `ToDictionary`. I'm having a hard time coming up with a better name.

It's consistent with [`Index`](https://morelinq.github.io/3.1/ref/api/html/Overload_MoreLinq_MoreEnumerable_Index.htm), which pairs each element of sequence with its position in the sequence. `IndexBy` does the same except the positions are relative to a key group. That's how the name makes sense to me; that it's a variation of `Index`.

> Since this is mostly `CountBy`, where the last parameter to `resultSelector` is decremented once, does this really make sense to add?

`CountBy` aggregates on keys. This gives you the source elements in the result so is different enough to be useful.

atifaziz:
I did a quick spike and implemented `IndexBy` in terms of `ScanBy` with 16f2cb4911da3fb3736572bd427a23ca03799e1a. In terms of design, I believe we should keep `IndexBy` as simple as `Index`. In other words, it shouldn't return any user-defined projection, but just key-value pairs where the key is the index within the group. The only downside is that it can be somewhat confusing to have a _by operator_ where the returned key is something other than what `keySelector` returns but I imagine that wouldn't be too surprising either since this is just another flavour of `Index`.

leandromoh:
@atifaziz glad you figure out the method benefícits. The implantation looks great. About no materialize entire source, we should add a ‘*DoesNotIterateUnnecessaryElements’ test to assure It.

leandromoh:
@atifaziz adjusts done

PR #578, opened 2019-05-01
Pairwise: Change example on summary for clarity

leandromoh:


leandromoh:
> Please change this PR to contain one of the two changes and open a separate one for the other.

done! thanks for that tip.

PR #583, opened 2019-05-05
ZipShortest: moved particular behavior from remarks to summary

leandromoh:
Currently, all zip methods have the same summary, what is a bad thing because user can get confused thinkg they all do the same thing. 

The particular behavior of each opertator is being quoted on remarks sessions, but if the particular behaviors are important enough for morelinq have 3 different kinds of zip, then it should be discern on the summary session.


leandromoh:
@atifaziz adjusts done!

atifaziz:
This was merged via PR #595. Thanks!

PR #584, opened 2019-05-05
EquiZip: moved differential behavior from remarks to summary

leandromoh:
Currently, all zip methods have the same summary, what is a bad thing because user can get confused thinkg they all do the same thing. 

The particular behavior of each opertator is being quoted on remarks sessions, but if the particular behaviors are important enough for morelinq have 3 different kinds of zip, then it should be discern on the summary session.

### DIFF
--- a/MoreLinq/EquiZip.cs
+++ b/MoreLinq/EquiZip.cs
@@ -27,7 +27,7 @@ namespace MoreLinq
         /// <summary>
         /// Returns a projection of tuples, where each tuple contains the N-th
         /// element from each of the argument sequences.
-        /// If the two input sequences are of different lengths then
+        /// If the input sequences are of different lengths then
         /// <see cref="InvalidOperationException"/> is thrown.
         /// </summary>
         /// <typeparam name="TFirst">Type of elements in first sequence.</typeparam>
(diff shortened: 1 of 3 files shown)


MoreLinq/EquiZip.cs
```diff
@@ -49,11 +51,7 @@ static partial class MoreEnumerable
         /// "2B", "3C", "4D" in turn.
         /// </example>
         /// <remarks>
-        /// <para>
```
atifaziz: This should be moved to into an `<exception>` section following the `<returns>` section, like to:

```c#
        /// <returns>
        /// A sequence that contains elements of the two input sequences,
        /// combined by <paramref name="resultSelector"/>.
        /// </returns>
        /// <exception cref="InvalidOperationException">
        /// The input sequences are of different lengths.</exception>
        /// <example>
```

MoreLinq/EquiZip.cs
```diff
@@ -27,6 +27,8 @@ static partial class MoreEnumerable
         /// <summary>
         /// Returns a projection of tuples, where each tuple contains the N-th
         /// element from each of the argument sequences.
+        /// If the input sequences are of different lengths then
```
atifaziz: Summaries generally don't reference types so let's change this to either:

```c#
        /// <summary>
        /// Returns a projection of tuples, where each tuple contains the N-th
        /// element from each of the argument sequences, if the input sequences
        /// are of different lengths.
        /// </summary>
```

If mentioning the throwing of an exception in the summary is important, then here's a version for that:

```c#
        /// <summary>
        /// Returns a projection of tuples, where each tuple contains the N-th
        /// element from each of the argument sequences. An exception is thrown
        /// if the input sequences are of different lengths.
        /// </summary>
```

I'm fine with either.

leandromoh:
@atifaziz adjusts done!

atifaziz:
This was merged via PR #595. Thanks!

Issue #538, opened 2018-10-10
Add deconstruction of sequence types


atifaziz:
Consider adding extensions like the following:

```c#
public static void Deconstruct<T>(this IEnumerable<T> source, out T item1, out T item2)
{
    using (var e = source.GetEnumerator())
    {
        var more = true;
        item1 = more && (more = e.MoveNext()) ? e.Current : default;
        item2 = more && (more = e.MoveNext()) ? e.Current : default;
    }
}

public static void Deconstruct<T>(this IEnumerable<T> source, out T item1, out T item2, out T item3)
{
    using (var e = source.GetEnumerator())
    {
        var more = true;
        item1 = more && (more = e.MoveNext()) ? e.Current : default;
        item2 = more && (more = e.MoveNext()) ? e.Current : default;
        item3 = more && (more = e.MoveNext()) ? e.Current : default;
    }
}
```

that would enable that would enable deconstruction of sequence, e.g.:

```c#
var (x, y) = new[] { 123, 456 };
var (a, b, c) = "foobar";
var (name, value) = "foo=bar".Split('=', 2); 
```

For more background, see [“Deconstructing a user-defined type with an extension method”](https://docs.microsoft.com/en-us/dotnet/csharp/deconstruct#deconstructing-a-user-defined-type-with-an-extension-method).

There should be overloads of `Deconstruct` to allow deconstruction of up to first 16 elements of a sequence.

These extensions should be _exclusively_ defined in a new namespace called `MoreLinq.Deconstruction` so that they have to be imported _explicitly_. They do not need to made available in the `MoreLinq.Extensions` namespace that designed to avoid conflict in extensions.

Deconstruction shouldn't throw exceptions if the sequence is too short.

fsateler:
I think this is a nice idea, but unfortunately, I think this:

> Deconstruction shouldn't throw exceptions if the sequence is too short

Might be the wrong choice. I'm not going to argue that the choice is per-se wrong, but that it depends on the context. Thus, I wonder if enumerable deconstruction is not general enough to fit in such a general library as MoreLINQ.

atifaziz:
[...]

PR #586, opened 2019-05-10
Add AsDeconstructible

atifaziz:
This PR implements deconstruction of sequences as proposed in #538.

### DIFF
--- a/MoreLinq.Test/DeconstructibleEnumerableTest.cs
+++ b/MoreLinq.Test/DeconstructibleEnumerableTest.cs
@@ -22,13 +22,13 @@ namespace MoreLinq.Test
     using NUnit.Framework;
 
     [TestFixture]
-    public class DeconstructibleSequenceTest
+    public class DeconstructibleEnumerableTest
     {
         [Test]
         public void InitWithNullSource()
         {
             var e = Assert.Throws<ArgumentNullException>(() =>
-                new DeconstructibleSequence<object>(null));
+                new DeconstructibleEnumerable<object>(null));
 
             Assert.That(e.ParamName, Is.EqualTo("source"));
         }
@@ -37,7 +37,7 @@ namespace MoreLinq.Test
         public void DeconstructDefault()
         {
             var e = Assert.Throws<InvalidOperationException>(() =>
-                default(DeconstructibleSequence<int>).Deconstruct(out _, out _));
+                default(DeconstructibleEnumerable<int>).Deconstruct(out _, out _));
             Assert.That(e.Message, Contains.Substring("short"));
         }
 
--- a/MoreLinq.Test/DeconstructibleEnumerableTest.cs
+++ b/MoreLinq.Test/DeconstructibleEnumerableTest.cs
@@ -312,53 +312,43 @@ namespace MoreLinq.Test
             }
         }
 
-        [TestCase(2)]
-        [TestCase(3)]
-        [TestCase(4)]
-        [TestCase(5)]
-        [TestCase(6)]
-        [TestCase(7)]
-        [TestCase(8)]
-        [TestCase(9)]
-        [TestCase(10)]
-        [TestCase(11)]
-        [TestCase(12)]
-        [TestCase(13)]
-        [TestCase(14)]
-        [TestCase(15)]
-        [TestCase(16)]
-        public void DeconstructWithTooFewElements(int count)
+        [TestCase( 2,  1)]
+        [TestCase( 3,  1)]
+        [TestCase( 4,  1)]
+        [TestCase( 5,  1)]
+        [TestCase( 6,  1)]
+        [TestCase( 7,  1)]
+        [TestCase( 8,  1)]
+        [TestCase( 9,  1)]
+        [TestCase(10,  1)]
+        [TestCase(11,  1)]
+        [TestCase(12,  1)]
+        [TestCase(13,  1)]
+        [TestCase(14,  1)]
+        [TestCase(15,  1)]
+        [TestCase(16,  1)]
+        [TestCase( 2, -1)]
+        [TestCase( 3, -1)]
+        [TestCase( 4, -1)]
+        [TestCase( 5, -1)]
+        [TestCase( 6, -1)]
+        [TestCase( 7, -1)]
+        [TestCase( 8, -1)]
+        [TestCase( 9, -1)]
+        [TestCase(10, -1)]
+        [TestCase(11, -1)]
+        [TestCase(12, -1)]
+        [TestCase(13, -1)]
+        [TestCase(14, -1)]
+        [TestCase(15, -1)]
+        [TestCase(16, -1)]
+        public void DeconstructWithMismatchedLength(int count, int actualCountDelta)
         {
-            using (var xs = Enumerable.Range(1, count - 1).AsTestingSequence())
+            using (var xs = Enumerable.Range(1, count + actualCountDelta).AsTestingSequence())
             {
                 var e = Assert.Throws<InvalidOperationException>(() =>
                     Deconstruct(xs, count));
-                Assert.That(e.Message, Does.Match($@"\bfew( \w+)* {Regex.Escape(count.ToString(CultureInfo.InvariantCulture))}\b"));
-            }
-        }
-
-        [TestCase(2)]
-        [TestCase(3)]
-        [TestCase(4)]
-        [TestCase(5)]
-        [TestCase(6)]
-        [TestCase(7)]
-        [TestCase(8)]
-        [TestCase(9)]
-        [TestCase(10)]
-        [TestCase(11)]
-        [TestCase(12)]
-        [TestCase(13)]
-        [TestCase(14)]
-        [TestCase(15)]
-        [TestCase(16)]
-        public void DeconstructWithTooManyElements(int count)
-        {
-            using (var xs = Enumerable.Range(1, count + 1).AsTestingSequence())
-            {
-                var e = Assert.Throws<InvalidOperationException>(() =>
-                    Deconstruct(xs, count));
-                Assert.That(e.Message, Does.Match($@"\bmany( \w+)* {Regex.Escape(count.ToString(CultureInfo.InvariantCulture))}\b"));
+                Assert.That(e.Message, Does.Match($@"\b{Regex.Escape(actualCountDelta > 0 ? "many" : "few")}( \w+)* {Regex.Escape(count.ToString(CultureInfo.InvariantCulture))}\b"));
             }
         }
 
--- a/MoreLinq.Test/DeconstructibleEnumerableTest.cs
+++ b/MoreLinq.Test/DeconstructibleEnumerableTest.cs
@@ -43,6 +43,14 @@ namespace MoreLinq.Test
         }
 
         [Test]
+        public void Enumeration()
+        {
+            var xs = Enumerable.Range(1, 10);
+            using (var ts = xs.AsTestingSequence())
+                ts.AsDeconstructible().AssertSequenceEqual(xs);
+        }
+
+        [Test]
         public void Deconstruct2()
         {
             using (var xs = Enumerable.Range(1, 2).AsTestingSequence())
--- a/MoreLinq.Test/DeconstructibleEnumerableTest.cs
+++ b/MoreLinq.Test/DeconstructibleEnumerableTest.cs
@@ -36,9 +36,8 @@ namespace MoreLinq.Test
         [Test]
         public void DeconstructDefault()
         {
-            var e = Assert.Throws<InvalidOperationException>(() =>
+            Assert.Throws<InvalidOperationException>(() =>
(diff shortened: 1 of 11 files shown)


MoreLinq/DeconstructibleEnumerable.cs
```diff
@@ -0,0 +1,86 @@
...
+            _source = source ?? throw new ArgumentNullException(nameof(source));
+
+        /// <inheritdoc />
+        public IEnumerator<T> GetEnumerator() =>
+            _source != null ? _source.GetEnumerator()
+                            : throw CreateTooShortError();
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        void Deconstruct(int count, out T item1, out T item2, out T item3, out T item4, out T item5, out T item6, out T item7, out T item8, out T item9, out T item10, out T item11, out T item12, out T item13, out T item14, out T item15, out T item16)
```
fsateler: I can't find where this method is used, it appears it isn't.

MoreLinq.Test/DeconstructibleEnumerableTest.cs
```diff
@@ -0,0 +1,387 @@
...
+
+            Assert.That(e.ParamName, Is.EqualTo("source"));
+        }
+
+        [Test]
+        public void DeconstructDefault()
+        {
+            var e = Assert.Throws<InvalidOperationException>(() =>
+                default(DeconstructibleEnumerable<int>).Deconstruct(out _, out _));
+            Assert.That(e.Message, Contains.Substring("short"));
```
fsateler: This should be a different message than for message too short: the `DeconstructibleEnumerable<>` object is in an invalid state.

MoreLinq/DeconstructibleEnumerable.cs
```diff
@@ -0,0 +1,86 @@
...
+    using System.Collections.Generic;
+    using System.Diagnostics;
+
+    /// <summary>
+    /// Represents a source whose elements can be sequentially deconstructed
+    /// into separate variables.
+    /// </summary>
+    /// <typeparam name="T">The type of the elements.</typeparam>
+
+    public partial struct DeconstructibleEnumerable<T> : IEnumerable<T>
```
fsateler: I assume this is a struct for optimization purposes. I wonder if we are not painting ourselves into a corner by restricting us to a single implementation. I could imagine wanting to implement a specialization for sources implementing `IList<T>`. OTOH, given that the number of elements is always short, maybe that optimization won't ever matter. What do you think?

MoreLinq/DeconstructibleEnumerable.cs
```diff
@@ -0,0 +1,86 @@
...
+        readonly IEnumerable<T> _source;
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="DeconstructibleEnumerable{T}"/>
+        /// with a sequence.
+        /// </summary>
+        /// <param name="source">
+        /// The source sequence on which to enable deconstruction.</param>
+
+        public DeconstructibleEnumerable(IEnumerable<T> source) =>
```
fsateler: Is it necessary to make this public? It could be internal, and then we expose less.

atifaziz: No and that's how I had it but I often feel that the one shouldn't lie about using a `struct`. Do we gain anything by blocking the use of this constructor?

MoreLinq.Test/DeconstructibleEnumerableTest.cs
atifaziz: Agree since one can't construct it with a null sequence. Fixed in a5dd44186cdc6389343928bb788a9059b692307c.

MoreLinq/DeconstructibleEnumerable.cs
atifaziz: This was leftover from an initial spike that was modelled after [`FoldImpl`](https://github.com/morelinq/MoreLINQ/blob/66b6bc4525201da0d763b51c4b91dcf4315df9d0/MoreLinq/Fold.cs#L25-L90) and thanks for spotting it. Removed in ad0df41c6852a1e0f88157d2a0456ef7f1d7b7c4.

MoreLinq/DeconstructibleEnumerable.cs
```diff
@@ -0,0 +1,86 @@
...
+            _source = source ?? throw new ArgumentNullException(nameof(source));
+
+        /// <inheritdoc />
+        public IEnumerator<T> GetEnumerator() =>
+            _source != null ? _source.GetEnumerator()
+                            : throw CreateTooShortError();
```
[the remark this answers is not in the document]
atifaziz: > I think a comment about that could aid readability, and this should throw a different error than the too short case.

Done in c2413aa78d05f1adb51ab17799f6cee37626f1b4.


> Also, indirection through a property would make it easier to pretend the source is never null.

Normally that'd be my approach too but seems excessive for the single use it would have here.

MoreLinq/DeconstructibleEnumerable.cs
```diff
@@ -0,0 +1,86 @@
...
+                if (e.MoveNext())
+                    throw CreateTooLongError();
+            }
+        }
+
+        static InvalidOperationException CreateTooShortError() =>
+            new InvalidOperationException("Sequence too short.");
+
+        static InvalidOperationException CreateTooLongError() =>
+            new InvalidOperationException("Sequence too long.");
```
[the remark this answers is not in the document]
atifaziz: Yes and addressed in e885659c7d9f8bb0e04c79bb898191c58bdf9f70.

MoreLinq/DeconstructibleEnumerable.cs
```diff
@@ -0,0 +1,86 @@
...
+                item14 = count >= 14 ? e.MoveNext() ? e.Current : throw CreateTooShortError() : default;
+                item15 = count >= 15 ? e.MoveNext() ? e.Current : throw CreateTooShortError() : default;
+                item16 = count >= 16 ? e.MoveNext() ? e.Current : throw CreateTooShortError() : default;
+                if (e.MoveNext())
+                    throw CreateTooLongError();
+            }
+        }
+
+        static InvalidOperationException CreateTooShortError() =>
+            new InvalidOperationException("Sequence too short.");
```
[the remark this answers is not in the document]
atifaziz: I was just being lazy but I'll go clean up the room now. :wink: Done in e885659c7d9f8bb0e04c79bb898191c58bdf9f70.

MoreLinq/DeconstructibleEnumerable.cs
atifaziz: The `struct` is there to make the _explicit intention to deconstruct_ cheap. Optimizations for other collection types would still be possible but they would belong with the `Deconstruct` overloads as opposed to `AsDeconstructible` deciding which implementation to return based on the source type. In other words, this is how I see it:

```c#
public void Deconstruct(out T item1, out T item2)
{
    const int count = 2;

    if (_source is List<T> list)
    {
        if (list.Count != 2)
            throw list.Count < count ? CreateTooShortError(count) : CreateTooLongError(count);
        item1 = list[0];
        item2 = list[1];
    }
    else
    {
        using (var e = GetEnumerator())
        {
            item1 = e.MoveNext() ? e.Current : throw CreateTooShortError(count);
            item2 = e.MoveNext() ? e.Current : throw CreateTooShortError(count);
            if (e.MoveNext())
                throw CreateTooLongError(count);
        }
    }
}
```

The decision to optimize is therefore deferred and delegated. The only downside I see to this is that the type test to choose an optimization path would have to be done on each call to `Deconstruct`, though that said, I honestly don't see anyone deconstructing more than once for any practical purpose as it would be costly unless the source is materialized or memoized.

MoreLinq/DeconstructibleEnumerable.cs
fsateler: Mostly freedom. I've usually thought it is better to offer the minimum visibility that is required. But I'm not sure it matters much in this case, as I don't think the constructor will be modified much.

MoreLinq/AsDeconstructible.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </returns>
+        /// <remarks>
+        /// This operator uses deferred execution semantics and streams its
+        /// results unless deconstructed.
+        /// </remarks>
+
+        public static DeconstructibleEnumerable<T> AsDeconstructible<T>(this IEnumerable<T> source)
+            => source == null
+             ? throw new ArgumentNullException(nameof(source))
+             : new DeconstructibleEnumerable<T>(source);
```
fsateler: I think `new DeconstructibleEnumerable<T>(source ?? throw new ArgumentNullException(nameof(source)));` looks nicer (sorry for not picking this one earlier).

MoreLinq/DeconstructibleEnumerable.cs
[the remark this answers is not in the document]
fsateler: You didn't print the actual number. Was that intentional?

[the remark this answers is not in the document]
atifaziz: Are you looking at outdated code? The code after the changes in e885659c7d9f8bb0e04c79bb898191c58bdf9f70 should read as follows:

```c#
static InvalidOperationException CreateTooShortError(int expectedCount) =>
    CreateLengthMismatchError(expectedCount, "few");

static InvalidOperationException CreateTooLongError(int expectedCount) =>
    CreateLengthMismatchError(expectedCount, "many");

static InvalidOperationException CreateLengthMismatchError(int expectedCount, string actualQuantifier) =>
    new InvalidOperationException($"Sequence contains too {actualQuantifier} elements when exactly {expectedCount} {(expectedCount == 1 ? "was" : "were")} expected.");
```

MoreLinq/DeconstructibleEnumerable.cs
atifaziz: > I've usually thought it is better to offer the minimum visibility that is required. 

I'm in that camp too, but technically speaking, this type stands on its own feet and does exactly what it advertises. I'm just playing the devil's advocate here. Will give it some more thought before merging the PR provided it's not a basis for your disapproval of the design.

MoreLinq/DeconstructibleEnumerable.cs
atifaziz: @fsateler Thoughts on this?

MoreLinq/DeconstructibleEnumerable.cs
fsateler: This is not a basis for disapproval, but I wondered if there was any need to make it public

MoreLinq/DeconstructibleEnumerable.cs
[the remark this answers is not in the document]
fsateler: Yes, I'm looking at the new code. I mean in the short case, we also know the actual number of elements encountered in the sequence.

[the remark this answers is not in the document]
atifaziz: I see what you're saying now. Unfortunately, your comment appears tagged next to the outdated code so that threw me off for a bit.

MoreLinq/DeconstructibleEnumerable.cs
atifaziz: …because I considered it [mostly harmless](https://en.wikipedia.org/wiki/Mostly_Harmless). 😄

MoreLinq/DeconstructibleEnumerable.cs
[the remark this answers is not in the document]
atifaziz: > we also know the actual number of elements encountered in the sequence.

True but wouldn't the information be somewhat superfluous? Does it add any value, especially when we can't do any better in the case of too many elements? Isn't knowing _the exact count_ expected the real thing? Call me lazy but I'd like to argue that this is an improvement we can make post-merge because it doesn't seem seem like a design defect. Also bear in mind that `Fold` has been the same for years now (and yet no one's complained that it could be better):

https://github.com/morelinq/MoreLINQ/blob/d0848d9bf9bfbde2b5141c16e90203d48fa44da6/MoreLinq/Fold.cs#L96-L99

MoreLinq/AsDeconstructible.cs
atifaziz: Hah! You should see the lengths to which people seem to be going these days to [please the JIT so that it will inline simple methods](https://github.com/dotnet/corefx/pull/35213) like this one! If we go as far, readability will go out the door.

MoreLinq/DeconstructibleEnumerable.cs
fsateler: I'm still giving it some thought. My instincts is towards the interface, but I don't think that practically we will need to ever add that optimization. Given the low item count, I'd be surprised if there was a significant gain to be made.

atifaziz:
Closing because of [decision to not move forward](https://github.com/morelinq/MoreLINQ/issues/538#issuecomment-518983110) with #538.

PR #590, opened 2019-05-17
Deconstruct for sequences (🎬 take 2)

atifaziz:
This is a second take on issue #538 and possibly supersedes #586 pending discussion and review. 

PR #586 implements a strict version that throws an exception unless the sequence contains the exact number of elements being deconstructed. The version proposed here is simpler. It extends `IEnumerable<>` directly with `Deconstruct` that outputs, in addition to elements, a count of the elements actually read. This in turn lets the caller decide what to do. The background, thought-process and benefits were laid out in [my following comment on issue #538](https://github.com/morelinq/MoreLINQ/issues/538#issuecomment-493042590):

> I'd like to give the choice of strict or non-struct deconstruction semantics another crack. If we provide a count of how many items were actually deconstructed, like so:
> 
> ```c#
> public static void Deconstruct<T>(this IEnumerable<T> source, out T item1, out T item2, out int count)
> {
>     count = 0;
>     using var e = source.GetEnumerator();
>     (count, item1) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
>     (count, item2) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
> }
> 
> public static void Deconstruct<T>(this IEnumerable<T> source, out T item1, out T item2, out T item3, out int count)
> {
>     count = 0;
>     using var e = source.GetEnumerator();
>     (count, item1) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
>     (count, item2) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
>     (count, item3) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
> }
> 
> public static void Deconstruct<T>(this IEnumerable<T> source, out T item1, out T item2, out T item3, out T item4, out int count)
> {
>     count = 0;
>     using var e = source.GetEnumerator();
>     (count, item1) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
>     (count, item2) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
>     (count, item3) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
>     (count, item4) = e.MoveNext() ? (count + 1, e.Current) : (count, default);
> }
> ```
> 
> then the caller has all the information and flexibility to decide what to do without having to worry about exceptions (and then also throwing an exception becomes a choice of the caller):
> 
> ```c#
> var result =
>     from s in new[]
>     {
>         "12",
>         "12,34",
>         "12,34,56",
>         "12,34,56,78",
>     }
>     select new
>     {
>         Input = s,
>         Result = s.Split(',').Select(int.Parse) switch
>         {
>             var (x, y, _, rc)    when rc == 2 => $"Point2D({x}, {y})",
>             var (x, y, z, _, rc) when rc == 3 => $"Point3D({x}, {y}, {z})",
>             _ => "?"
>         }
>     };
> 
> result.ForEach(Console.WriteLine);
> 
> /* prints:
> { Input = 12, Result = ? }
> { Input = 12,34, Result = Point2D(12, 34) }
> { Input = 12,34,56, Result = Point3D(12, 34, 56) }
> { Input = 12,34,56,78, Result = ? }
> */
> ```
> 
> Notice the use of [discard](https://docs.microsoft.com/en-us/dotnet/csharp/discards) (`_`) to probe an extra element so as to capture the case when there are more elements than desired.
> 
> With this approach, we don't need to hop over an additional method like `AsDeconstructible`, [as proposed earlier](https://github.com/morelinq/MoreLINQ/issues/538#issuecomment-487918865), just for sake of communicating the potential surprise of an exception due strict semantics.
> 
> I only see benefits:
> 
> - Back to simple `Deconstruct` extensions for sequences (that can be optimised for other collection types)
> - No semantic hop method like `AsDeconstructible`
> - No new superficial type like `DeconstructibleEnumerable` (as we have in PR #586 at this time) is needed
> - No exceptions
> - Caller has all control
> 

Moreover:

> the…example can be written with a constant pattern for the (read) count (`rc`) instead of using `when`:
> 
> ```cs
> var result =
>     from s in new[]
>     {
>         "12",
>         "12,34",
>         "12,34,56",
>         "12,34,56,78",
>     }
>     select new
>     {
>         Input = s,
>         Result = s.Split(',').Select(int.Parse) switch
>         {
>             (var x, var y, _, 2)        => $"Point2D({x}, {y})",
>             (var x, var y, var z, _, 3) => $"Point3D({x}, {y}, {z})",
>             _ => "?"
>         }
>     };
> 
> result.ForEach(Console.WriteLine);
> ```
> 
> While it may not be clear what 2 and 3 mean, one can use comments, as in `(var x, var y, _, /* count = */ 2)`, or revert to the `when` form and use a more meaningful variable name like `count`. Where immaterial, the count can be discarded. Whichever way you cut it, the choice is up to the developer. 🙌

### DIFF
--- a/MoreLinq.Test/DeconstructTest.cs
+++ b/MoreLinq.Test/DeconstructTest.cs
@@ -1,0 +1,126 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2019 Atif Aziz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq.Test
+{
+    using System;
+    using System.Collections.Generic;
+    using NUnit.Framework;
+    using NUnit.Framework.Interfaces;
+
+    [TestFixture]
+    public class DeconstructTest
+    {
+        [TestCase(2)]
+        [TestCase(3)]
+        [TestCase(4)]
+        [TestCase(5)]
+        [TestCase(6)]
+        [TestCase(7)]
+        [TestCase(8)]
+        [TestCase(9)]
+        [TestCase(10)]
+        [TestCase(11)]
+        [TestCase(12)]
+        [TestCase(13)]
+        [TestCase(14)]
+        [TestCase(15)]
+        [TestCase(16)]
+        public void WithNullSource(int width)
+        {
+            var source = (object[])null;
+
+            var e = Assert.Throws<ArgumentNullException>(() =>
+            {
+                switch (width)
+                {
+                    case 2: (_, _, _) = source; break;
+                    case 3: (_, _, _, _) = source; break;
+                    case 4: (_, _, _, _, _) = source; break;
+                    case 5: (_, _, _, _, _, _) = source; break;
+                    case 6: (_, _, _, _, _, _, _) = source; break;
+                    case 7: (_, _, _, _, _, _, _, _) = source; break;
+                    case 8: (_, _, _, _, _, _, _, _, _) = source; break;
+                    case 9: (_, _, _, _, _, _, _, _, _, _) = source; break;
+                    case 10: (_, _, _, _, _, _, _, _, _, _, _) = source; break;
+                    case 11: (_, _, _, _, _, _, _, _, _, _, _, _) = source; break;
+                    case 12: (_, _, _, _, _, _, _, _, _, _, _, _, _) = source; break;
+                    case 13: (_, _, _, _, _, _, _, _, _, _, _, _, _, _) = source; break;
+                    case 14: (_, _, _, _, _, _, _, _, _, _, _, _, _, _, _) = source; break;
+                    case 15: (_, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _) = source; break;
+                    case 16: (_, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _) = source; break;
+                    default: throw new ArgumentOutOfRangeException(nameof(width), width, null);
+                }
+            });
+
+            Assert.That(e.ParamName, Is.EqualTo("source"));
+        }
+
+        static readonly IEnumerable<ITestCaseData> DeconstructCases =
+            from w in Enumerable.Range(2, 16 - 1)
+            from c in Enumerable.Range(0, w + 2)
+            select new TestCaseData(w, c);
+
+        [TestCaseSource(nameof(DeconstructCases))]
+        public void Deconstruct(int width, int count)
+        {
+            var xs = Enumerable.Range(1, count);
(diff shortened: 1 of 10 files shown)


atifaziz:
810ebc0b08fcaf8a7934cb171d66fe15f2d6a73a is designed to pick up additional testing added with d2e2d6defa0e1f6598d1e902056ae85f017b3c2b (“Assert no iteration past end in sequence test wrapper”).

fsateler:
> Where immaterial, the count can be discarded.

I'm still thinking about this, but I'm not convinced that the "unknown count, but I can handle many options" case is common enough to warrant being the default. It smells to me that this would lead to "unchecked return"-type errors from when I used C. Users will just ignore the count and (un)happily use a `default` (very likely `null`) value.

atifaziz:
> > Where immaterial, the count can be discarded.
> 
> I'm still thinking about this, but I'm not convinced that the "unknown count, but I can handle many options" case is common enough to warrant being the default. It smells to me that this would lead to "unchecked return"-type errors from when I used C. Users will just ignore the count and (un)happily use a `default` (very likely `null`) value.

You can't prevent people from being sloppy. The fact that you are _forced_ to deconstruct with the count should make someone think twice and if they still choose to discard it then they deserve the bugs that'll bite them. 😇 In C, your intention to discard or ignore a returned value wasn't required to be explicitly expressed or captured in code whereas here you would be forced to discard (like [`ignore`](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/operators.ignore%5b't%5d-function-%5bfsharp%5d) in F#).

The trickier bit here is not so much discarding the count but not _deconstructing & discarding an extra element_, an in the 3rd tuple element in `(var x, var y, _, 2)`, to ensure that the sequence is not longer than expected (if that's important). The count is the _actual count of elements deconstructed_ out of those _requested by the pattern_. I'll agree it's trickier to write the correct pattern here but throwing an exception feels too prescriptive and takes away a whole slew of choices. It's fair to throw an exception if deconstruction cannot be fulfilled or behaviour would be undefined under some conditions but deconstruction as defined here will always succeed. The caller then gets to decide what success actually means for the application based on the results of deconstruction.

atifaziz:
> I'm still thinking about this

Do you think we can wrap this up soon?

atifaziz:
@leandromoh Would you still like to remain on board as a reviewer? I'm assuming you're very busy and won't have a chance to review since I haven't heard from you at all so far?

fsateler:
Sorry for the delay.

> You can't prevent people from being sloppy

Ah, but you can, by removing the possibility of sloppiness :wink: . I'm a big fan of [Pit of Success](https://blogs.msdn.microsoft.com/brada/2003/10/02/the-pit-of-success/) interfaces.

> It's fair to throw an exception if deconstruction cannot be fulfilled or behaviour would be undefined under some conditions but deconstruction as defined here will always succeed.

But this would make it super easy to fall into `NullReferenceException`.

> throwing an exception feels too prescriptive and takes away a whole slew of choices

Not so:

```csharp
var (a,b,c,d) = source.Take(4).Pad(myDefaultValue).AsDeconstructible();
```

Although a bit more verbose, it allows unpacking unknown lengths. Perhaps we can even ship an overload that does that automatically:

```csharp
public static DeconstructibleEnumerable AsDeconstructible<T>(this IEnumerable<T> source, T defaultValue) =>
     source.Take(4).Pad(defaultValue).AsDeconstructible();
```

atifaziz:
Closing because of [decision to not move forward](https://github.com/morelinq/MoreLINQ/issues/538#issuecomment-518983110) with #538.

@fsateler Thanks for all your help to review this.

Issue #326, opened 2017-06-22
Multiple aggregations in a single iteration


atifaziz:
The idea I wish to propose/discuss is best illustrated through examples. Suppose you want to count _and_ sum up the numbers in a sequence. With LINQ's [`Count`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.count?view=netframework-4.5) and [`Sum`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sum?view=netframework-4.5), you'd do it like this:

```c#
var xs = Enumerable.Range(10, 10);
Console.WriteLine((Count: xs.Count(), Sum: xs.Sum()));
```

Suppose now you want to also get the average:

```c#
Console.WriteLine((Count  : xs.Count(),
                   Sum    : xs.Sum(),
                   Average: xs.Average()));
```

Unfortunately, using these methods individually means the source ends up being iterated multiple times, once per aggregation. You could use LINQ's [`Aggregate`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.aggregate?view=netframework-4.5) to do multiple aggregations while iterating through the source once:

```c#
Console.WriteLine(
    xs.Aggregate(
        (Count: 0, Sum: 0),
        (a, x) => (a.Count + 1, a.Sum + x)));

Console.WriteLine(
    xs.Aggregate(
        (Count: 0, Sum: 0),
        (a, x) => (a.Count + 1, a.Sum + x),
        a => (Count: a.Count,
              Sum  : a.Sum,
              Avg  : (double) a.Sum / a.Count))); // average for free
```

However, as you increase the number of aggregations, some of which may not be trivial, the code becomes less readable and the seeds & accumulators get pulled further apart. By adding `Aggregate` overloads that allow multiple seeds, accumulators and a projector of all results like below:

```c#
public static TResult Aggregate<T, TState1, TState2, TResult>(
    this IEnumerable<T> source,
    TState1 seed1, Func<TState1, T, TState1> accumulator1,
    TState2 seed2, Func<TState2, T, TState2> accumulator2,
    Func<TState1, TState2, TResult> resultSelector) =>
    source.Aggregate(
        (seed1, seed2),
        (s, e) => (accumulator1(s.Item1, e), accumulator2(s.Item2, e)),
        s => resultSelector(s.Item1, s.Item2));
```

one can write something more natural:

```c#
Console.WriteLine(
    xs.Aggregate(
[...]

PR #599, opened 2019-06-24
Aggregate with multiple accumulators

atifaziz:
This PR implements the design where accumulators are `IObservable<>` queries, [as laid out in my comment](https://github.com/morelinq/MoreLINQ/issues/326#issuecomment-311709607) in #326. It was chosen on the following basis & benefits that were already identified:

- It adds no new types
- No aggregators need to be defined; just use what's in Rx
- There is less testing to do because of the previous point
- No conflicts with projects using Rx

Since the final design didn't reach a conclusion (almost 2 years since), I've added `Aggregate` to the experimental namespace with the intention to release it for field usage and hoping that feedback will help identify any design oversight and usage issues.


atifaziz:
I have also added the version of the API proposed in the [opening comment](https://github.com/morelinq/MoreLINQ/issues/326#issue-237975385) of #326 directly under the `MoreLinq` namespace while the one based on accumulator/aggregator queries (using obversables) remains under `MoreLinq.Experimental`

I'd love to get some help or hear clever ideas on how to test code for 16 overloads and all possible combinations of things that can go wrong. I'd hate to resort to T4 generation for it. Currently, the tests are more like usage demos though still better than nothing.

atifaziz:
@fsateler, @leandromoh: I will merge this on Monday. If you have time to review before then great otherwise I don't want this dragging out further. If you don't think you'll have time or you're happy to pass the opportunity then either is fine but just say so. The silence is deafening otherwise. 😉

MoreLinq/Assignment.cs
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq
+{
+    static class Assignment
+    {
+        public static T Set<T>(ref T slot, T newValue)
```
fsateler: This (undocumented) method serves to elide one line in each call location. Is it necessary?

MoreLinq/Delegating.cs
```diff
@@ -0,0 +1,90 @@
...
+        public static IObservable<T> Observable<T>(Func<IObserver<T>, IDisposable> delegatee) =>
+            new DelegatingObservable<T>(delegatee);
+
+        public static IObserver<T> Observer<T>(Action<T> onNext,
+                                               Action<Exception> onError = null,
+                                               Action onCompleted = null) =>
+            new DelegatingObserver<T>(onNext, onError, onCompleted);
+    }
+
+    partial class DelegatingDisposable : IDisposable
```
fsateler: This class doesn't appear to need to be partial.

And it should probably be sealed. Same for the other Delegating classes

MoreLinq/Reactive/Subject.cs
```diff
@@ -0,0 +1,92 @@
...
+// limitations under the License.
+#endregion
+
+namespace MoreLinq.Reactive
+{
+    using System;
+    using System.Collections.Generic;
+    using Delegate = Delegating.Delegate;
+
+    sealed class Subject<T> : IObservable<T>, IObserver<T>
```
fsateler: I can't say I'm thrilled at having yet another Subject (and IObservable helpers). I don't see a way out without depending on Rx though.

fsateler:
> I will merge this on Monday. If you have time to review before then great otherwise I don't want this dragging out further.

I agree work should not stall for lack of review. Sorry for posting my review the last day.

MoreLinq/Aggregate.g.tt
```diff
@@ -0,0 +1,201 @@
...
+    {
+#>
+
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+
+    static partial class MoreEnumerable
+    {
```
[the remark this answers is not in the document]
atifaziz: Yes because the same template is used to generate `Aggregate` implementations based on the two designs, where one design lives the experimental namespace and the other not.

MoreLinq/Aggregate.g.tt
```diff
@@ -0,0 +1,201 @@
...
+        /// <para>
+        /// Each accumulator argument is a function that receives an
+        /// <see cref="IObservable{T}"/> that produces the items in the
+        /// <paramref name="source"/> sequence (in original order) and that
+        /// must return an <see cref="IObservable{T}"/> that produces a single
+        /// aggregate on completion (when <see cref="IObserver{T}.OnCompleted"/>
+        /// is called). An error is raised at run-time if the
+        /// <see cref="IObserver{T}"/> returned by an accumulator function
+        /// produces no result or produces more than a single result.
+        /// </para>
```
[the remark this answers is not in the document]
atifaziz: The only and exclusive consumer is `Aggregate` so I think the note will distract and confuse rather than add any value.

MoreLinq/Reactive/Subject.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        public void OnError(Exception error)
+        {
+            if (IsMuted)
+                return;
+
+            _error = error;
+
+            if (!(Assignment.Set(ref _observers, default) is List<IObserver<T>> observers))
```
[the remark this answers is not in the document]
atifaziz: > Why do we need to set _observers back to null?

Because no further events can occur after `OnError` or `OnCompleted`. Any future subscription after a subject has reached a concluded state can be communicated immediately on subscription without needing to keep the obvserver. Like Rx, protocol violations are not reported as exceptions.

> but maybe it would be nice to add a comment why?

Okay, done with 261e36037e49480c2b2f73d42721d1940b064f89.

MoreLinq/Aggregate.g.tt
```diff
@@ -0,0 +1,201 @@
...
+                a<#= arg.Number #> = accumulator<#= arg.Number #>(a<#= arg.Number #>, item);
+<#      } #>
+            }
+
+            return resultSelector(<#= string.Join(", ", from x in o.Arguments select "a" + x.Number) #>);
+        }
+    }
+}
+
+namespace MoreLinq.Experimental
```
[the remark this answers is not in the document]
atifaziz: Yes, speaking strictly, but since _should_ doesn't mean _must_, 😉 I would rather leave it here in order to avoid maintaining duplicate templates. I am hoping the design based on observables won't live in the experimental namespace for too many versions.

MoreLinq/Assignment.cs
atifaziz: It allows to compress three lines into one in a reliable way. Nothing is ever necessary but it helped me to encode a pattern in a reusable manner because it felt dirty to have the same thing twice, once in `OnError` and another time in `OnCompleted`. That said, I found that there was a lot more duplication between the two and so I decided to consolidate them entirely and now `Assignment.Set` is gone with 850b16f1aadc35b90f716909abe0225dc80934fe.

MoreLinq/Delegating.cs
atifaziz: These were copied verbatim from [Delegating](https://github.com/atifaziz/Delegating) and I try to maintain the original sources by as much as possible. Keeping them as partial definitions is harmless but they cannot be sealed without modifying the source code or sealing sub-classes (which seems like an overkill). Since I don't expect these to change and it's a small maintenance cost to keep the code in sync if they do change, I have gone and made these changes in 27e20d1136b1ca21878d9729474f23379fd0b6bf and 099a642622442997472d8a8f2566a8a27adbf1bc.

MoreLinq/Reactive/Subject.cs
atifaziz: But then again, this _is_ the way out. :) Adding a dependency on Rx, especially on someone who doesn't use the experimental version, seems rude.

MoreLinq/Reactive/Subject.cs
```diff
@@ -0,0 +1,92 @@
...
+        bool HasObservers => (_observers?.Count ?? 0) > 0;
+        List<IObserver<T>> Observers => _observers ?? (_observers = new List<IObserver<T>>());
+
+        bool IsMuted => _completed || _error != null;
+
+        public IDisposable Subscribe(IObserver<T> observer)
+        {
+            if (observer == null) throw new ArgumentNullException(nameof(observer));
+
+            if (_error is Exception error)
```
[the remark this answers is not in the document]
atifaziz: Not really that weird if you see this as _pattern matching_ and this is becoming a bit of a habit. I also see `foo is null` as a surer way of writing `foo == null` (though it shouldn't matter for well-behaved classes) but writing `!(_error is null)` seems less readable than `_error is Exception error`; while I find it dangerous to mention the type again, one can introduce a variable that doesn't require nullability (`?`). Anyway, we can keep it simple; reverted in b0decf7bbc89408b660133c93ae4d0bd00fe1021.

atifaziz:
> I agree work should not stall for lack of review. Sorry for posting my review the last day.

Appreciate the review. Let me know if with a follow-up “approval” review (ideally today and sorry to rush) if the latest changes are good to go. Thanks!

atifaziz:
⏳ Will merge tomorrow on the assumption that most concerns from the review have been addressed.

fsateler:
Sorry I didn't manage to take a second look. Thanks for your work

atifaziz:
> Sorry I didn't manage to take a second look. Thanks for your work

No worries. Thanks for your first review and didn't mean to pressure you but I figured our schedules are simply not lining up & I needed to give myself some deadline in order to make progress. If you have any issues (hopefully non-critical) with the implementation still, feel free to open them as new issues.

PR #638, opened 2019-11-03
UnionBy extension method

salmanbabri:
I have added a UnionBy extension method. The purpose is to remove the necessity of creating IEqualityComparer when using Union method.

### DIFF
--- a/MoreLinq.Test/UnionByTest.cs
+++ b/MoreLinq.Test/UnionByTest.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using NUnit.Framework;
+
+namespace MoreLinq.Test
+{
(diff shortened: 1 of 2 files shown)


salmanbabri:
> 
> 
> Thanks but how is this any different than doing [`Concat`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.concat) + [`DistinctBy`](https://morelinq.github.io/3.2/ref/api/html/Overload_MoreLinq_MoreEnumerable_DistinctBy.htm)?

Well, same could be said about (Union = Concat + Distinct) in System.Linq. It just acts as a syntactic sugar, by combing both concepts into a single method.

PR #639, opened 2019-11-04
Zip refactoring

Orace:
- introduce CustomZip with separate behavior by source (internal)
- use CustomZip for EquiZip, ZipLongest and ZipShortest
- add ValueTuple overloads
- use t4 file to have up to 8 input sequences.

### DIFF
--- a/MoreLinq.Test/WindowLeftTest.cs
+++ b/MoreLinq.Test/WindowLeftTest.cs
@@ -30,6 +30,51 @@ namespace MoreLinq.Test
         }
 
         [Test]
+        public void WindowModifiedBeforeMoveNextDoesNotAffectNextWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.WindowLeft(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            window1[1] = -1;
+            e.MoveNext();
+            var window2 = e.Current;
+
+            Assert.That(window2[0], Is.EqualTo(1));
+        }
+
+        [Test]
+        public void WindowModifiedAfterMoveNextDoesNotAffectNextWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.WindowLeft(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            e.MoveNext();
+            window1[1] = -1;
+            var window2 = e.Current;
+
+            Assert.That(window2[0], Is.EqualTo(1));
+        }
+
+        [Test]
+        public void WindowModifiedDoesNotAffectPreviousWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.WindowLeft(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            e.MoveNext();
+            var window2 = e.Current;
+            window2[0] = -1;
+
+            Assert.That(window1[1], Is.EqualTo(1));
+        }
+
+        [Test]
         public void WindowLeftWithNegativeWindowSize()
         {
             AssertThrowsArgument.OutOfRangeException("size", () =>
--- a/MoreLinq.Test/WindowRightTest.cs
+++ b/MoreLinq.Test/WindowRightTest.cs
@@ -30,6 +30,51 @@ namespace MoreLinq.Test
         }
 
         [Test]
+        public void WindowModifiedBeforeMoveNextDoesNotAffectNextWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.WindowRight(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            window1[0] = -1;
+            e.MoveNext();
+            var window2 = e.Current;
+
+            Assert.That(window2[0], Is.EqualTo(0));
+        }
+
+        [Test]
+        public void WindowModifiedAfterMoveNextDoesNotAffectNextWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.WindowRight(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            e.MoveNext();
+            window1[0] = -1;
+            var window2 = e.Current;
+
+            Assert.That(window2[0], Is.EqualTo(0));
+        }
+
+        [Test]
+        public void WindowModifiedDoesNotAffectPreviousWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.WindowRight(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            e.MoveNext();
+            var window2 = e.Current;
+            window2[0] = -1;
+
+            Assert.That(window1[0], Is.EqualTo(0));
+        }
+
+        [Test]
         public void WindowRightWithNegativeWindowSize()
         {
             AssertThrowsArgument.OutOfRangeException("size", () =>
--- a/MoreLinq.Test/WindowTest.cs
+++ b/MoreLinq.Test/WindowTest.cs
@@ -34,6 +34,51 @@ namespace MoreLinq.Test
             new BreakingSequence<int>().Window(1);
         }
 
+        [Test]
+        public void WindowModifiedBeforeMoveNextDoesNotAffectNextWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.Window(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            window1[1] = -1;
+            e.MoveNext();
+            var window2 = e.Current;
+
+            Assert.That(window2[0], Is.EqualTo(1));
+        }
+
+        [Test]
+        public void WindowModifiedAfterMoveNextDoesNotAffectNextWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.Window(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            e.MoveNext();
+            window1[1] = -1;
+            var window2 = e.Current;
+
+            Assert.That(window2[0], Is.EqualTo(1));
+        }
+
+        [Test]
+        public void WindowModifiedDoesNotAffectPreviousWindow()
+        {
+            var sequence = Enumerable.Range(0, 3);
+            using var e = sequence.Window(2).GetEnumerator();
+
+            e.MoveNext();
+            var window1 = e.Current;
+            e.MoveNext();
+            var window2 = e.Current;
+            window2[0] = -1;
+
+            Assert.That(window1[1], Is.EqualTo(1));
+        }
+
         /// <summary>
         /// Verify that a negative window size results in an exception
         /// </summary>
@@ -42,7 +87,7 @@ namespace MoreLinq.Test
         {
             var sequence = Enumerable.Repeat(1, 10);
 
-            AssertThrowsArgument.OutOfRangeException("size",() =>
+            AssertThrowsArgument.OutOfRangeException("size", () =>
                 sequence.Window(-5));
         }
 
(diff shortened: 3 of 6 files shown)


Orace:
I was thinking of removing the ZipSourceConfiguration part (it doesn't handle all the case) and use (3) differents implementation of a ZipIteration strategy that will receive the iterators only once.

The final loop will then look like

```c#
while(iterationStrategy.MoveNext())
{
  yield return resultSelector(...);
}
```

MoreLinq/Zip.g.tt
```diff
@@ -0,0 +1,379 @@
...
+<# foreach (var arg in o.Arguments) { #>
+                using var e<#= arg.Number #> = <#= arg.Ordinal #>Source.GetEnumerator();
+<#} #>
+
+                for (;;)
+                {
+                    if (e<#= o.Arguments.First().Number #>.MoveNext())
+                    {
+                        if (<# foreach (var arg in o.Arguments.Skip(1)) { #>e<#= arg.Number #>.MoveNext()<#= arg.Num < o.Count ? " && " : "" #><#}#>)
+                        {
```
atifaziz: Braces on these single statement blocks seem excessive. I would remove them and there is little risk in doing so on the assumption that our tests would catch any accidental change (and if they don't then it highlights missing coverage).

MoreLinq/Zip.g.tt
```diff
@@ -0,0 +1,379 @@
...
+        {
+            return ZipShortest(
+<# foreach (var arg in o.Arguments) { #>
+                <#= arg.Ordinal #>Source,
+<#}#>
+                ValueTuple.Create);
+        }
+
+<#  } #>
+        static class ZipHelper
```
atifaziz: This is static or non-generated code so let's move it into a separate file called `Zip.cs`

MoreLinq/Zip.g.tt
```diff
@@ -0,0 +1,379 @@
...
+                    e<#= arg.Number #> = <#= arg.Ordinal #>Source.GetEnumerator();
+<#} #>
+
+<# foreach (var arg in o.Arguments) { #>
+                    var v<#= arg.Number #> = default(T<#= arg.Number #>);
+<#} #>
+
+                    while (
+<# foreach (var arg in o.Arguments) { #>
+                        ZipHelper.MoveNextOrDefault<T<#= arg.Number #>>(ref e<#= arg.Number #>, ref v<#= arg.Number #>)<#= arg.Num < o.Count ? " |" : ")" #>
```
atifaziz: You want conditional OR (`||`) instead of the bitwise OR `|` here…

```suggestion
                        ZipHelper.MoveNextOrDefault<T<#= arg.Number #>>(ref e<#= arg.Number #>, ref v<#= arg.Number #>)<#= arg.Num < o.Count ? " ||" : ")" #>
```

It bothers me that a test did not spot this error in refactoring. It means there's coverage missing that should be fixed.

Orace: I use the [logical OR | operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#logical-or-operator-) here and not the `||` one because I need to evaluate all the operands. (`||` will stop evaluation if it finds a true).

Orace: I need to evaluate all the operands here, `||` will stop evaluation on the first true.
[Logical OR operator |](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/boolean-logical-operators#logical-or-operator-) evaluates both operands even if the left-hand operand evaluates to true.

If you put `||` the test fails.

Orace:
> This is not something that's been agreed on **so kindly revert to the previous arity**. This can be discussed in a new issue and a separate PR can follow if and when this one is merged. A refactoring PR shouldn't add to the public surface area of the API unless it's a refactoring of the design (but here it's just the implementation).

Sorry but I need clarification here.
I'm not sure if you want me to revert anything here.
There are tt files in the first commit of this PR.

Do you want me to remove the tt files and keep the generated code ?

MoreLinq/Zip.g.tt
Orace: With the actual implementation, `ZipLongest` will consume all the sequence and dispose them as their end are reached. And that is actually tested.

atifaziz:
> Do you want me to remove the tt files and keep the generated code ?

Not at all. I just don't want new overloads as part of this PR because this is a refactoring of internals that shouldn't be expanding the public API surface as part of the same commit; it just belongs in a separate commit.

Orace:
> I just don't want new overloads as part of this PR

Thank you for the clarifications. I will reduce the overload in this PR and open a new one with the full stuff.

atifaziz:
Is there a reason for the split in e8925dc2187f5475a1cdf3bdd498645846fac2ed?

Orace:
> Is there a reason for the split in [e8925dc](https://github.com/morelinq/MoreLINQ/commit/e8925dc2187f5475a1cdf3bdd498645846fac2ed)?

Readability of the .g.tt file.
And to respect the (de facto) convention: file name match method name.

MoreLinq/Zip.g.tt
atifaziz: Thanks for correcting me and like the neat trick.

MoreLinq/EquiZip.g.tt
```diff
@@ -0,0 +1,130 @@
...
+    using System;
+    using System.Collections.Generic;
+
+    static partial class MoreEnumerable
+    {
+<#  foreach (var o in overloads)
+    {
+#>
+        /// <summary>
+        /// Returns a projection of tuples, where each tuple contains the N-th
```
atifaziz: Can we add variation to the summary, as in “Returns a projection of _N_-element tuples, …” where N is the count? Otherwise summaries of all overloads read the same.

MoreLinq/EquiZip.g.tt
```diff
@@ -0,0 +1,130 @@
...
+        /// element from each of the input sequences. An exception is thrown
+        /// if the input sequences are of different lengths.
+        /// </summary>
+<# foreach (var arg in o.Arguments) { #>
+        /// <typeparam name="<#=arg.Type#>">Type of elements in <#=arg.Name#> input sequence.</typeparam>
```
atifaziz: Spacing here and elsewhere helps readability and consistency:

```suggestion
        /// <typeparam name="<#= arg.Type #>">Type of elements in <#= arg.Name #> input sequence.</typeparam>
```

Ditto elsewhere.

MoreLinq/EquiZip.g.tt
```diff
@@ -0,0 +1,130 @@
...
+<#  } #>
+        /// <typeparam name="TResult">Type of elements in result sequence.</typeparam>
+<# foreach (var arg in o.Arguments) { #>
+        /// <param name="<#=arg.Name#>">The <#=arg.Ordinal#> source sequence.</param>
+<#  } #>
+        /// <param name="resultSelector">
+        /// Function to apply to each tuple of elements.</param>
```
atifaziz:
```suggestion
        /// Function to apply to elements combined from each sequence.</param>
```

MoreLinq/ZipLongestHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq
+{
+    using System.Collections.Generic;
+
+    static class ZipLongestHelper
```
atifaziz: This has nothing to with `ZipLongest` specifically even though that's where it's used. The only method that it contains is fairly general-purpose for any enumerator so let's reflect that in the name:

```suggestion
    static class Enumerator
```

MoreLinq/ZipLongestHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <typeparam name="T">The type of element that are enumerated.</typeparam>
+        /// <param name="enumerator">The enumerator to iterate or dispose.</param>
+        /// <param name="value">The new current value of <paramref name="enumerator"/> or
+        /// <c>default</c> if <paramref name="enumerator"/> has no more element.
+        /// </param>
+        /// <remarks>
+        /// Because <paramref name="enumerator"/> and <paramref name="value"/> may both be modified
+        /// they are both passed by reference.
+        /// </remarks>
+        /// <returns>A <c>bool</c> value indicating if the enumerator has moved to the next element.</returns>
```
atifaziz: Blank line after for readability:

```suggestion
        /// <returns>A <c>bool</c> value indicating if the enumerator has moved to the next element.</returns>

```

MoreLinq/ZipLongestHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+        public static bool MoveNextOrDispose<T>(ref IEnumerator<T> enumerator, ref T value)
```
atifaziz: Rename to simply `Read` (advance + retrieve next = read). The by-reference passing (and your detailed comment) make it fairly obvious at the call site that the method has side-effects on its inputs. I would however replace the `ref` modifier on `value` to `out`.

Nitpick: For sake of consistency, let's rename `value` to `item` as _item_ and _element_ are the more common ways of referring to the things in a sequence.

MoreLinq/ZipLongestHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+#endregion
+
+namespace MoreLinq
+{
+    using System.Collections.Generic;
+
+    static class ZipLongestHelper
+    {
+        /// <summary>
+        /// Move the <paramref name="enumerator"/> to the next position and put the
```
atifaziz: Use `<para>` that's designed to separate paragraphs.

MoreLinq/EquiZip.g.tt
Orace: It should read "where the N-th tuple contains the N-th element from each of the input sequences."
It's the original summary I believe.

MoreLinq/ZipLongestHelper.cs
atifaziz: > Rename to simply `Read` (advance + retrieve next = read). The by-reference passing (and your detailed comment) make it fairly obvious at the call site that the method has side-effects on its inputs. I would however replace the `ref` modifier on `value` to `out`.

I don't see this one addressed unless you have some objection?

MoreLinq/Enumerator.cs
```diff
@@ -0,0 +1,65 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
```
atifaziz:
```suggestion
// Copyright (c) 2019 Pierre Lando. All rights reserved.
```

README.md
```diff
@@ -218,9 +218,11 @@ This method has 2 overloads.
 
 ### EquiZip
 
-Returns a projection of tuples, where each tuple contains the N-th
-element from each of the argument sequences. An exception is thrown
-if the input sequences are of different lengths.
+Returns a sequence of projections, each projection is build from K elements.
```
atifaziz: I'm afraid that these new descriptions read much more complicated by introducing K into the picture in addition to N already. Do these really need revision as part of this _refactoring_?

MoreLinq/ZipLongestHelper.cs
Orace: As explained in the method description, If `enumerator` is null, `value` remain unchanged (this behavior is useful for upcoming padding with last value functionality). This can't be achieved with an out parameter.

README.md
Orace: Are you ok with "multiple" instead of K ?
I'm not ok with 'tuples' in the original description, it's misleading since no tuple object are created, and the result selector doesn't accept a tuple as input argument.

atifaziz: > Are you ok with "multiple" instead of K ?

Definitely better than K.

MoreLinq/ZipLongestHelper.cs
Orace: Since default is anyway used at the end of the enumerator it will be `out`

Orace: And I will rename it `TryRead` since it's a try pattern method (run a success bool)

PR #640, opened 2019-11-05
Zip while

Orace:
Depend on #639

ZipWhile allow the user to made custom length Zip.

An example of use is when the user wants to create a zip from a main sequence and a secondary sequence and expect the result to be the length of the main sequence regardless of the length of the secondary sequence.

```C#
  // continue zip until first source (users) is consumed
  var usersAndEmail = users.ZipWhile(emails, liveSources => liveSources.Contains(1));
```

If the secondary sequence is longest that the main sequence, the enumeration of the secondary sequence does not reach end.
If the secondary sequence is shortest that the main sequence, the default value is used for padding.


Orace:
My first thought about the predicate was something like that:

```C#
  var usersAndEmail = users.ZipWhile(emails, aliveSources => aliveSources .Contains(users));
```

But I get a warning on `Contains(users)` because it's used while it may have been disposed.

An other option to avoid 1-based index is to expose an object with as many _status_ property as they are enumerable.

```C#
  var usersAndEmail = users.ZipWhile(emails, status => status.IsSource1Alive);
  var usersAndEmail = users.ZipWhile(emails, status => status.HasSource1Data);
  var usersAndEmail = users.ZipWhile(emails, status => !status.IsSource1Consumed);
```
I'm open for propositions here.

atifaziz:
I don't have enough of the context of your problem to make a call here but something tells me that this is something for Rx and observables (reactive sources), not sequences (interactive sources).

atifaziz:
I'm closing this since there's been [no follow-up to provide additional context](https://github.com/morelinq/MoreLINQ/pull/640#issuecomment-550006703), but reconsider if there's renewed interest to continue the conversation on motivation for this.

PR #641, opened 2019-11-05
Simplify lag implementation

Orace:


Orace:
The proposed implementation is not optimised in terms of branching.

PR #644, opened 2019-11-05
SkipLast refactoring

### LINKED ISSUE #626: SkipLast is not optimal
~SkipLast use CountDown and therefor store all the elements.~
~We can manage to store only the number of element to skip.~
~Store the full sequence could and should be avoided.~

Orace:
Fix #626

- Add a test for SkipLast for ICollection input sequence
- Add a custom implementation for SkipLast that avoid object creation for each item

### DIFF
--- a/MoreLinq/SkipLast.cs
+++ b/MoreLinq/SkipLast.cs
@@ -40,12 +40,28 @@ namespace MoreLinq
             if (count < 1)
                 return source;
 
-            return
-                source.TryGetCollectionCount() is int collectionCount
-                ? source.Take(collectionCount - count)
-                : source.CountDown(count, (e, cd) => (Element: e, Countdown: cd ))
-                        .TakeWhile(e => e.Countdown == null)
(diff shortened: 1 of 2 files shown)


MoreLinq/SkipLast.cs
```diff
@@ -40,12 +40,28 @@ public static IEnumerable<T> SkipLast<T>(this IEnumerable<T> source, int count)
...
+            return _(); IEnumerable<T> _()
+            {
+                var queue = new Queue<T>(count);
+                using var enumerator = source.GetEnumerator();
+
+                while (count-- > 0 && enumerator.MoveNext())
+                {
+                    queue.Enqueue(enumerator.Current);
+                }
+
```
Orace: MoveNext is called after the sequence end if it is too short
Tests should have detected that.

atifaziz:
Closing as this is no longer an issue (see 59818a3f0af9388029d5ffcb24c7171d5783184f in #759).

Issue #645, opened 2019-11-05
`Subsets(0)` should return an empty set


Orace:
The 0-Subsets of a sequence should be a one item sequence, and this item should be an empty list.
Indeed there is always one subset of length 0 : the empty set.

```C#
  Enumerable.Range(1, 4).Subsets(0);  // [] instead of [[]]
```

It returns an empty sequence instead.

A call to `Subsets()` return a sequence that contains an empty list, as expected.

atifaziz:
Thanks for reporting this. Unfortunately, since this is a breaking change and `Subsets` has been around for a long time, I am currently inclined to defer this to the next major version update.

PR #646, opened 2019-11-05
Fix `Subsets` so it returns an empty set when k = 0

Orace:
This PR addresses #645


MoreLinq/Subsets.cs
```diff
@@ -108,7 +108,9 @@ public static IEnumerable<IList<T>> Subsets<T>(this IEnumerable<T> sequence, int
             // preconditions. This however, needs to be carefully considered - and perhaps
             // may change after further thought and review.
 
-            return new SubsetGenerator<T>(sequence, subsetSize);
+            return subsetSize == 0
+                ? (IEnumerable<IList<T>>) new[] {new List<T>()}
+                : new SubsetGenerator<T>(sequence, subsetSize);
```
Crossbow78: Since you're handling the 0 case at this level, would it make sense to restrict the usage of the `SubsetGenerator` to only allow `subsetSize > 0`?
This argument validation takes place in its constructor at line 216.

Orace: Since the current implementation of 'SubsetGenerator' doesn't return the expected result for k = 0, yes you are totally right.
The less lazy but more risky option is to fix SubsetGenerator.

MoreLinq.Test/SubsetTest.cs
```diff
@@ -1,3 +1,5 @@
+using System.Collections.Generic;
```
atifaziz: Remove import as it isn't used:

```suggestion
```

MoreLinq.Test/SubsetTest.cs
```diff
@@ -116,6 +118,19 @@ public void TestAllSubsetsExpectedResults()
...
+        /// See issue #645
+        /// </summary>
+        [Test]
+        public void Test0SubsetIsEmptyList()
+        {
+            var sequence = Enumerable.Range(1, 4);
+            var actual = sequence.Subsets(0);
+            var expected = new[] {new int[0]};
+
+            CollectionAssert.AreEquivalent(expected, actual);
```
atifaziz: Seems simpler to just say: `actual.AssertSequenceEqual(new int[0]);`

MoreLinq/Subsets.cs
```diff
@@ -108,7 +108,9 @@ public static IEnumerable<IList<T>> Subsets<T>(this IEnumerable<T> sequence, int
             // preconditions. This however, needs to be carefully considered - and perhaps
             // may change after further thought and review.
 
-            return new SubsetGenerator<T>(sequence, subsetSize);
+            return subsetSize == 0
+                ? (IEnumerable<IList<T>>) new[] {new List<T>()}
```
atifaziz: Return a cached empty array as opposed to create an empty list each time. While [.NET offers ones](https://docs.microsoft.com/en-us/dotnet/api/system.array.empty?view=netcore-3.1#System_Array_Empty__1), I think we should introduce our own to avoid adding an additional targets (as it's available since .NET 4.6 and .NET Standard 1.3):

```c#
static class EmptyArray<T>
{
    public static readonly T[] Value = new T[0];
}
```

Orace: It's not an empty array. It's an array of one element. This element is an empty list.

I don't believe we can have a cache for that.

If we use a cache, obviously the list must be immutable / read-only.

In other cases, the method return an `IEnumerable<IList<T>>` where the elements are not read-only list.

MoreLinq.Test/SubsetTest.cs
Orace: `new[] {new int[0]};` not `new int[0]`.
Maybe I should keep the `var expected = ...` line and comment it.

MoreLinq/Subsets.cs
Orace: I made SubsetGenerator handle the k=0 case

MoreLinq/Subsets.cs
Orace: I made SubsetGenerator handle the k=0 case

atifaziz: I meant you can cache the _empty set_, which is a zero element array that's immutable anyway; then you could have returned another cached sequence of that. Anyway, you have changed `SubsetGenerator` to handle _k_ = 0 case so this point is mute. I think those optimizations would have been rarely used and not any better than the current implementation that returns an empty list followed by subsets as arrays for all _k_ where _k_ > 0 (which is inconsistent). I think we can add the empty set optimization in another PR.

Orace: https://github.com/morelinq/MoreLINQ/blob/07bd0861658b381ce97c8b44d3b9f2cd3c9bf769/MoreLinq/Subsets.cs#L165

`SubsetEnumerator.Current` clone the current state (an array) and return it as an `IList`. 😨

Arrays as `IList` are not read-only, _generally_ not immutable (you can change the value of an element), but are not fully mutable, you can't add or remove elements.

Indeed in the case of the empty array, we have a _de facto_ immutable `IList`, but `IsReadonly` return false.

I think this should be rethought.

My opinion is that it should return a `List`.

If we look at other methods that returns `IEnumerable<IList<T>>` or `IEnumerable<IEnumerable<T>>`,  `Subsets` is the only method that clone the current state when `Enumerator.Current` is called.

```C#
void Study(IEnumerable<IEnumerable<int>> lists)
{
    var e = lists.GetEnumerator();
    e.MoveNext();
    var current = e.Current;
    if (current is List<int>)
        Console.Write("List  ");
    else if (current is int[])
        Console.Write("Array ");
    else
        Console.Write("????  ");

    var otherCurrent = e.Current;

    Console.WriteLine(ReferenceEquals(current, otherCurrent) ? "same ref" : "new ref");
}

var r = Enumerable.Range(0, 10);
var rr = Enumerable.Repeat(r, 10);
Study(r.Batch(3));                 // Array same ref
Study(r.Permutations());           // Array same ref
Study(r.Segment(v => v % 2 == 1)); // List  same ref
Study(r.Split(v => v % 2 == 1));   // List  same ref
Study(r.Subsets(5));               // Array new ref
Study(rr.Transpose());             // Array same ref
Study(r.Window(5));                // Array same ref
```

atifaziz: Yes, some inconsistencies have slipped through the cracks over the years.

MoreLinq.Test/SubsetTest.cs
atifaziz: Addressed with 29dbde66e27725e172ea7a1e8e63cdb2cba65ca0.

atifaziz (approved):

Issue #652, opened 2019-11-06
Windows of Window(Left|Right) are not isolated from each other


Orace:
List returned by Window* methods should be isolated from each other.

With current implementation, modifying one window will modify the next windows accordingly:

```C#
var sequence = Enumerable.Repeat(0, 3);
foreach (var window in sequence.Window(2))
{
  window[1] = 1; // modify the last element

  // on second iteration window[0] is 1 and this fail
  Assert.That(window[0], Is.EqualTo(0));
}
```

atifaziz:
This is by-design. You are allowed to do what you want with the returned windows. What's important is that those windows are independent of the original sequence on which they were created such that modifying an element through the window does not modify the source.

Orace:
> those windows are independent of each other 

That is exactly what I show here: windows are not independent from each others

atifaziz:
I re-read the code and you're absolutely right. Sorry for closing this but your explanation and test in #653 didn't read clear enough. Thanks for persevering.

PR #655, opened 2019-11-06
Return isolated windows from Window* methods

Orace:
This allow tests from #653 to pass and fix #652

Issue #654, opened 2019-11-06
Window tests are incomplete


Orace:
Actually Window test doesn't cover

 - [x] enumerator disposal (in #655)
 - [x] MoveNext calls after enumerator end (in #655)
 - [x] Returned Window are isolated from each others (see #652) (in #655)
 - [x] Test Window method on known results (in #655)
 - [x] Test WindowLeft method on known results  (in #656)
 - [x] Test WindowRight method on known results (in #656)

Orace:
All those tests are now implemented in #653 but they don't pass because of #652

Orace:
Modification inside window are normal see #652 
The test should be updated accordingly.

Orace:
All those tests are now pass in #655

Orace:
Fixed with #655

Issue #630, opened 2019-11-01
Window, WindowLeft and WindowRight implementation should be shared


Orace:
This is a fork of #629 and #627.

We should introduce a private method to share the implementation among the three Window* methods.

atifaziz:
I haven't done a study of whether it would help or not but if you'd like to take a crack at this through a PR then I'd be happy to review it. If you'd rather that I first do a study and accept this with conviction (so you don't risk wasting time with work that may not get merged) then you'll have to be patient as this is not going to be a high priority for me.

PR #656, opened 2019-11-06
Improve window

Orace:


Orace:
This PR address #630 
It depends on #655 that fix #652
It also do a part of #654

Orace:
@atifaziz you can review #655 first

MoreLinq/Window.cs
```diff
@@ -85,5 +57,101 @@ public static IEnumerable<IList<TSource>> Window<TSource>(this IEnumerable<TSour
...
+                {
+                    window[i] = iter.Current;
+                    hasNext = iter.MoveNext();
+                }
+
+                // Ensure correct size on partial window cases
+                if (window.Length != i)
+                    Array.Resize(ref window, i);
+            }
+
```
Orace: i != size
We also can test for early exit if hasPartialEnd is false

MoreLinq/Window.cs
```diff
@@ -85,5 +57,101 @@ public static IEnumerable<IList<TSource>> Window<TSource>(this IEnumerable<TSour
...
+                window = new TSource[size];
+                int i;
+                for (i = 0; i < size && hasNext; i++)
+                {
+                    window[i] = iter.Current;
+                    hasNext = iter.MoveNext();
+                }
+
+                // Ensure correct size on partial window cases
+                if (window.Length != i)
```
Orace:
```suggestion
                if (i != size)
                {
                    if (hasPartialEnd)
                        Array.Resize(ref window, i);
                    else
                        yield break;
                }
```

MoreLinq/Window.cs
```diff
@@ -85,5 +57,107 @@ public static IEnumerable<IList<TSource>> Window<TSource>(this IEnumerable<TSour
...
+
+                // Ensure correct size on partial window cases
+                if (i != size)
+                {
+                    if (hasPartialEnd)
+                        Array.Resize(ref window, i);
+                    else
+                        yield break;
+                }
+                    Array.Resize(ref window, i);
```
Orace:
```suggestion
```

PR #660, opened 2019-11-07
Add test for Exclude, Lag, Lead and SkipUntil that use TestingSequences

### LINKED ISSUE #649: Missing using instruction in Exclude and Lead
Here the list of not disposed enumerators
Unit test should be added for it (maybe we can do a generic test like NullArgumentTest).

- [x] MoreLinq\Exclude.cs:48
- [x] MoreLinq\Lead.cs:68
~SkipUntil~

We should have test for them

Orace:
Fix #649

### DIFF
--- a/MoreLinq.Test/ExcludeTest.cs
+++ b/MoreLinq.Test/ExcludeTest.cs
@@ -1,3 +1,5 @@
+using System.Collections.Generic;
+
 namespace MoreLinq.Test
 {
     using NUnit.Framework;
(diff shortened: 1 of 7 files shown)


Orace:
78322c2 fix #664

Orace:
7414473  Fix #665

atifaziz:
Thanks for those tests and fixes but this PR is too big. Can you please break it up into the four separate fixes (for `Exclude`, `Lag`, `Lead` and `SkipUntil`) that it is, even if the fix is logically the same?

Orace:
Exclude call to MoveNext after sequence end is fixed here #692

Issue #664, opened 2019-11-07
SkipUntil calls MoveNext after enumeration ends


Orace:
A test using TestingSequence was added here #660 
This is linked to #649

PR #666, opened 2019-11-08
Fix SkipUntil to not call MoveNext past end of sequence

Orace:
~Fix a part of #649~
Fork of #660 

SkipUntil called MoveNext after sequence end

### DIFF
--- a/MoreLinq.Test/SkipUntilTest.cs
+++ b/MoreLinq.Test/SkipUntilTest.cs
@@ -18,6 +18,8 @@
 namespace MoreLinq.Test
 {
     using NUnit.Framework;
+    using NUnit.Framework.Interfaces;
+    using System.Collections.Generic;
 
     [TestFixture]
     public class SkipUntilTest
@@ -56,6 +58,21 @@ namespace MoreLinq.Test
             // started returning items after -1.
             var sequence = Enumerable.Range(-2, 5).SkipUntil(x => 1 / x == -1);
             sequence.AssertSequenceEqual(0, 1, 2);
+        }
+
+        public static readonly IEnumerable<ITestCaseData> TestData =
+            from e in new[]
+            {
+                new {s = new[] {1, 2, 3}, o = 1, r = new[] {2, 3}},
+                new {s = new[] {1, 2, 3}, o = 2, r = new[] {3}},
+                new {s = new[] {1, 2, 3}, o = 3, r = new int[0]}
+            }
+            select new TestCaseData(e.s, e.o).Returns(e.r);
+
+        [Test, TestCaseSource(nameof(TestData))]
+        public int[] TestSkipUntilOnKnownInput(int[] source, int min)
+        {
+            return source.AsTestingSequence().SkipUntil(v => v >= min).ToArray();
         }
     }
 }
--- a/MoreLinq.Test/SkipUntilTest.cs
+++ b/MoreLinq.Test/SkipUntilTest.cs
@@ -63,9 +63,14 @@ namespace MoreLinq.Test
         public static readonly IEnumerable<ITestCaseData> TestData =
             from e in new[]
             {
+                new {s = new int[0]     , o = 0, r = new int[0]},   // empty sequence
+                new {s = new[] {0}      , o = 0, r = new int[0]},   // one-item sequence, predicate succeed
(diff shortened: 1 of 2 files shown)


Orace:
> How is this related to #649?

It is not, sorry I copy paste to quickly. I updated the first comment..

It's a re-call MoveNext on sequence end

atifaziz:
So this is the fix for issue #664?

Orace:
> So this is the fix for issue #664?

Indeed, I forgot to mention it :confused:

MoreLinq.Test/SkipUntilTest.cs
```diff
@@ -57,5 +59,25 @@ public void SkipUntilEvaluatesPredicateLazily()
             var sequence = Enumerable.Range(-2, 5).SkipUntil(x => 1 / x == -1);
             sequence.AssertSequenceEqual(0, 1, 2);
         }
+
+        public static readonly IEnumerable<ITestCaseData> TestData =
+            from e in new[]
+            {
+                new {s = new int[0]     , o = 0, r = new int[0]},   // empty sequence
```
atifaziz: Can we use more meaningful (Pascal-case) names for the properties?

```diff
-                new {s = new int[0]     , o = 0, r = new int[0]},   // empty sequence
+                new { Source = new int[0]     , Min = 0, Expected = new int[0] },   // empty sequence
```

Also, space after `{` and before `}`.

MoreLinq/SkipUntil.cs
```diff
@@ -61,14 +61,19 @@ public static IEnumerable<TSource> SkipUntil<TSource>(this IEnumerable<TSource>
             {
                 using var iterator = source.GetEnumerator();
 
-                while (iterator.MoveNext())
+                for(;;)
```
atifaziz:
```suggestion
                for (;;)
```

MoreLinq/SkipUntil.cs
```diff
@@ -61,14 +61,19 @@ public static IEnumerable<TSource> SkipUntil<TSource>(this IEnumerable<TSource>
...
                 {
+                    if (!iterator.MoveNext())
+                        yield break;
+
                     if (predicate(iterator.Current))
                         break;
                 }
 
                 while (iterator.MoveNext())
+                {
```
atifaziz: Seems inconsistent with above so remove:

```suggestion
```

MoreLinq/SkipUntil.cs
```diff
@@ -61,14 +61,19 @@ public static IEnumerable<TSource> SkipUntil<TSource>(this IEnumerable<TSource>
...
                     yield return iterator.Current;
+                }
```
atifaziz: Seems inconsistent with above so remove:

```suggestion
```

MoreLinq/SkipUntil.cs
Orace: It also can be a. do/while loop.
Anyway, never to late to ask, is there any formating rules description somewhere ? I use resharper for auto-formating and I didn't customize the settings...
In the same fashion of questions : `for(;;)` or `while(true)`

MoreLinq/SkipUntil.cs
```diff
@@ -59,16 +59,16 @@ public static IEnumerable<TSource> SkipUntil<TSource>(this IEnumerable<TSource>
...
 
-                while (iterator.MoveNext())
+                do
                 {
-                    if (predicate(iterator.Current))
-                        break;
-                }
+                    if (!enumerator.MoveNext())
+                        yield break;
+                } while(!predicate(enumerator.Current))
```
Orace:
```suggestion
                } while(!predicate(enumerator.Current));
```

Orace:
A squash merge may reduce my shame

MoreLinq/SkipUntil.cs
atifaziz: Nothing document but hope fd32a8ad5754aea1288c277dedb927053e1c8368 helps as a starting point. Other than that, readability comes first (even if it means breaking rules), then consistency and finally general C# guidelines. Unfortunately, this means I have to switch off all auto-formatting in the IDE; prefer to write for humans, not machines.

> In the same fashion of questions : `for(;;)` or `while(true)` ?

Either is fine as long as there is as space after `for` or `while`:

https://github.com/morelinq/MoreLINQ/blob/fd32a8ad5754aea1288c277dedb927053e1c8368/.editorconfig#L51

PR #667, opened 2019-11-08
Lag : Add test

Orace:
Fix a part of #649
Fork of #660


Orace:
In unit testing pattern we should not even use TestingSequence since it tests multiple things. (Enumerator disposal, multiple enumeration, multiple MoveNext at the end of the source).

I'm agree that TestLagPassesCorrectLagValuesOffsetBy1 and  TestLagPassesCorrectLagValuesOffsetBy2 can be replaced by new cases in TestLagOnKnownInput

The _proper_ (in terms of unit testing) way to work is to split TestingSequence in 3 new _mock_ (TestingSequenceForDisposed, etc..), add one test for each of it, add relevant cases for each of it.

I don't really want to push that since it's a lot of rework..

PR #668, opened 2019-11-08
Fix Lead that's missing enumerator disposal & test

Orace:
Fix a part of #649
Fork of #660


MoreLinq.Test/LeadTest.cs
```diff
@@ -1,6 +1,25 @@
+#region License and Terms
```
atifaziz: Thanks for adding this header but it belongs in a separate PR.

Orace:
Closed. The implementation of Lead is fixed here #672

PR #672, opened 2019-11-09
Missing using instruction in Exclude and Lead

### LINKED ISSUE #649: Missing using instruction in Exclude and Lead

Orace:
This fix #649


Orace:
Wrong origin. See #673

Issue #683, opened 2019-11-10
AppVeyor build succeeds in spite of Aggregate T4 generation failing


atifaziz:
Builds on AppVeyor are succeeding in spite of [T4 code generation failing for `Aggregate.g.tt`](https://ci.appveyor.com/project/raboof/morelinq/builds/28747501#L352):

```
Generating extensions wrappers (MoreLinq\Extensions.g.cs)...Done.
Generating extensions wrappers (MoreLinq\Extensions.ToDataTable.g.cs)...Done.
cmd : dotnet t4 "C:\projects\morelinq\MoreLinq\Aggregate.g.tt"
At line:7 char:1
+ cmd /c call pack.cmd ci-$id
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (dotnet t4 "C:\p...Aggregate.g.tt":String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
dotnet t4 "C:\projects\morelinq\MoreLinq\Cartesian.g.tt"
dotnet t4 "C:\projects\morelinq\MoreLinq\Fold.g.tt"
dotnet t4 "C:\projects\morelinq\MoreLinq\ToDelimitedString.g.tt"
```

atifaziz:
This is a false negative from PowerShell since `tt.cmd` writes to the standard error stream and the writing in red plus the `NativeCommandError` threw me off.

PR #684, opened 2019-11-10
Fix AppVeyor builds to fail on error in batch

atifaziz:
This PR is to investigate and fix issue #683.


atifaziz:
Close this PR as #683 turned out to be a non-issue.

Issue #687, opened 2019-11-10
Segment calls MoveNext twice on an empty sequence


Orace:
If the first MoveNext return false, the function should return immediately.
And then the last test will be useless:

https://github.com/morelinq/MoreLINQ/blob/ae1a86396abf9ee46912027eed13232e50d5fe22/MoreLinq/Segment.cs#L115

PR #691, opened 2019-11-12
Fix Segment to not call MoveNext past end of sequence

Orace:
This PR addresses #687

### DIFF
--- a/MoreLinq.Test/SegmentTest.cs
+++ b/MoreLinq.Test/SegmentTest.cs
@@ -139,17 +139,17 @@ namespace MoreLinq.Test
             from e in new[]
             {
                 // input sequence is empty
-                new {s = Enumerable.Empty<int>(), r = Enumerable.Empty<IEnumerable<int>>()},
+                new { Source = Enumerable.Empty<int>(), Expected = Enumerable.Empty<IEnumerable<int>>() },
                 // input sequence contains only new segment start
-                new {s = Seq(0, 3, 6), r = Seq(Seq(0), Seq(3), Seq(6))},
+                new { Source = Seq(0, 3, 6),            Expected = Seq(Seq(0), Seq(3), Seq(6)) },
                 // input sequence do not contains new segment start
-                new {s = Seq(1, 2, 4, 5), r = Seq(Seq(1, 2, 4, 5))},
+                new { Source = Seq(1, 2, 4, 5),         Expected = Seq(Seq(1, 2, 4, 5)) },
                 // input sequence start with a segment start
-                new {s = Seq(0, 1, 2, 3, 4, 5), r = Seq(Seq(0, 1, 2), Seq(3, 4, 5))},
+                new { Source = Seq(0, 1, 2, 3, 4, 5),   Expected = Seq(Seq(0, 1, 2), Seq(3, 4, 5)) },
                 // input sequence do not start with a segment start
-                new {s = Seq(1, 2, 3, 4, 5), r = Seq(Seq(1, 2), Seq(3, 4, 5))}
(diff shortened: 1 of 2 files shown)


MoreLinq/Segment.cs
```diff
@@ -76,44 +76,41 @@ public static IEnumerable<IEnumerable<T>> Segment<T>(this IEnumerable<T> source,
...
                 // behavior. Segmentation always begins with
                 // the second element in the sequence.
-                if (iter.MoveNext())
-                {
-                    ++index;
-                    segment.Add(iter.Current);
-                    prevItem = iter.Current;
-                }
+                var prevItem = iter.Current;
+                var segment = new List<T> {iter.Current};
```
atifaziz:
```suggestion
                var segment = new List<T> { iter.Current };
```

MoreLinq/Segment.cs
```diff
@@ -76,44 +76,41 @@ public static IEnumerable<IEnumerable<T>> Segment<T>(this IEnumerable<T> source,
...
                     prevItem = iter.Current;
 
-                    if (!isNewSegment)
+                    if (isNewSegment)
+                    {
+                        // yield the completed segment
+                        yield return segment;
+
+                        // start a new segment...
+                        segment = new List<T> {iter.Current};
```
atifaziz:
```suggestion
                        segment = new List<T> { iter.Current };
```

MoreLinq.Test/SegmentTest.cs
```diff
@@ -1,3 +1,6 @@
+using System.Collections.Generic;
```
atifaziz: Move under namespace declaration with other imports.

MoreLinq/Segment.cs
```diff
@@ -76,44 +76,41 @@ public static IEnumerable<IEnumerable<T>> Segment<T>(this IEnumerable<T> source,
...
+                var prevItem = iter.Current;
+                var segment = new List<T> { iter.Current };
 
-                while (iter.MoveNext())
+                for (var index = 1; iter.MoveNext(); index++)
                 {
-                    ++index;
                     // check if the item represents the start of a new segment
                     var isNewSegment = newSegmentPredicate(iter.Current, prevItem, index);
                     prevItem = iter.Current;
```
atifaziz: If this (`prevItem = iter.Current;`) is moved to the bottom of the `for` loop (which seems more natural) then `isNewSegment` can be removed and its assigned expression in-lined into the condition for the following `if`, on line 98. It makes the overall code then much simpler to read and reason about.

MoreLinq.Test/SegmentTest.cs
```diff
@@ -146,5 +148,29 @@ public void VerifyCanSegmentByPrevious()
...
+                new { Source = Seq(1, 2, 4, 5),         Expected = Seq(Seq(1, 2, 4, 5)) },
+                // input sequence start with a segment start
+                new { Source = Seq(0, 1, 2, 3, 4, 5),   Expected = Seq(Seq(0, 1, 2), Seq(3, 4, 5)) },
+                // input sequence do not start with a segment start
+                new { Source = Seq(1, 2, 3, 4, 5),      Expected = Seq(Seq(1, 2), Seq(3, 4, 5)) }
+            }
+            select new TestCaseData(e.Source).Returns(e.Expected);
+
+        [Test, TestCaseSource(nameof(TestData))]
+        public IEnumerable<IEnumerable<int>> TestSegmentOnKnownInput(IEnumerable<int> source)
```
atifaziz: Simplify name to just:

```suggestion
        public IEnumerable<IEnumerable<int>> TestSegment(IEnumerable<int> source)
```

Issue #665, opened 2019-11-07
`Exclude` calls `MoveNext` after enumeration ends


Orace:
A test using TestingSequence was added here #660
This is linked to #649

PR #692, opened 2019-11-12
Fix `Exclude` to not call `MoveNext` past end of sequence

Orace:
This PR addresses #665


atifaziz (approved):
Thanks for this fix! I updated the tests to using `TestingSequence<T>`, which highlighted the issue without the need for adding a new test (as was done initially).

Issue #694, opened 2019-11-13
Interleave eagerly calls MoveNext


Orace:
From this part of the code:

https://github.com/morelinq/MoreLINQ/blob/13c8c8355de0b6ea428d6106f6f0a92be03a8955/MoreLinq/Interleave.cs#L135-L139

We can see that interleave eagerly have called MoveNext on all the enumerators from all the input sequence before yielding the first value.

This is useless for the `ImbalancedInterleaveStrategy.Skip` strategy. And this strategy is the only one accessible from the user (the strategies enum is not public).

atifaziz:
I don't think that's the case. It removes any iterator whose `MoveNext` returns `false` and updates the counters like `iterCount` accordingly:

https://github.com/morelinq/MoreLINQ/blob/13c8c8355de0b6ea428d6106f6f0a92be03a8955/MoreLinq/Interleave.cs#L122-L127

The loop that yields then uses the updated list of iterators and count.

Your test in PR #696 is [flawed](https://github.com/morelinq/MoreLINQ/pull/696#pullrequestreview-321455535).

Orace:
> I don't think that's the case. It removes any iterator whose `MoveNext` returns `false` and updates the counters like `iterCount` accordingly

The subject of this issue is not about the possible multiple calls to `MoveNext` after a sequence end.
It's about useless call to `MoveNext`.

Considering this code.

```C#
var resultSequence = sequenceA.Interleave(sequenceB);
```

**Expected behavior**
One call to `MoveNext` on an enumerator of `resultSequence` should result on only one call to `MoveNext` on **one** enumerator of the input sequences.

**Current behavior**

One call to `MoveNext` on an enumerator of `resultSequence`  result on the call to `MoveNext` to all the enumerators of all of the input sequences.

Here the test I originally committed:

https://github.com/morelinq/MoreLINQ/blob/c9aa710f68c43c0e86b7efa14e5aa8d28b0cc2de/MoreLinq.Test/InterleaveTest.cs#L38-L53 

Here what is tested: 
If you take one element from the result sequence, only sequenceA should be enumerated.

atifaziz:
Okay that makes more sense. I was focusing too much on the [loop you quoted initially](https://github.com/morelinq/MoreLINQ/issues/694#issue-522461981) but you are right, it should avoid iterating other sequences unless needed. Thanks for the clarification.

PR #696, opened 2019-11-14
Fix Interleave to prevent eager MoveNext calls

Orace:
This PR addresses #694


MoreLinq/Interleave.cs
```diff
@@ -45,127 +43,63 @@ public static partial class MoreEnumerable
...
-            return _(); IEnumerable<T> _()
-            {
-                var sequences = new[] { sequence }.Concat(otherSequences);
+            return InterleaveSkip(otherSequences.Prepend(sequence));
+        }
 
-                // produce an iterator collection for all IEnumerable<T> instancess passed to us
-                var iterators = sequences.Select(e => e.GetEnumerator()).Acquire();
-                List<IEnumerator<T>> iteratorList = null;
+        private static IEnumerable<T> InterleaveSkip<T>(this IEnumerable<IEnumerable<T>> sequences)
```
Orace:
```suggestion
        private static IEnumerable<T> InterleaveSkip<T>(IEnumerable<IEnumerable<T>> sequences)
```

atifaziz:
I fixed your test in 61e49217d74020fbd57a4ae63747fc884e1d1c80, used the origin `Interleave` implementation in b790e53103e24ea12a124efaa016ce7a9fa63f44, even merged [the simpler yield loop](https://github.com/morelinq/MoreLINQ/commit/9fc8486b9d8819690c9bc67fe41e74d4e1e2d765) with c5694791b952ce070486b70863d110df157054d1, and as you can see, [all tests pass](https://ci.appveyor.com/project/raboof/morelinq/builds/29046843):

```
Testing netcoreapp3.0 (Release)...
NUnitLite 3.12.0 (.NET Standard 2.0)
Copyright (c) 2019 Charlie Poole, Rob Prouse
Runtime Environment
   OS Version: Microsoft Windows 10.0.17763
  CLR Version: 3.0.0
Test Files
    C:/projects/morelinq/MoreLinq.Test/bin/Release/netcoreapp3.0/MoreLinq.Test.dll
Test Discovery
  Start time: 2019-11-22 10:39:37Z
    End time: 2019-11-22 10:39:37Z
    Duration: 0.355 seconds
Tests Not Run
1) Explicit : MoreLinq.Test.MemoizeTest.MemoizeIsThreadSafe
2) Ignored : MoreLinq.Test.PartialSortByTests.PartialSortByIsStable
TODO
3) Ignored : MoreLinq.Test.PartialSortTests.PartialSortByIsStable
TODO
4) Explicit : MoreLinq.Test.RandomSubsetTest.TestRandomSubsetIsUnbiased
Run Settings
    Number of Test Workers: 2
    Work Directory: C:\projects\morelinq
    Internal Trace: Off
Test Run Summary
  Overall result: Warning
  Test Count: 2165, Passed: 2161, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 4
    Skipped Tests - Ignored: 2, Explicit: 2, Other: 0
  Start time: 2019-11-22 10:39:37Z
    End time: 2019-11-22 10:39:40Z
    Duration: 2.465 seconds
Results (nunit3) saved as C:\projects\morelinq\TestResult.xml
```

Orace:
>Both your sample sequences are one element long and you just throw for the retrieval of the element in the second sequence. 

Exactly, I throw because I asked for **one** element and this element came from the first sequence. 

It throw for the retrieval of the element in the second sequence because it should not eagerly retrieve the first element of the second sequence.

atifaziz:
Right, I've reverted your originally proposed test and fix with 8a972a9acb96ce27472fdaeb92a306b183dc1c53 and will review.

MoreLinq/Interleave.cs
```diff
@@ -45,125 +43,63 @@ public static partial class MoreEnumerable
...
-                            if (!iteratorList[index].MoveNext())
-                            {
-                                // check if all iterators have been consumed and we can terminate
-                                // or if the imbalance strategy informs us that we MUST terminate
-                                if (++consumedIterators == iterCount || imbalanceStrategy == ImbalancedInterleaveStrategy.Stop)
-                                {
-                                    shouldContinue = false;
-                                    break;
-                                }
+                    if (sequence == null)
```
atifaziz: This argument check will take place when the iteration starts and not when `Interleave` is called. It would need to be moved out.

MoreLinq/Interleave.cs
```diff
@@ -45,125 +43,63 @@ public static partial class MoreEnumerable
...
-                List<IEnumerator<T>> iteratorList = null;
+        private static IEnumerable<T> InterleaveSkip<T>(this IEnumerable<IEnumerable<T>> sequences)
+        {
+            var enumerators = new LinkedList<IEnumerator<T>>();
 
-                try
+            try
+            {
+                // First pass. create enumerators.
+                foreach (var sequence in sequences)
```
atifaziz: We have [`Acquire`](https://morelinq.github.io/3.2/ref/api/html/M_MoreLinq_MoreEnumerable_Acquire__1.htm) for this so I would take advantage of it here rather than roll out a duplicate implementation that needs separate testing.

MoreLinq/Interleave.cs
Orace: Indeed. #697 drives me to this inadequate implementation.

MoreLinq/Interleave.cs
Orace: FWIU, `Acquire` do not return a `IDisposable` so `Interleave` keep the responsibility of disposing the enumerators when the result sequence enumerator is disposed.
Achieve this with `using` is not feasible since the number of `using` depend on the sequence length.
The only way I know to achieve this in a `yield` context is to use some `try {} finally {}` blocks. Because the `finally` block is called when the enumerator is disposed.


`DisposableGroup<T>` looks to achieve this, but `DisposableGroup<T>` (and `Acquire`) return an array where I use a `LinkedList`.

I have chosen to use a `LinkedList` since `Remove` operation is O(1) there.
If we use a list or an array, where the the `Remove` operation is O(N), enumerating N sequences of 1 element will coast O(N²).
I know that it is an uncommon use case (it may be less uncommon with #697), anyway this implementation manage it.

MoreLinq/Interleave.cs
Orace: By the way a test for that is missing here.

MoreLinq/Interleave.cs
atifaziz: > FWIU, `Acquire` do not return a `IDisposable`

It would return an array of enumerators that are disposable. Question is where there is something to be gained out of not calling `GetEnumerator()` on other sequences until you get to them, but that will only optimize the case of someone eventually taking fewer elements than sequences!

Orace: `Acquire` will not remove the necessity of a `try {} finally {}` pattern, and then is just an alias of `ToArray()`

atifaziz: > `Acquire` will not remove the necessity of a `try {} finally {}` (with disposal of the enumerators in the finally block), and then is just an alias of `ToArray()`.

`Acquire` _itself_ does not need to be inside a try-finally block and so it's not the same as `ToArray()`. It doesn't mean it prevents you from having one. It is, for example, perfectly safe to return the array from `Acquire` as the return value of a function:

```c#
static Stream[] OpenFiles(params string[] paths) =>
    paths.Select(p => File.OpenRead(p)).Acquire();
```

Of course, the disposal becomes the responsibility of the caller and will need a try-catch. It does prevent that.

> * If `Acquire` is used and if an error occur while _acquiring_ (the only purpose of `Acquire` is to manage this case), elements will be disposed twice, which is useless.

It is useless but it's not a problem since it is a requirement for `IDisposable.Dispose` to be idempotent ([per docs](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable.dispose#remarks)):

> If an object's [Dispose](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable.dispose) method is called more than once, the object must ignore all calls after the first one. 

> I have chosen to use a `LinkedList` since `Remove` operation is O(1) there.

That's a fine choice but I am not debating the most optimal solution first. My point with `Acquire` was that it's carefully designed and tested and so might be best to keep using it instead of putting everything into question. This PR was about fixing one problem but by trying to be cleverly optimal at the same time, it may introduce others bugs (and I believe it does)! This means it takes longer to review as the scope of change expands. Instead, I find that it helps to make it work, make it right and then make it fast.

MoreLinq/Interleave.cs
```diff
@@ -45,125 +44,63 @@ public static partial class MoreEnumerable
...
-                        }
-
-                        if (shouldContinue) // only if all iterators could be advanced
-                        {
-                            // yield the values of each iterator's current position
-                            foreach (var iterator in iteratorList)
-                                yield return iterator.Current;
-                        }
+                        yield return enumerator.Current;
+                        enumerators.AddLast(enumerator);
```
atifaziz: I believe you will leak the resource here if `yield` never returns.

Orace: Indeed. I will try to add a test for that.

MoreLinq/Interleave.cs
Orace: I just push a simple (I hope) implementation of Interleave based on the array returned by `Acquire`.
The `LinkedList` implementation is in #726

MoreLinq/Interleave.cs
```diff
@@ -80,90 +52,41 @@ static IEnumerable<T> Interleave<T>(this IEnumerable<T> sequence, ImbalancedInte
 
             return _(); IEnumerable<T> _()
             {
-                var sequences = new[] { sequence }.Concat(otherSequences);
+                var sequences = new[] {sequence}.Concat(otherSequences);
```
atifaziz:
```suggestion
                var sequences = new[] { sequence }.Concat(otherSequences);
```

MoreLinq/Interleave.cs
```diff
@@ -80,90 +52,41 @@ static IEnumerable<T> Interleave<T>(this IEnumerable<T> sequence, ImbalancedInte
...
-                                        --consumedIterators; // decrement consumer iterator count to stay in balance
-                                        break;
-                                }
+                            var enumerator = enumerators[i];
+                            if (enumerator == null)
+                                continue;
 
+                            if (!enumerator.MoveNext())
+                            {
+                                enumerator.Dispose();
```
atifaziz: Swap with next line.

MoreLinq.Test/InterleaveTest.cs
```diff
@@ -35,6 +36,41 @@ public void TestInterleaveIsLazy()
             new BreakingSequence<int>().Interleave(new BreakingSequence<int>());
         }
 
+        /// <summary>
+        /// Verify that Interleave early throw ArgumentNullException when an element
+        /// of otherSequences is null.
+        /// </summary>
+        [Test]
+        public void TestInterleaveEarlyThrowOnNullElementInOtherSequences()
```
atifaziz: This test seems unrelated to this PR so best to move it to another.

MoreLinq/Interleave.cs
Orace: Fixed in 205c371a

MoreLinq/Interleave.cs
Orace: Fixed in 205c371a

atifaziz: I think you misunderstood me here. I meant, instead of:

```c#
enumerator.Dispose();
enumerators[i] = null;
```

swap to this:

```c#
enumerators[i] = null;
enumerator.Dispose();
```

This is just good habit: update own state first then call external parties. If the enumerator's `Dispose` is well-implemented then it should never throw but if a poorly implemented one does then our state is consistent (that is, we consider the enumerator gone nonetheless) if someone catches and tries to resume iteration regardless.

Orace: > This is just good habit: update own state first then call external parties.

Will made a tattoo of this one :)

MoreLinq.Test/InterleaveTest.cs
Orace: Fixed in 7d0aa0d and a7f2c6f

MoreLinq/Interleave.cs
Orace: @atifaziz with the implementation using `Acquire` in this PR (currently at a7f2c6f4), calls to `GetEnumerator` are made eagerly and the code below throw:

```C#
var sequenceA = Enumerable.Range(1, 1);
var sequenceB = new BreakingSequence<int>();
sequenceA.Interleave(sequenceB).Take(1).Consume();
```

I don't know it it's acceptable

MoreLinq.Test/InterleaveTest.cs
```diff
@@ -35,6 +35,23 @@ public void TestInterleaveIsLazy()
...
+        {
+            void Code()
+            {
+                var sequenceA = Enumerable.Range(1, 1);
+                var sequenceB = MoreEnumerable.From<int>(() => throw new TestException());
+
+                sequenceA.Interleave(sequenceB).Take(1).Consume();
+            }
+
+            Assert.DoesNotThrow(Code);
```
atifaziz: This seems redundant in a test, don't you think? If an exception is thrown then the test will fail anyhow. You could then throw away the `Code` local function. If you prefer to see a test do some sort of assertion at the end then I would suggest rewriting this simply as:

```c#
[Test]
public void TestInterleaveDoNoCallMoveNextEagerly()
{
    var sequenceA = Enumerable.Range(1, 1);
    var sequenceB = MoreEnumerable.From<int>(() => throw new TestException());

    var result = sequenceA.Interleave(sequenceB).Take(1);

    Assert.DoesNotThrow(() => result.Consume());
}
```

Orace: I simplified it in efd0f4c

Issue #695, opened 2019-11-13
[Proposal] EquiInterleave


Orace:
Like `EquiZip`, `EquiInterleave` throw if input sequences differ in size.

Orace:
Does `{1, 2}.EquiInterleave({3})` should yield:
 - 1, 3, 2, exception
 - 1, 3, exception
?!

atifaziz:
This can be achieved relatively easily with `EquiZip` + `SelectMany`:

```c#
var xs =
    from e in Enumerable.Range(1, 10)
                        .EquiZip(Enumerable.Range(11, 10), ValueTuple.Create)
    from x in MoreEnumerable.Return(e.Item1)
                            .Append(e.Item2)
    select x;

foreach (var x in xs)
    Console.WriteLine(x);
```

Outputs:

    1
    11
    2
    12
    3
    13
    4
    14
    5
    15
    6
    16
    7
    17
    8
    18
    9
    19
    10
    20

Yes, it's never going to be as optimal as an operator implemented by hand, but I'm afraid I don't think it's common enough to warrant adding and maintaining a whole new operator along with its tests.

I'm happy to reconsider if it can be demonstrated to be extremely common in some field.

PR #700, opened 2019-11-14
Add `EquiInterleave`

Orace:
This PR addresses #695

### DIFF
--- a/MoreLinq/EquiInterleave.cs
+++ b/MoreLinq/EquiInterleave.cs
@@ -75,7 +75,7 @@ namespace MoreLinq
                     var (isHomogeneous, hasNext) = enumerators.Select(e => e.MoveNext()).IsHomogeneous();
 
                     if (isHomogeneous == false)
-                        throw new InvalidOperationException("Input sequences are of different length.");
+                        throw new InvalidOperationException("Input sequences are of different lengths.");
 
                     if (!hasNext)
                         break;
--- a/MoreLinq/EquiInterleave.cs
+++ b/MoreLinq/EquiInterleave.cs
@@ -70,16 +70,20 @@ namespace MoreLinq
                 if (enumerators.Count == 0)
                     yield break;
 
-                for (;;)
+                bool MoveNext()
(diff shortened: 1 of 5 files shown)


MoreLinq/EquiInterleave.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                if (enumerators.Count == 0)
+                    yield break;
+
+                for (;;)
+                {
+                    var (isHomogeneous, hasNext) = enumerators.Select(e => e.MoveNext()).IsHomogeneous();
+
+                    if (isHomogeneous == false)
+                        throw new InvalidOperationException("Input sequences are of different length.");
```
Orace:
```suggestion
                        throw new InvalidOperationException("Input sequences are of different lengths.");
```

Crossbow78:
Out of curiosity, what does this add compared to the existing `Interleave` operation, other than an exception with imbalanced sequences?

Wouldn't it be more elegant to make the second `Interleave` overload that accepts a `ImbalancedInterleaveStrategy` argument public, and add a "Throw" strategy to that enumeration? Or at least use it internally, since the enumeration logic should be identical?

Orace:
> Out of curiosity, what does this add compared to the existing `Interleave` operation, other than an exception with imbalanced sequences?
> 
> Wouldn't it be more elegant to make the second `Interleave` overload that accepts a `ImbalancedInterleaveStrategy` argument public, and add a "Throw" strategy to that enumeration? Or at least use it internally, since the enumeration logic should be identical?

The "strategy as parameter" pattern has been discussed for the zip methods. (I will try to find the reference later).
Turns out that "strategy as overload" pattern has been chosen for the zip methods : ZipShortest, ZipLongest and EquiZip.
I just kept this pattern here.

TL; DR; not my choice in first place.

atifaziz:
Closing as #695 was [closed](https://github.com/morelinq/MoreLINQ/issues/695#issuecomment-1399078449) as unplanned for now.

PR #701, opened 2019-11-14
Update README.md : Interleave actually only have one overload

Orace:
Interleave actually only have one overload


Orace:
Link:
https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/Interleave.cs

PR #702, opened 2019-11-14
Added Triplewise operator

jonasdmentia:
This is basically a clone of Pairwise, modified for iterating over triplets instead of pairs.

It probably wouldn't take much to generalise Pairwise and Triplewise as a method that overloads on the number of arguments to the resultSelector such that the following is valid:
```
source.Tuplewise((a, b) => a + b);
source.Tuplewise((a, b, c) => a + b + c);
source.Tuplewise((a, b, c, d) => a + b + c + d);
```


Orace:
Maybe this could be overloads for [Batch ](https://github.com/morelinq/MoreLINQ/blob/master/README.md#batch)?

atifaziz:
You can do this today with [`Window`](https://morelinq.github.io/3.2/ref/api/html/M_MoreLinq_MoreEnumerable_Window__1.htm) + [`Fold`](https://morelinq.github.io/3.2/ref/api/html/Overload_MoreLinq_MoreEnumerable_Fold.htm) except window allocation might be expensive for simple cases. Also, the window size and `Fold` arity can be redundant and error prone to maintain.

> It probably wouldn't take much to generalise Pairwise and Triplewise

I like `Tuplewise` as the more general name and solution so why not just start with that? `Pairwise` would then just delegate to `Tuplewise` of 2.

Orace:
A `.tt` template looks appropriate here.

MoreLinq.Test/TriplewiseTest.cs
```diff
@@ -0,0 +1,49 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008-2011 Jonathan Skeet. All rights reserved.
```
Orace: Don't be shy, put your name here.

MoreLinq.Test/TriplewiseTest.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            Assert.That(result, Is.Empty);
+        }
+
+        [Test]
+        public void TriplewiseWideSourceSequence()
+        {
+            var result = new[] { "a", "b", "c", "d", "e" }.Triplewise((x, y, z) => x + y + z);
+            result.AssertSequenceEqual("abc", "bcd", "cde");
+        }
```
Orace: Add tests cases mechanics can allow you to merge `TriplewiseWithSequenceShorterThanTwo` and `TriplewiseWideSourceSequence`.

Here an example:
https://github.com/morelinq/MoreLINQ/blob/5ad0a6528a6d7708189518396026fd0c05df65c1/MoreLinq.Test/ZipLongestTest.cs#L31-L52

MoreLinq.Test/TriplewiseTest.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            new BreakingSequence<object>().Triplewise(BreakingFunc.Of<object, object, object, int>());
+        }
+
+        [TestCase(0)]
+        [TestCase(1)]
+        [TestCase(2)]
+        public void TriplewiseWithSequenceShorterThanTwo(int count)
+        {
+            var source = Enumerable.Range(0, count);
```
Orace: You should use `AsTestingSequence()`
https://github.com/morelinq/MoreLINQ/blob/183ee85f914882a4d3e34f13dee60100c3d61a5d/MoreLinq.Test/TestingSequence.cs#L30

MoreLinq.Test/TriplewiseTest.cs
jonasdmentia: Hahah, oops. I just copy-n-pasted the copyright block without paying any attention to it :)

atifaziz:
> Because I only thought of it while submitting the PR :)

Happens to me all the time. 😆

atifaziz:
> > A .tt template looks appropriate here.
> 
> Seems reasonable. I've not played with .tt templates much but I'll see what I come up with.

Awesome and an occasion to learn and there are plenty of examples in MoreLINQ.

jonasdmentia:
> It probably wouldn't take much to generalise Pairwise and Triplewise
> 
> 
> 
> I like Tuplewise as the more general name and solution so why not just start with that? Pairwise would then just delegate to Tuplewise of 2.
> 
> Because I only thought of it while submitting the PR :)
> 
> In light of this, I vote to abandon this PR in favour of a new one that adds Tuplewise with its overloads.

That's fair, will do.

atifaziz:
> That's fair, will do.

Thanks @jonasdmentia! Closing this meanwhile.

PR #714, opened 2019-11-20
Improve Insert tests

Orace:
Added tests for:
 - InsertIsLazy
 - Improve tests coverage:
   - two empty sequence
   - first sequence empty
   - second sequence empty and inserted at beginning/middle/end of the first sequence

Refactored some test to use `ITestCaseData`.
Impacted tests are:
 - first and second sequence not empty and second inserted at beginning/middle/end of the first sequence


atifaziz:
There's been no follow-up so closing as stale for now.

PR #715, opened 2019-11-20
Zip refactoring

Orace:
This PR continues #639 since I erroneously deleted it's branch.


Orace:
To improve the documentation,  we may get a lesson from [System.Linq.Enumerable.Zip](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.zip?view=netcore-3.0) method.

Issue #716, opened 2019-11-20
Support conditional application to avoid breaking the chain


veloek:
It would be very nice to have a way to choose to apply parts of a LINQ chain based on a condition or a predicate. Consider this fictional method that returns a processed stream of messages:

```csharp
public static IEnumerable<string> ProcessMessages(IEnumerable<string> messages, bool ordered = false, int limit = 0)
{
    if (!messages.Any())
        messages = messages.Append("Initial");

    messages = messages
        .Where(s => !string.IsNullOrEmpty(s))
        .Select(s => s.ToLower());

    if (ordered)
        messages = messages.OrderBy(s => s);

    messages = messages
        .Select((s, i) => $"{i}: {s}");

    if (limit > 0)
        messages = messages.Take(limit);

    return messages;
}
```

The need to split the chain and make new assignments to store parts of the chain is annoying and could be omitted by implementing a conditional application method `If` based on a given condition or predicate. With this new method, we could refactor the code to this:

```csharp
public static IEnumerable<string> ProcessMessagesWithIf(IEnumerable<string> messages, bool ordered = false, int limit = 0)
{
    return messages
        .If(m => !m.Any(), m => m
            .Append("Initial"))
        .Where(s => !string.IsNullOrEmpty(s))
        .Select(s => s.ToLower())
        .If(ordered, m => m
            .OrderBy(s => s))
        .Select((s, i) => $"{i}: {s}")
        .If(limit > 0, m => m
            .Take(limit));
}
```

I'll make a PR with an implementation of this method to demonstrate the behavior.

atifaziz:
@veloek Thanks for wanting to contribute to the project. However, I am afraid that what you are proposing has nothing to do with LINQ or sequences. See [my comment on your PR](https://github.com/morelinq/MoreLINQ/pull/717#pullrequestreview-320240610). As a result, I think it belongs as a method in a more general-purpose extensions library.

Also, have a look at [`If` in System.Interactive](https://github.com/dotnet/reactive/blob/rxnet-v4.2.0/Ix.NET/Source/System.Interactive/System/Linq/Operators/If.cs) that has existed since some time.

atifaziz:
I am going to close this for now based on [my comment above](https://github.com/morelinq/MoreLINQ/issues/716#issuecomment-556332089) but don't let that discourage you from proposing other ideas.

I am happy to reconsider if you have a compelling argument to review this in a new light.

atifaziz:
[...]

PR #717, opened 2019-11-20
Add `If` method to do conditional application

veloek:
This new "helper" method will make it possible to have conditional parts of a LINQ chain. See #716 for more information about the problem it's meant to solve.

I'm not too familiar with the code base, but I've tried to keep it similar to the other methods in the project.


atifaziz:
Closing as related issue #716 has been closed too.

Issue #719, opened 2019-11-22
SortedMerge doesn't throw on null parameter.


Orace:
Like interleave in #696 (see 76e3516), SortedMerge doesn't throw immediately when a `null` element is passed via `params`.

atifaziz:
Closing because `Interleave` has been aligned (in #1029) with `SortedMerge` behaviour with respect to this issue.

PR #720, opened 2019-11-22
SortedMerge doesn't throw immediately on null parameter.

Orace:
This PR addresses #719.

The first commit add TestSortedMergeEarlyThrowOnNullElementInOtherSequences that show the flaw.


atifaziz:
Abandoning as related issue #719 has been closed.

Issue #721, opened 2019-11-22
From doesn't throw on null parameter


Orace:
Like Interleave in #696 (see 76e3516), From doesn't throw immediately when a null element is passed via params or as classic argument in any of the overloads.

atifaziz:
This [was discussed](https://github.com/morelinq/MoreLINQ/pull/333#pullrequestreview-53813671) in PR #333. Please see all the comments and review there.

I am closing this as by-design.

PR #722, opened 2019-11-22
From doesn't throw immediately on null parameter.

Orace:
This PR addresses #721.

The first commit add TestFromWithXParametersEarlyThrowOnNullParameter that show the flaw.

From should not depend on Evaluate that re-test for functions to not be null.


Orace:
This was discussed in PR #333. Please see all the comments and review there.

I am closing this as by-design.

PR #723, opened 2019-11-22
Evaluate : add a specific message if an element of the provided functions is null

Orace:
Add a more specific message if an element of the provided functions is null
update the documentation accordingly.

### DIFF
--- a/MoreLinq/Evaluate.cs
+++ b/MoreLinq/Evaluate.cs
@@ -34,10 +34,10 @@ namespace MoreLinq
         /// <typeparam name="T">The type of the object returned by the functions.</typeparam>
         /// <param name="functions">The functions to evaluate.</param>
         /// <returns>A sequence with results from invoking <paramref name="functions"/>.</returns>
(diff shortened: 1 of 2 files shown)


atifaziz:
Closing this as by-design.

atifaziz:
One can add something in the `<remarks>` section to highlight that the sequence elements are not validated for null references.

Orace:
> That would be incorrect. It throws `NullReferenceException` for any function in the sequence that is `null` and this is by-design.

Actual code throws `ArgumentNullException`.

https://github.com/morelinq/MoreLINQ/blob/9fc8486b9d8819690c9bc67fe41e74d4e1e2d765/MoreLinq/Evaluate.cs#L40

So in case of `ArgumentNullException` the caller can't know if it's because `functions` is null or if one element from `functions` is null.

I'm ok to put a `NullReferenceException` for the second case (with or without message), but we can't leave it as it is.

@atifaziz , can you please re-open this.

Issue #724, opened 2019-11-22
Add private method ToDisposableLinkedList to replace DisposableGroup and Acquire


Orace:
`Transpose`, `Interleave` and  `SortedMerge` are three methods that have to deals with enumeration of a collection of `IEnumerable`.

In this three methods, implementation have to ensure that all enumerators from the input enumerables are disposed when:

1. The enumeration of the input collection of IEnumerable fail.
2. The result sequence enumerator is disposed.

*1.* is assured by `Acquire` in the three methods.
*2.* is assured by `DisposableGroup` in `SortedMerge` and by some `try {} finaly {}` blocks in `Transpose` and `Interleave` (and those make `Acquire` useless as discussed here #696 ).

`Acquire` return an array, `DisposableGroup` is based on an list.
For the three methods we can build cases where elements are heavily removed from the array/list. Since remove is O(N) in time for array/list, that lead to O(N²) time complexity.
This can be easily fixed with the use of a `LinkedList`.

I propose a `ToDisposableLinkedList()` method that *1.* have the functionality of Acquire (on error, dispose already acquired elements) and provide a `DisposableLinkedList` that *2.* have the functionality of `DisposableGroup<T>` (being disposable and dispose its content when disposed).

It will be used like that:

```C#
using var enumerators = source.Select(e => e.GetEnumerator()).ToDisposableLinkedList();
```

The `using` here allow to remove the `try {} finally {}` from here:
https://github.com/morelinq/MoreLINQ/blob/9fc8486b9d8819690c9bc67fe41e74d4e1e2d765/MoreLinq/Transpose.cs#L92-L96
and here:
https://github.com/morelinq/MoreLINQ/blob/9fc8486b9d8819690c9bc67fe41e74d4e1e2d765/MoreLinq/Interleave.cs#L141-L146

and improve the `DisposableGroup/Acquire` combo from here:

https://github.com/morelinq/MoreLINQ/blob/9fc8486b9d8819690c9bc67fe41e74d4e1e2d765/MoreLinq/SortedMerge.cs#L101

Orace:
A first step is to use the `Disposable/Acquire` combo in `Transpose` and `Interleave`.

atifaziz:
This is just an idea but `Acquire` can be overloaded to return any type of collection, including a `LinkedList<>`:

```c#
public static TCollection Acquire<TSource, TCollection>(this IEnumerable<TSource> source,
                                                        TCollection disposables)
[...]

PR #726, opened 2019-11-25
Use `LinkedList` in `Interleave` to avoid many null checks

Orace:
Use a `LinkedList` in `Interleave` implementation to avoid many null checks.
See #724, #696


atifaziz (approved):
Thanks, this certainly makes the implementation simpler and better!

viceroypenguin (approved):

Issue #731, opened 2019-12-12
Transpose eagerly calls GetEnumerator


Orace:
Like #729, with current implementation of `Transpose`,
The code below throws:

```C#
var sequences = new IEnumerable<int>[]
{
    TestingSequence.Of(1),
    new BreakingSequence<int>()
};

var transpose = sequences.Transpose().Take(1).ToList();
```

While we can expect `transpose` to be a list of one element.

Orace:
Since transpose emit batch, it's a normal behavior.

PR #732, opened 2019-12-12
Transpose: Do not call GetEnumerator eagerly.

Orace:
This PR addresses #731


Orace:
Since transpose emit batch, it's a normal behavior.

PR #735, opened 2019-12-12
SortedMerge: Test do not call MoveNext eagerly

Orace:
This PR add `TestSortedMergeDoNotCallMoveNextEagerly`.
Unlike for `Interleave` (#694), this test pass.

### DIFF
--- a/MoreLinq.Test/SortedMergeTest.cs
+++ b/MoreLinq.Test/SortedMergeTest.cs
@@ -55,6 +55,17 @@ namespace MoreLinq.Test
         }
 
         /// <summary>
+        /// Verify that SortedMerge do not call MoveNext method eagerly
+        /// </summary>
+        [Test]
+        public void TestSortedMergeDoNotCallMoveNextEagerly()
+        {
+            var sequenceA = TestingSequence.Of(1, 3);
+            var sequenceB = MoreEnumerable.From(() => 2, () => throw new TestException());
+            sequenceA.SortedMerge(OrderByDirection.Ascending, sequenceB).Take(2).Consume();
+        }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


MoreLinq.Test/SortedMergeTest.cs
```diff
@@ -54,6 +54,17 @@ public void TestSortedMergeDisposesOnError()
...
 
+        /// <summary>
+        /// Verify that SortedMerge do not call MoveNext method eagerly
+        /// </summary>
+        [Test]
+        public void TestSortedMergeDoNotCallMoveNextEagerly()
+        {
+            var sequenceA = TestingSequence.Of(1, 3);
+            var sequenceB = MoreEnumerable.From(() => 2, () => throw new TestException());
+            sequenceA.SortedMerge(OrderByDirection.Ascending, sequenceB).Take(2).Consume();
```
atifaziz, line 65: Would be great if we can update this to bring it in line with the similar test that was added for `Interleave` in PR #696:

https://github.com/morelinq/MoreLINQ/blob/e98d6322c75af1902359947661b85a867c40df50/MoreLinq.Test/InterleaveTest.cs#L94-L106

It just needs the following changes (use testing sequences, dispose them & always end in an assertion):

```suggestion
            using var sequenceA = TestingSequence.Of(1, 3);
            using var sequenceB = MoreEnumerable.From(() => 2, () => throw new TestException())
                                                .AsTestingSequence();

            var result = sequenceA.SortedMerge(OrderByDirection.Ascending, sequenceB).Take(2);

            Assert.That(() => result.Consume(), Throws.Nothing);
```

atifaziz (changes requested):
FWIW after all this time, just one change and we can merge this.

Thanks @Orace!

atifaziz (approved):
Thank you!

Issue #10, opened 2015-08-21
Rename Pipe to ViaAction/Cause/Visit


atifaziz:
The `Pipe` operator sounds too general and vague. Any set of LINQ operations chained together could be seen as a pipeline or pipe formation/composition. `Pipe` also does not imply any side effect. Consequently, consider renaming `Pipe` to `ViaAction`. `ViaAction` clearly states two things: (1) the operator is pass-through, as in via, and (2) it implies an action, which for the majority case is understood to be a side-effect.

---

Originally reported on Google Code with ID 10

Reported by @atifaziz on 2009-02-17 21:05:47

JamesFaix:
What about `DoYield`?
`var avg = new[] {1, 2, 3}.DoYield(x => Console.WriteLine(x)).Average()`

atifaziz:
@JamesFaix I think it would be reasonable/sensible at this stage to borrow [`Do` from Rx](https://msdn.microsoft.com/en-us/library/system.reactive.linq.observable.do%28v=vs.103%29.aspx?f=255&MSPPError=-2147217396) here and move on.

PR #744, opened 2020-03-08
New operator: BatchWhile

Kir-Antipov:
I was really surprised that MoreLINQ still has no method with such functionality: break the sequence into subsequences by predicate, so I've made it myself :)

## Example

Let's recall some commonplace problem from school: find the maximum length of a subsequence of repeating numbers in an array. Using `BatchWhile` this can be written like this:

```csharp
int[] array = new[] { 1, 2, 3, 3, 3, 3, 4, 5, 5, 5 };

int maxLength = array.BatchWhile((currentElement, bucket) =>
{
    // Every bucket has at least 1 element
    int previousElement = bucket[bucket.Count - 1];

    // If predicate will return `false`, a new subsequence begins
    return currentElement == previousElement;
},
// Transform each bucket into `int` representing its length
bucket => bucket.Count)
// No comments :)
.Max();

// maxLength == 4
```

Or easier:
```csharp
int maxLength = array.BatchWhile((x, bucket) => x == bucket[0], bucket => bucket.Count).Max();
```

## Overloads

```csharp
IEnumerable<TResult> BatchWhile<TSource, TElement, TResult>(this IEnumerable<TSource> source, Func<TSource, TElement> elementSelector, Func<TElement, IBucket<TElement>, bool> predicate, Func<IBucket<TElement>, TResult> resultSelector);

IEnumerable<TResult> BatchWhile<TSource, TResult>(this IEnumerable<TSource> source, Func<TSource, IBucket<TSource>, bool> predicate, Func<IBucket<TSource>, TResult> resultSelector);

IEnumerable<IBucket<TElement>> BatchWhile<TSource, TElement>(this IEnumerable<TSource> source, Func<TSource, TElement> elementSelector, Func<TElement, IBucket<TElement>, bool> predicate);

IEnumerable<IBucket<TSource>> BatchWhile<TSource>(this IEnumerable<TSource> source, Func<TSource, IBucket<TSource>, bool> predicate);
```


Kir-Antipov:
`codecov` doesn’t like code duplication for `IReadOnlyList<T>` case, but it was forced, as you know that `IList<T>` and `IReadOnlyList<T>` implement indexing in their own way, so... ¯\\_(ツ)_/¯

leandromoh:
>  break the sequence into subsequences by predicate

`Segment` operator already does it @Kir-Antipov .

atifaziz:
@Kir-Antipov Thanks for your contribution, but I wish you had opened an issue to check before submitting a PR with a full implementation. As @leandromoh [point out already](https://github.com/morelinq/MoreLINQ/pull/744#issuecomment-613082110), we have this except it's called [`Segment`](https://morelinq.github.io/3.3/ref/api/html/Overload_MoreLinq_MoreEnumerable_Segment.htm). I just hope you didn't end up wasting too much time.

If you want to [submit a working example](https://github.com/morelinq/examples/issues/68) for [`Segment`](https://github.com/morelinq/examples/blob/4d846ba0680ae8e9a706fad2e3a04ecdc8c59caf/m/segment.md) then that would be much welcome & appreciated.

FoggyFinder:
@atifaziz are you sure that `BatchWhile` is totally duplicate what `Segment` can do? From what I can see `BatchWhile` is wider.

atifaziz:
> @atifaziz are you sure that `BatchWhile` is totally duplicate what `Segment` can do? From what I can see `BatchWhile` is wider.

I think so based on the example in the [opening comment of this PR](https://github.com/morelinq/MoreLINQ/pull/744#issue-385217866), though I'll admit I didn't look at the code, and the following description (bearing in mind it's missing details that are open for interpretation):

> break the sequence into subsequences by predicate, so I've made it myself :)

The problem the example intends to solve can be rewritten with `Segment`:

```c#
int[] array = new[] { 1, 2, 3, 3, 3, 3, 4, 5, 5, 5 };
int maxLength = array.Segment((curr, prev, _) => curr != prev)
                     .Select(g => g.Count())
                     .Max();
```

as well as, and more simply, with [`GroupAdjacent`](https://morelinq.github.io/3.3/ref/api/html/Overload_MoreLinq_MoreEnumerable_GroupAdjacent.htm):

```c#
int[] array = new[] { 1, 2, 3, 3, 3, 3, 4, 5, 5, 5 };
int maxLength = array.GroupAdjacent(x => x)
                     .Select(g => g.Count())
                     .Max();
```

Kir-Antipov:
> [Segment](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/Segment.cs) operator already does it @Kir-Antipov .

@leandromoh, @atifaziz, nah, not really. As @FoggyFinder noted, `BatchWhile` operator can do much more than `Segment`.

How can you express the desire *"break a sequence into subsequences so that the sum of the elements of each of them does not exceed 42"* with `Segment`? Don't even try, 'cause it's impossible :)
In the same time:

```csharp
array.BatchWhile((current, bucket) => bucket.Sum() + current <= 42)
```

Btw, I want to note that the name `Segment` is extremely unobvious. A couple of times my needs could be covered even by the `Segment`, but after the `Batch` I no longer expected to see such operator. Why? `Take/TakeWhile` - first takes a number as an input, second - a predicate. `Batch/BatchWhile` - the analogy is preserved, `Batch/Segment` - nope.

If we continue the topic of LINQ analogies, `Segment` has one more drawback: LINQ methods that return `IEnumerable<IEnumerable<T>>` (e.g., `GroupBy`) make it possible to project inner `IEnumerable<T>` into new form so that you don't have to call an enumeration in the second round with `Select`. `BatchWhile` obeys this rule, `Segment` does not.

Imho, `BatchWhile` should replace `Segment` at all, since one is a superset of the other. But I clearly understand that something like this will never happen here, so let's just keep two operators together.

P.S. - I apologize for answering so long. A little slack because of the weird times we live in :)

atifaziz:
> How can you express the desire _"break a sequence into subsequences so that the sum of the elements of each of them does not exceed 42"_ with `Segment`? Don't even try, 'cause it's impossible :)

No one is saying you have to do everything with `Segment`. That was just to show that your initial example could be solved with it or even more simply with `GroupAdjacent`. Now your new problem might be impossible for `Segment` but who says you can't combine existing operations cleverly? In fact, as we've seen, `GroupAdjacent` is simpler and I'd even say more powerful for your cases so far. Here's a solution using [`Scan`][scan] together with `GroupAdjacent`:

  [scan]: https://morelinq.github.io/3.3/ref/api/html/M_MoreLinq_MoreEnumerable_Scan__1.htm

```c#
// break a sequence into subsequences so that the sum
// of the elements of each of them does not exceed 42

var result =
    MoreEnumerable
        .Random(10)
        .Take(100)
        .Select(x => new { N = x, Sum = x, Batch = 0 })          // seed all in same batch
        .Scan((s, e) => s.Sum + e.N is {} sum && sum <= 42       // running sum <= 42?
                      ? new { e.N, Sum = sum, s.Batch }          // yes => same batch
                      : new { e.N, e.Sum, Batch = s.Batch + 1 }) // no  => new batch (+1)
        .GroupAdjacent(e => e.Batch)                             // group by batch
        .Select(g => $"({g.Select(m => m.N).ToDelimitedString(", ")}) = {g.Last().Sum}")
        .ToDelimitedString(Environment.NewLine);

Console.WriteLine(result);
```

Here is the output from one run:

```
41 = 9 + 0 + 8 + 3 + 9 + 5 + 7
42 = 3 + 1 + 4 + 6 + 6 + 3 + 2 + 4 + 7 + 6
36 = 4 + 8 + 4 + 9 + 7 + 3 + 1 + 0
39 = 9 + 9 + 4 + 3 + 6 + 0 + 0 + 3 + 5
39 = 5 + 5 + 6 + 2 + 6 + 3 + 1 + 8 + 3
40 = 6 + 5 + 7 + 8 + 3 + 1 + 3 + 6 + 1
41 = 3 + 8 + 2 + 1 + 6 + 7 + 3 + 0 + 5 + 2 + 0 + 4
37 = 3 + 6 + 2 + 8 + 7 + 8 + 0 + 3
41 = 8 + 4 + 2 + 1 + 4 + 6 + 6 + 2 + 0 + 5 + 3
42 = 9 + 0 + 9 + 6 + 5 + 2 + 3 + 8
40 = 9 + 9 + 6 + 8 + 8
26 = 7 + 9 + 2 + 8
```

and here is another:

```
38 = 4 + 9 + 6 + 2 + 3 + 2 + 3 + 9
42 = 6 + 1 + 7 + 0 + 1 + 0 + 9 + 9 + 9 + 0
38 = 3 + 4 + 9 + 7 + 6 + 3 + 0 + 1 + 4 + 1
35 = 6 + 3 + 3 + 4 + 3 + 7 + 8 + 1
36 = 8 + 1 + 6 + 8 + 6 + 7
38 = 9 + 0 + 6 + 8 + 7 + 8
36 = 6 + 8 + 9 + 7 + 6
42 = 8 + 9 + 7 + 2 + 3 + 1 + 2 + 7 + 1 + 2
40 = 2 + 8 + 2 + 6 + 4 + 2 + 8 + 0 + 8
40 = 4 + 5 + 7 + 2 + 8 + 6 + 7 + 1
40 = 3 + 6 + 5 + 8 + 4 + 6 + 8
39 = 5 + 8 + 0 + 8 + 5 + 9 + 3 + 1
21 = 5 + 4 + 7 + 2 + 3
```

So `Scan` is used to keep a running total/sum and as soon as it is about to exceed 42, the accumulator is _reset_ with a new batch number. Next `GroupAdjacent` is used to simply group elements by their batch number. A nice thing about this is that both, `Scan` and `GroupAdjacent`, _stream_ their results so you can work with very large data efficiently and observe the results as soon as they become available. In the case of `GroupAdjacent`, the groups (as opposed to the elements) are streamed; that is, a group is yielded as soon as it ends and before the next one is starts. At any one time, only a single group is committed to memory and so need to only account for the largest group in terms of space complexity.

Note that batches don't need to be numbered _uniquely_ since `GroupAdjacent` will _switch_ groups whenever the key changes. You can rewrite the above with a Boolean key that you simply toggle (`!`) whenever the sum of elements is about to exceed the threshold (below, thinking in terms of two _bins_):

```c#
var result =
    MoreEnumerable
        .Random(10)
        .Take(100)
        .Select(x => new { N = x, Sum = x, Bin = false })   // seed all in same bin
        .Scan((s, e) => s.Sum + e.N is {} sum && sum <= 42  // running sum <= 42?
                      ? new { e.N, Sum = sum, s.Bin }       // yes => same bin
                      : new { e.N, e.Sum, Bin = !s.Bin })   // no  => "switch" bin
        .GroupAdjacent(e => e.Bin)                          // group by bin
        .Select(g => $"{g.Last().Sum} = {g.Select(m => m.N).ToDelimitedString(" + ")}")
        .ToDelimitedString(Environment.NewLine);
```

> Btw, I want to note that the name `Segment` is extremely unobvious. A couple of times my needs could be covered even by the `Segment`, but after the `Batch` I no longer expected to see such operator. Why? `Take/TakeWhile` - first takes a number as an input, second - a predicate. `Batch/BatchWhile` - the analogy is preserved, `Batch/Segment` - nope.

Yeah, [naming is hard](https://martinfowler.com/bliki/TwoHardThings.html) 😉  and we've had naming debates in the past that go as early as issue #10 (some 11 years ago) in the project's long history. I am not quite sure how to solve the problem of allowing someone to quickly navigate the 100+ extensions and get familiar with them. Sometimes, people come with a specific idea and try to find a matching name that's already in their head, but often a solution can be found by combining two or more operators cleverly.

> But I clearly understand that something like this will never happen here

I hope you are not discouraged from contributing in the future. Part of the review process is to challenge things that have often lead to more generalisations or dead ends (which has happened to me several times). For what it's worth, I checked out the code you submitted and it's really high on the quality bar and I wish every contribution would begin like yours. Right now, it seems redundant with existing operators, which is something important to consider for maintenance.

> P.S. - I apologize for answering so long. A little slack because of the weird times we live in :)

It's understandable and there is no need to apologize now or any another time since this is open source. It's generally volunteered work and time. Remember, I took also almost 2 weeks to post my first reply. 😉

Issue #565, opened 2019-04-04
Ease usage in next major version by renaming methods conflicting with Enumerable methods


voxoid0:
For historical reasons, some of MoreLinq's method names conflict with System.Linq.Enumerable's, causing errors like the following from [Issue 527](https://github.com/morelinq/MoreLINQ/issues/527):

`Cannot execute text selection: CS0121 The call is ambiguous between the following methods or properties: 'System.Linq.Enumerable.Append(System.Collections.Generic.IEnumerable, TSource)' and 'MoreLinq.MoreEnumerable.Append(System.Collections.Generic.IEnumerable, T)'`

This can often be resolved by using a static import of one of MoreLinq's *Extension classes, such as MoreLinq.Extensions.AppendExtension.

However, having to manually do a static import rather than being able to continue to use tooling's auto-import (such as provided by Visual Studio or ReSharper), while it makes a work-around possible, is still a big pain and a slow-down. Furthermore it seems that some cases of using methods from both Linq and MoreLinq in the same file are only resolvable by calling the extension method from the static class name directly, rather than using it as an extension of IEnumerable.

For this reason I propose that in the next major version of MoreLinq, the conflicting method names be renamed where they may cause this conflict. (The same thing could happen again in the future if more Enumerable methods are added to System.Linq, but at least the conflicts will be reduced, and for a while exhaustively.)

atifaziz:
Below is a copy of relevant comments from #527 for context.

---

@voxoid0 [wrote](https://github.com/morelinq/MoreLINQ/issues/527#issuecomment-479874302):

> Having to manually do a static import rather than being able to continue to use tooling's auto-import, while it makes a work-around possible, is still a big pain and a slow-down. Furthermore it seems that some cases of using methods from both Linq and MoreLinq are only resolvable by calling the extension method from the static class name directly, rather than using it as an extension of IEnumerable.
> 
> For this reason I propose that in the next major version of MoreLinq, the conflicting method names be renamed where they may cause this conflict.

@atifaziz [wrote](https://github.com/morelinq/MoreLINQ/issues/527#issuecomment-479895529):

> > some cases of using methods from both Linq and MoreLinq are only resolvable by calling the extension method from the static class name directly
> 
> @voxoid0 Do you have examples of those cases?
> 
> > For this reason I propose that in the next major version of MoreLinq, the conflicting method names be renamed where they may cause this conflict.
> 
[...]

PR #749, opened 2020-07-09
Modern framework support

### LINKED ISSUE #738: ToHashSet has been superseded in .NET Core 2.0+, .NET Framework 4.7.1+
The base .Net framework, starting in .Net Framework 4.7.1, and .Net Core 2.0, includes its own IEnumerable extension method, .ToHashSet(). It should be removed from MoreLinq with a compiler flag for 4.7.1+ and Core 2.0+. Currently, it's easy to end up in a place where you can't compile if you have using statements for both `System.Linq` and `MoreLinq`.

https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tohashset?view=netframework-4.8

viceroypenguin:
Properly implement solution for hiding extension methods on different frameworks by removing the `this` modifier. 

* Supports binary compatibility by still compiling duplicated methods. Since extension methods are implemented via syntactic sugar, as long as the methods still exist, the `this` modifier can be removed and callers will still be able to reach the method.
* Added more build frameworks and test frameworks in order to handle different times new methods were added.

Fixes #738; improvement over #739


viceroypenguin:
@atifaziz I have released morelinq.temp as a proof of concept based on this branch. I have found this to improve compile-time compatibility while retaining run-time compatibility. This should be easier than using the `using static` method, that is still viable for all users. 

I have checked .NET 5, and there are no changes to `System.Linq.Enumerable`, so maintenance is not affected by the upcoming release.

Please let me know if there are any issues with this solution.

fsateler:
I wonder if, instead of polluting the code with `#if NET_blabla`, it would be possible to apply a transform using Fody or any modern roslyn-based equivalent?

viceroypenguin:
I could do it with a custom Fody weaver, but I would want to know that you guys are comfortable adding a dependency on Fody. That said, I think this is probably simpler than doing that, IMHO. A custom weaver would require an additional project and some work to identify the appropriate methods and strip the `this` attribute from the parameter.

Arithmomaniac:
Awaiting this eagerly. Anything holding it up?

viceroypenguin:
@atifaziz  I've updated with build tests to future proof this. It will become more important with .net 6, which is adding several functions that will overlap with MoreLinq (dotnet/runtime#27687).

viceroypenguin:
@atifaziz  - Updated to include .net6 support. Breaking change in `.MaxBy()` and `.MinBy()` due to conflict in behavior with the new matching methods in the fx. Binary compatibility is maintained, but source is not. Please advise if you will be reviewing this or not.

viceroypenguin:
I have released a preview version of a fork `SuperLinq`, which I will keep updated with newer .NET versions.

Arithmomaniac:
@viceroypenguin You should probably [unfork it](https://stackoverflow.com/a/66470086/1180926) once you have no intention of merging into MoreLinq anymore.

For starters, I'd want to file some issues 😉

viceroypenguin:
If you want to help by filing some issues, please do! My initial focus was just on cleaning things up and putting a preview on nuget. I'd like to start copying some issues over and then start addressing them.

viceroypenguin:
@Arithmomaniac thanks for pointing that out - I turned the issues setting on. I like your issue, so please file it. Already thinking about how to start implementing long-term.

viceroypenguin:
Closing per comments in #565 and fork released via SuperLinq

PR #755, opened 2020-08-11
Update all test dependencies

atifaziz:


atifaziz:
Coverage seems to have dropped a bit. Wonder if it's due to coverlet-coverage/coverlet#810? 💭

Issue #661, opened 2019-11-07
ZipMap: ([x], x → y) → [(x, y)]


atifaziz:
This was [initially proposed](https://github.com/morelinq/MoreLINQ/issues/624#issuecomment-549077029) by @Orace as part of the discussion in #624 but I am adding it here with the proposed name of `ZipMap`:

```c#
public static IEnumerable<(TSource Item, TResult Result)>
    ZipMap<TSource, TResult>(this IEnumerable<TSource> source, Func<TSource, TResult> func)
```

Element names are up for discussion.

declard:
Can't this be a composition of functions like `IEnumerable<(A Left, A Right)> SelectPairs(IEnumerable<A> source)` and `IEnumerable<(A Left, C RIght)> MapRight<A, B, C>(IEnumerable<(A Left, B Right)> source, Func<B, C> f)` which may become a good generic addition to MoreLinq?

atifaziz:
@declard Interesting idea! I like the generalization.

viceroypenguin:
Interesting idea, but I'm not sure it adds value. Is there a use-case for `SelectPairs()` outside of `ZipMap`; that is, is there a reason someone would want a single value as both parts of a tuple other than for use with a `MapRight` method? 

Similarly, while I can see potential use-cases, I'm not sure that `MapRight()` (or an expected matching `MapLeft()`) is something that would be used often. Even when used in queries, `ValueTuples` are generally used wholly, not partially, in my experience. As such, it looks like a solution in search of a problem.

`ZipMap` is a fairly simple function in the first place, basically `.Select(x => (x, func(x)))`, but eliminating the closure. I'm not sure that further generalization would be helpful.

declard:
Agree, `ValueTuple<,>` is not the best candidate. We have SelectPairs, MapKeys, and MapValues working with `KVP<,>` on a project, which allow things like
```csharp
collection
    .SelectPairs().MapValues(SomeFuncReturningOptionals).CollectSomeOptValues().MapValues(AnotherFunc).ToDictionary()`, very convenient.

If closure is in concern, then why not introducing something generic like
```csharp
IEnumerable<TResult> Select<TElement, TPayload, TResult>(
    this IEnumerable<TElement> collection,
    TPayload payload,
    Func<TElement, TPayload, TResult> func)
{
    using var iterator = collection.GetEnumerator();

    while (iterator.MoveNext())
    {
        yield return func(iterator.Current, payload);
    }
}
```
[...]

PR #763, opened 2020-10-01
Add ZipMap

matachi:
Implementation of `ZipMap` proposed in #661.

### DIFF
--- a/MoreLinq/ZipMap.cs
+++ b/MoreLinq/ZipMap.cs
@@ -1,0 +1,61 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Applies a function on each element and returns a sequence of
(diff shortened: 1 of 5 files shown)


MoreLinq/ZipMap.cs
```diff
@@ -0,0 +1,61 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2008 Jonathan Skeet. All rights reserved.
```
viceroypenguin, line 3: Notice has incorrect date and author

MoreLinq/ZipMap.cs
```diff
@@ -0,0 +1,61 @@
...
+        public static IEnumerable<(TSource Item, TResult Result)> ZipMap<TSource, TResult>(
+            this IEnumerable<TSource> source, Func<TSource, TResult> func)
+        {
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (func == null) throw new ArgumentNullException(nameof(func));
+
+            foreach (var item in source)
+            {
+                yield return (item, func(item));
+            }
```
atifaziz, line 58: The `NullArgumentTest` tests are [breaking](https://ci.appveyor.com/project/raboof/morelinq/builds/46674517/job/qad4h4lki7rhwh9q). The fix is to validate the arguments eagerly and wrap the iterator in a local function as follows (otherwise arguments are validated when the sequence is iterated rather than when `ZipMap` is called):

```suggestion
            return _(); IEnumerable<(TSource, TResult)> _()
            {
                foreach (var item in source)
                {
                    yield return (item, func(item));
                }
            }
```

README.md
```diff
@@ -710,6 +710,11 @@ for padding.
 
 This method has 3 overloads.
 
+### ZipMap
+
+Returns a sequence of tuples containing the source elements and the result of
+a function applied on each element.
```
atifaziz, line 716:
```suggestion
a function applied to each element.
```

MoreLinq/ZipMap.cs
atifaziz: Right. Should be changed to:

```suggestion
// Copyright (c) 2020 Daniel Jonsson. All rights reserved.
```

atifaziz (changes requested):
@matachi Thanks for contributing this and sorry for having taken so long to get back with a review. Usually, I don't pay attention to a new PR unless the CI build errors are fixed. Anyway, since `master` had moved on considerably since this PR was opened, I've aligned it with a merge and added my review. Hope you'll find the time to address the comments soon.

PR #771, opened 2020-11-27
#191 force

dzmitry-lahoda:
#191

### DIFF
--- a/MoreLinq/Experimental/Force.cs
+++ b/MoreLinq/Experimental/Force.cs
@@ -1,0 +1,57 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2017 Leandro F. Vieira (leandromoh). All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq.Experimental
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Linq;
+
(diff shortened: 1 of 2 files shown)


MoreLinq/Experimental/Force.cs
```diff
@@ -0,0 +1,57 @@
...
+            switch (source)
+            {
+                case T[] _:
+                case List<T> _:
+                case HashSet<T> _:
+                case Stack<T> _:
+                case Queue<T> _:
+                case SortedSet<T> _:
+                case LinkedList<T> _:
+                case null:
```
viceroypenguin, line 48: Add `IReadOnlyCollection<T>` to cover any custom types that may be defined outside the base fx but implement the interface.
```suggestion
                case IReadOnlyCollection<T> _:
                case null:
```

dzmitry-lahoda: nope, https://github.com/morelinq/MoreLINQ/issues/191#issuecomment-264584986 - risk free - i can impl Task<IROC<T>> which can be lazy (but says about it count).

viceroypenguin: Then I would do `IReadOnlyList<>` at least, because that must be realized already (in order to have O(1) access to each element non-sequentially).

dzmitry-lahoda: Not all counted collections form the switch list do the List:) some of them do the Collection. So Force is not ToList() so to speak. It is Force Lazy to Eager (means countable, but does not means indexed).

viceroypenguin: Agreed. I'm not saying get rid of the other items in the list. Only add `IReadOnlyList<T>` _as well_ to save re-materialization in other cases where it is not necessary.

viceroypenguin: I would update the `<return>` information, because it says that any countable enumeration will be returned as is. Correct it to match the summary that only well known collection types are left untouched.

dzmitry-lahoda: I agree that IReadOnlyList is much more safe addition to switch, but it can be still added later as low risk backward compatible option. Also I agree that it would be kind of strange to implement IReadOnlyList  in lazy mode, mode, but if that happens... So for now 100% eager, but over allocated sometimes. So i propose 100 safe for now and add discussion to Force #191 if more relaxed switch used. Again, assume that popular code bases of DB ORMS or some WEB apis use List/Array usually, rather than IReadOnlyList  to have unified optimization of force.

PR #772, opened 2020-12-03
Fix typo in Partition return value documentation

TAGC:
Replaced "staisfying" with "satisfying" in documentation for the `Partition` extension method.


atifaziz (approved):
Thank you!

PR #775, opened 2020-12-18
SelectMany that merges async streams projected from a sequence

atifaziz:
This PR is an implementation of `SelectMany` that projects asynchronous stream from each element of a source sequence and merges the elements of the projected asynchronous stream into a single asynchronous stream result.

By default, the projected asynchronous streams are consumed concurrently. Additional options permit setting maximum desired concurrency.

### Demo

```c#
using System;
using System.Linq;
using System.Threading.Tasks;
using MoreLinq.Experimental.Async;

var q =
    from x in Enumerable.Range(1, 5)
    from e in AsyncEnumerable.Range(x, 5)
                             .Zip(Interval(TimeSpan.FromSeconds(0.2)), (y, _) => (Y: y, Time: DateTime.Now))
    select new { X = x, e.Y, e.Time };

var qq = args.Length > 0
       ? q.MaxConcurrency(int.Parse(args[0]))
       : q;

await foreach (var e in qq)
    Console.WriteLine($"{e.Time:HH:mm:ss.fff}: x = {e.X}, y = {e.Y}");

static async IAsyncEnumerable<int> Interval(TimeSpan interval)
{
    for (var i = 0; ; i++)
    {
        await Task.Delay(interval).ConfigureAwait(false);
        yield return i;
    }
}
```

### DIFF
--- a/MoreLinq/Experimental/Async/AsyncQuery.cs
+++ b/MoreLinq/Experimental/Async/AsyncQuery.cs
@@ -1,0 +1,85 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2020 Atif Aziz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


atifaziz:
The `SelectMany` approach proposed here has too many dimensions to consider for now. A (seemingly) simpler version, `Merge`, that I'd like to start with is being worked on in PR #793 and supersedes this.

PR #776, opened 2020-12-25
Integrate templated code generation into project file

atifaziz:
This PR integrates `dotnet t4` invocation directly into `MoreLinq.csproj` for the following benefits:

- Works with `dotnet build`
- Shell scripts like `tt.*` are not needed anymore
- Uses incremental building built-into MSBuild; `dotnet t4` is not invoked unless one of the `*.g.tt` files is touched

### DIFF
--- a/MoreLinq/Fold.g.tt
+++ b/MoreLinq/Fold.g.tt
@@ -25,6 +25,8 @@ namespace MoreLinq
     using System;
     using System.Collections.Generic;
 
+    // This is a test
+
     partial class MoreEnumerable
     {
 <#      const int max = 16;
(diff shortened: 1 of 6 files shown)


atifaziz:
Unfortunately, parallel builds in MSBuild are incompatible with T4 code generation. It causes sporadic file locking issues when T4 tries to generate code to the same output file in parallel. The project therefore ends in an error if T4 code generation needs running but building is happening in parallel. Re-running with `dotnet build -p:BuildInParallel=false` fixes the issues. This is somewhat inconvenient, but will only happen if someone touches a template file (`*.tt`), forget to transform it so that the output file is in sync and runs `dotnet build` on the command line. In Visual Studio, the output is usually generated as soon as the template is touched and save so the two will generally be in sync. In `build.cmd`/`build.sh`, the T4 code generation is done with the `-p:BuildInParallel=false` switch before the main such that all generated output will be in sync and won't err due to parallel building being the default.

Issue #602, opened 2019-07-17
Move to single CI/CD platform


atifaziz:
In an effort to reduce maintenance, I suggest moving to using AppVeyor for CI/CD exclusively now that it supports [building on Linux](https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/) in addition to Windows.

What will be lost is validating builds against OS X and Mono, but I think that's an acceptable trade-off. If there are issues with either, we can count on people reporting them and then dealing with them on a case-by-case basis.

Another option would be to use [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/) since it supports Windows, Linux and macOS but I am less familiar with that platform and it would require me to spend more time than at the moment (unless someone wants to jump in and help with that).

@fsateler Thoughts?

fsateler:
> In an effort to reduce maintenance

How much maintenance does keeping travis require?

> What will be lost is validating builds against OS X and Mono

The Linux page on appveyor says mono is available. So if that works then it would be only OSX lost.

> Another option would be to use Azure Pipelines since it supports Windows, Linux and macOS but I am less familiar with that platform and it would require me to spend more time than at the moment (unless someone wants to jump in and help with that).

Maybe leave this as a new issue with the help wanted tag.


Anyway, I think it's fine. As time moves on, I think most non windows users be on .net core rather than full net framework.

atifaziz:
> How much maintenance does keeping travis require?

Not much to be perfectly honest, but thinking, one is less than two. For now, I have it sorted for C# 8 for both platforms in PR #582 but it did require tweaking and effort and reading two sets of docs. The end goal is to seek _proficiency_ in one platform, especially for non-functional/infrastructural bits.

> > Another option would be to use Azure Pipelines since it supports Windows, Linux and macOS but I am less familiar with that platform and it would require me to spend more time than at the moment (unless someone wants to jump in and help with that).
> 
> Maybe leave this as a new issue with the help wanted tag.

Good idea! I already added the tags in the first place, but let's leave this as moving to Azure Pipelines.

atifaziz:
Both [AppVeyor](https://www.appveyor.com/blog/2019/11/20/build-macos-projects-with-appveyor/) and [GitHub Actions](https://github.com/features/actions) are now an option, in addition to [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/). AppVeyor would naturally be the path of least work and resistance. 😄

[...]

PR #778, opened 2020-12-26
Unify to single CI platform

atifaziz:
This PR addresses #602 by unifying on AppVeyor as the single CI platform.


appveyor.yml
```diff
@@ -1,24 +1,64 @@
...
+  matrix:
+    only:
+      - image: Visual Studio 2019
+  environment:
+    IMAGE_NAME: win
+  deploy:
+  - provider: NuGet
+    server: https://www.myget.org/F/morelinq/api/v2/package
+    api_key:
+      secure: fhGwXyO35FSshRzs5GWmF1LJTrd1sIqmS/jNCSfO2LfOciuYAKiXuFMYZFGiTAl+
```
fsateler, line 20: Isn't this supposed to be secret?

fsateler (commented):
Hi!

This looks interesting. Did you unify on appveyor only because it was easier? Or for another reason?

I ask, because Github Actions seems to be the new kid on the block everyone is talking about :wink:

appveyor.yml
atifaziz: Yes, but the secret is [secured](https://ci.appveyor.com/tools/encrypt) (as suggested in [Publishing to NuGet feed](https://www.appveyor.com/docs/deployment/nuget/)) and something only AppVeyor's build runner can decrypt.

atifaziz:
> Did you unify on appveyor only because it was easier?

Yes, as I have already some experience with it.

> Or for another reason?

Time. 😄 

> I ask, because Github Actions seems to be the new kid on the block everyone is talking about 😉

I know and I have tried it with just one project so I am barely proficient in it. Everything system needs time getting used to with some trial and error. I have been using AppVeyor for ages and it has served me well. It now supports Linux and macOS images so it was the path of least friction. I am happy to move to GitHub Actions if someone well-versed in it can help with the setup and maintenance. However, given that issue #602 has been open since over a year and no one really picked it up, I just wanted to get this sorted out for now so we can move to more interesting (less infrastructural) work.

fsateler (approved):
LGTM

PR #793, opened 2021-01-14
Merge sequence of async streams into single async stream

atifaziz:
This PR adds `Merge` that concurrently merges all the elements of multiple asynchronous streams into a single asynchronous stream.

An overload allows concurrency to be controlled from 1 (sequential) to some _N_ (or maximum degree of concurrency).

When concurrency is greater than 1, elements in the resulting stream _may_ appear out of order.

If iterations are synchronous in whole or in part then the whole or parts are yielded/streamed synchronously without causing allocations.

This PR supersedes #775.


MoreLinq/Experimental/Async/Merge.cs
```diff
@@ -0,0 +1,251 @@
...
+                        }
+
+                        return default;
+                    }
+
+                    while (enumeratorList.Count > 0)
+                    {
+                        while (pendingTaskList.Count is { } ptc
+                               && ptc < enumeratorList.Count
+                               && ptc < maxConcurrent)
```
viceroypenguin, line 141: enhanced pattern matching?
```suggestion
                        while (pendingTaskList.Count is >= 0
                               && < enumeratorList.Count
                               && < maxConcurrent)
```

viceroypenguin:
Should there also be a version with an explicit `CancellationToken` for those who use fluent syntax instead of `foreach` for processing `IAsyncEnumerable<>`s?

MoreLinq/Experimental/Async/Merge.cs
atifaziz: Thanks, but did you even try that? I'm asking because it's illegal in C#? First of all `&&` would have to be `and` and you cannot use non-literals/constants with relational patterns so while comparing to `>= 0` is fine, you can't do `< enumeratorList.Count` or `< maxConcurrent`.

atifaziz:
> Should there also be a version with an explicit `CancellationToken` for those who use fluent syntax instead of `foreach` for processing `IAsyncEnumerable<>`s?

Most probably, but this is still in draft. :wink: However, the motivation for that isn't so much to satisfy the fluent syntax members of the audience but to allow options to be changed at time of composition or evaluation.

MoreLinq/Experimental/Async/Merge.cs
viceroypenguin: Obviously not... 😉 Definitely I screwed up the `&&` vs `and`. However, I could have sworn I had used a variable value in the the range in some of my code last month. I did test it finally and you're right, each value must be constant. Alas... Excuse me while I go hang my head in shame. 🤦‍♂️😂

viceroypenguin:
Completed in viceroypenguin/SuperLinq#35 - to be released in SuperLinq 4.1.0

PR #800, opened 2021-03-17
Fix dictionary bug when null key has null value

atifaziz:
This PR fixes a bug introduced by PR #582 in our dictionary implementation that accepts `null` keys. The bug can be observed via `ScanBy`:

```c#
var nil = (object)null;
var source = new[] { "foo", null, "bar", null, "baz" };
var result = source.ScanBy(c => c, k => nil, (i, k, e) => nil);
foreach (var item in result)
    Console.WriteLine(item);
```

It prints the following:

    [foo, ]
    [, ]
    [bar, ]

before throwing an `ArgumentNullException`:

    System.ArgumentNullException: Value cannot be null. (Parameter 'key')
       at System.Collections.Generic.Dictionary`2.FindValue(TKey key)
       at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
       at MoreLinq.Collections.Dictionary`2.TryGetValue(TKey key, TValue& value) in C:\projects\morelinq\MoreLinq\Collections\Dictionary.cs:line 69
       at MoreLinq.MoreEnumerable.<>c__DisplayClass226_0`3.<<ScanBy>g___|0>d.MoveNext() in C:\projects\morelinq\MoreLinq\ScanBy.cs:line 100

### DIFF
--- a/MoreLinq.Test/ScanByTest.cs
+++ b/MoreLinq.Test/ScanByTest.cs
@@ -116,6 +116,21 @@ namespace MoreLinq.Test
         }
 
         [Test]
+        public void ScanByWithNullSeed()
+        {
+            var nil = (object)null;
+            var source = new[] { "foo", null, "bar", null, "baz" };
+            var result = source.ScanBy(c => c, k => nil, (i, k, e) => nil);
+
+            result.AssertSequenceEqual(
+                KeyValuePair.Create("foo"       , nil),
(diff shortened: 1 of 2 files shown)


viceroypenguin (approved):

atifaziz:
@viceroypenguin Thanks for the review!

PR #801, opened 2021-03-17
Fix "TrySingle" signature with honest nullable annotations

atifaziz:
There was a to-do comment pending since PR #582 that questioned whether we could be more honest with the signatures of `TrySingle` regarding `T` being `null` when cardinality isn't one. Starting with C# 9, it looks like we can finally address this with [unconstrained type parameter annotations](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/unconstrained-type-parameter-annotations#-annotation):

> In C#8, `?` annotations could only be applied to type parameters that were explicitly constrained to value types or reference types. In C#9, `?` annotations can be applied to any type parameter, regardless of constraints.

This PR fixes the signatures to be more honest and removes the need for the use of the dammit (`!`) operator within the implementation body.

We still can't express that `T` isn't `null` when cardinality is one (e.g. through some of [the helper attributes for the compiler](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis)) but the new annotations are far better than what was there before.

### DIFF
--- a/MoreLinq/Experimental/TrySingle.cs
+++ b/MoreLinq/Experimental/TrySingle.cs
@@ -51,7 +51,7 @@ namespace MoreLinq.Experimental
         /// than two elements from the sequence.
         /// </remarks>
 
-        public static (TCardinality Cardinality, T Value)
+        public static (TCardinality Cardinality, T? Value)
             TrySingle<T, TCardinality>(this IEnumerable<T> source,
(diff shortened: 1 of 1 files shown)


viceroypenguin (approved):

atifaziz:
@viceroypenguin Thanks for the review!

Issue #803, opened 2021-03-18
☂ Review nullability reference annotations


atifaziz:
This is a follow-up “umbrella issue” to track review of nullability reference annotations applied in PR #582. The annotations were slated for release [3.4.0](https://github.com/morelinq/MoreLINQ/milestone/24) but that milestone missed its _desired_ schedule by a big margin. On the positive side, C# 9 got released meanwhile with improvements that could help in revising some of the annotations to be more “honest” (especially with regards to public signatures). The review should therefore focus on:

- benefitting from [unconstrained type parameter annotations](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/unconstrained-type-parameter-annotations#-annotation)
- remove uses of [null-forgiving operator (`!`)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving), if possible, based on improvements in static analysis
- pattern matches on `{}` versus `var`.

PR #582 touched on a number of methods, but the following are the priority for this review:

- [x] `Await`
- [x] `CountBy`
- [x] `EquiZip` (#908)
- [x] `FallbackIfEmpty` (#808)
- [x] `Flatten` (#847)
- [x] `Fold`
- [x] `FullGroupJoin` (#806)
- [x] `GroupAdjacent` (#809)
- [x] `Lag` (#805)
- [x] `Lead` (#805)
- [x] `LeftJoin`
- [x] `MaxBy` (#872)
- [x] `MinBy` (#872)
- [x] `Memoize`
- [x] `OrderBy`
- [x] `OrderedMerge`
- [x] `Pad` (#804)
- [x] `PadStart` (#804)
- [x] `PartialSort` (#888)
- [x] `Partition` (#881)
- [x] `Permutations`
- [x] `Rank` (#913)
- [x] `RightJoin`
- [x] `ScanBy`
- [x] `ToDataTable` (#914)
- [x] `ToDelimitedString`
- [x] `ToDictionary` (#852)
- [x] `TrySingle` (#801)
- [x] `ZipLongest` (#900)
- [x] `ZipShortest` (#908)

/cc @sid-6581

viceroypenguin:
Await does not have any nullability concerns as far as I can tell.

PR #804, opened 2021-03-18
Fix "Pad*" signatures with more honest nullable annotations

atifaziz:
This PR adds to #803. Where a `TSource` is defaulted for padding, the signature now clearly says `TSource?`.


viceroypenguin (approved):
Idle curiosity on NRT handling: I wonder if passing a `Func<int, T>` will return an `IEnumerable<T>` or `IEnumerable<T?>`. Looks good regardless

atifaziz:
> Idle curiosity on NRT handling: I wonder if passing a `Func<int, T>` will return an `IEnumerable<T>` or `IEnumerable<T?>`.

Will return based on whatever `T` is for the instantiation.

> Looks good regardless

Almost. Had one the overloads wrong but hopefully fixed with 58ca7f95c771f762b68fea37ef689895a2506844.

viceroypenguin:
Without the `Defaultable<T>`, it complains? Is it better to add the `Defaultable<T>` or to just `!` it away? Seems like a lot of work to avoid a single character.

atifaziz:
> Without the `Defaultable<T>`, it complains?

Yes.

> Is it better to add the `Defaultable<T>` or to just `!` it away?

I'll admit I'm being sneaky with the compiler, but I am toying with the idea and wanted to throw it here to get reactions from reviewers so thank you for yours!

> Seems like a lot of work to avoid a single character.

JIT should optimise `Defaultable<T>` away 100% so it should have zero runtime impact. Putting that technical detail aside, my goal here was not to save on the single character, but:

1. not having to `default!` at all every call site (usually less important).
2. making it obvious in the (always private) method signature that the argument may be defaulted due to mutual exclusion with some other argument.
3. quickly find all references of `Defaultable<T>` at some point in the future for code analysis.

While it might seem extreme for `Pad`, I am hoping it will help in other cases like `FallbackIfEmpty`. If it ends up being a bad or an over-the-top idea then one can find all references (3) and revert.

PR #805, opened 2021-03-19
Fix "Lag" & "Lead" signatures with more honest nullable annotations

atifaziz:
This PR adds to #803. Where a `TSource` is defaulted for missing lead/lag value, the signature now clearly says `TSource?`.

Unfortunately, this requires a small projection hop (via `Select`) of an [_optuple_](https://github.com/atifaziz/Optuple).

### DIFF
--- a/MoreLinq/Lag.cs
+++ b/MoreLinq/Lag.cs
@@ -19,6 +19,7 @@ namespace MoreLinq
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
 
     public static partial class MoreEnumerable
     {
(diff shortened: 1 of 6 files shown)


MoreLinq/Lead.cs
```diff
@@ -37,9 +38,13 @@ public static partial class MoreEnumerable
         /// <param name="resultSelector">A projection function which accepts the current and subsequent (lead) element (in that order) and produces a result</param>
         /// <returns>A sequence produced by projecting each element of the sequence with its lead pairing</returns>
 
-        public static IEnumerable<TResult> Lead<TSource, TResult>(this IEnumerable<TSource> source, int offset, Func<TSource, TSource, TResult> resultSelector)
+        public static IEnumerable<TResult> Lead<TSource, TResult>(this IEnumerable<TSource> source, int offset, Func<TSource, TSource?, TResult> resultSelector)
         {
-            return Lead(source, offset, default!, resultSelector);
+            if (source is null) throw new ArgumentNullException(nameof(source));
```
viceroypenguin, line 43: Consistency: `is null` vs `== null` in L66-68. Probably prefer `is null`, but I'm good either way.

MoreLinq/Lag.cs
```diff
@@ -36,9 +37,13 @@ public static partial class MoreEnumerable
         /// <param name="resultSelector">A projection function which accepts the current and lagged items (in that order) and returns a result</param>
         /// <returns>A sequence produced by projecting each element of the sequence with its lagged pairing</returns>
 
-        public static IEnumerable<TResult> Lag<TSource, TResult>(this IEnumerable<TSource> source, int offset, Func<TSource, TSource, TResult> resultSelector)
+        public static IEnumerable<TResult> Lag<TSource, TResult>(this IEnumerable<TSource> source, int offset, Func<TSource, TSource?, TResult> resultSelector)
         {
-            return Lag(source, offset, default!, resultSelector);
+            if (source == null) throw new ArgumentNullException(nameof(source));
```
viceroypenguin, line 42: Consistency: `is null` vs `== null`. Probably prefer `is null`, but I'm good either way.

viceroypenguin (approved):

MoreLinq/Lead.cs
atifaziz: You're right, it's inconsistent, but since it probably exists in many places, it's best to roll it into a separate PR.

PR #806, opened 2021-03-19
Fix "FullGroupJoin" signature with more honest nullable annotations

sid-6581:
Adding to morelinq#803. This PR fixes nullable annotations for comparers that are allowed to be null as per the documentation comments.

### DIFF
--- a/MoreLinq/FullGroupJoin.cs
+++ b/MoreLinq/FullGroupJoin.cs
@@ -76,7 +76,7 @@ namespace MoreLinq
             IEnumerable<TSecond> second,
             Func<TFirst, TKey> firstKeySelector,
             Func<TSecond, TKey> secondKeySelector,
-            IEqualityComparer<TKey> comparer)
+            IEqualityComparer<TKey>? comparer)
         {
             return FullGroupJoin(first, second, firstKeySelector, secondKeySelector, ValueTuple.Create, comparer);
         }
@@ -139,7 +139,7 @@ namespace MoreLinq
             Func<TFirst, TKey> firstKeySelector,
             Func<TSecond, TKey> secondKeySelector,
             Func<TKey, IEnumerable<TFirst>, IEnumerable<TSecond>, TResult> resultSelector,
-            IEqualityComparer<TKey> comparer)
+            IEqualityComparer<TKey>? comparer)
         {
             if (first == null) throw new ArgumentNullException(nameof(first));
             if (second == null) throw new ArgumentNullException(nameof(second));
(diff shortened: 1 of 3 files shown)


MoreLinq/FullGroupJoin.cs
```diff
@@ -76,7 +76,7 @@ static partial class MoreEnumerable
             IEnumerable<TSecond> second,
             Func<TFirst, TKey> firstKeySelector,
             Func<TSecond, TKey> secondKeySelector,
-            IEqualityComparer<TKey> comparer)
+            IEqualityComparer<TKey>? comparer)
```
viceroypenguin, line 79: Nullable is not enabled for this file, I don't believe. Maybe should enable before finishing PR?

sid-6581: Nullable is enabled for the whole project in `.csproj`, unless I'm missing something? As far as I can tell it's enabled by default for the project and disabled on a file by file basis for certain files (like `Lookup.cs`).

viceroypenguin: Oh. Sure enough. I was remembering reading the `#nullable enable` in #805, but I see now that's just in the `MoreLinq.Tests` project, and not the main project. I'm good then! :)

MoreLinq/FullGroupJoin.cs
```diff
@@ -151,7 +151,7 @@ static partial class MoreEnumerable
 
             IEnumerable<TResult> _(IEqualityComparer<TKey> comparer)
             {
-                var alookup = Lookup<TKey,TFirst>.CreateForJoin(first, firstKeySelector, comparer);
+                var alookup = Lookup<TKey, TFirst>.CreateForJoin(first, firstKeySelector, comparer);
```
atifaziz, line 154: I know this is just a cosmetic change, but it's orthogonal to this PR so can we revert it?

MoreLinq/FullGroupJoin.cs
atifaziz: Like in #805, how about adding `#nullable enable` in `FullGroupJoinTest.cs` and changing the following line in the test code:

https://github.com/morelinq/MoreLINQ/blob/7458bf13c3e47490abfccdb09f18284654a371dd/MoreLinq.Test/FullGroupJoinTest.cs#L138

to:

```c#
                    return listA.FullGroupJoin(listB, getKey, getKey, ValueTuple.Create, comparer: null);
```

This would ensure that the project doesn't compile is someone removes the annotation by accident.

atifaziz (changes requested):
Thanks for this PR! Just two comments and then I think we're good to merge.

MoreLinq/FullGroupJoin.cs
atifaziz: I've applied this change to `master` with 56b1079c455683cb52522a7e713121a7015949cb. You could either revert this line or just merge with `master` and it should remove this edit.

sid-6581: Apologies for the delay in getting to this! I've had a really busy week wrapping up a project to hand over to QA, but I'm back on this now. I've made the two changes you mentioned, hopefully everything looks good now. I'll catch up on the latest changes and see what else still needs to be done in the rest of the files.

atifaziz: > Apologies for the delay in getting to this! I've had a really busy week wrapping up a project to hand over to QA

No need to apologies. We all get busy. Thanks for the follow-up and looking forward to your upcoming/future contributions.

atifaziz (approved):

PR #807, opened 2021-03-19
Fix "ReverseComparer" signature with more honest nullable annotations

sid-6581:
This PR fixes the nullable annotation for the underlying comparer supplied to the `ReverseComparer` constructor. This class is not part of the public API, but the PR is part of the overall effort of correcting nullable annotations.

### DIFF
--- a/MoreLinq/ReverseComparer.cs
+++ b/MoreLinq/ReverseComparer.cs
@@ -23,7 +23,7 @@ namespace MoreLinq
     {
         readonly IComparer<T> _underlying;
 
-        public ReverseComparer(IComparer<T> underlying)
+        public ReverseComparer(IComparer<T>? underlying)
         {
             _underlying = underlying ?? Comparer<T>.Default;
         }


MoreLinq/ReverseComparer.cs
```diff
@@ -23,7 +23,7 @@ sealed class ReverseComparer<T> : IComparer<T>
     {
         readonly IComparer<T> _underlying;
 
-        public ReverseComparer(IComparer<T> underlying)
+        public ReverseComparer(IComparer<T>? underlying)
```
viceroypenguin, line 26: Nullable is not enabled for this file. Should probably add it before finishing this PR.

atifaziz: It's enabled for the whole project:

https://github.com/morelinq/MoreLINQ/blob/54ccb69f972943a0057a5a546cc795e5e704494b/MoreLinq/MoreLinq.csproj#L123

atifaziz (approved):
Good catch! Thanks!

PR #808, opened 2021-03-19
Cleanup ! operator usage in FallbackIfEmpty

sid-6581:
As part of reviewing nullable reference annotations (#803), this PR attempts to simplify usage of the `!` operator in the internals of `FallbackIfEmpty.cs`. No public APIs are affected.

The following line from the code prior to this PR will cause a compiler warning CS8619 (nullability of reference types in source type doesn't match target type):

```cs
return FallbackIfEmptyImpl(source, 1, fallback, default!, default!, default!, null);
```

Because potential null values are passed to `FallbackIfEmptyImpl`, the return value will be `IEnumerable<T?>`.

One solution would be to use another `!` at the end of the method call. However, the overuse of `!` doesn't feel very clean. Also, the signature of `FallbackIfEmptyImpl` doesn't seem to signal intent as clearly as making the fallback values nullable. This PR also brings some consistency, since the `fallback` collection is already nullable.

Now that the fallback values are nullable, we can remove the `!` from each of the defaults passed to `FallbackIfEmptyImpl`, but we need to use the `!` operator on the fallback values when returning them. This seems a reasonable tradeoff to clean up this file, and should be safe considering the private accessibility of the method.

### DIFF
--- a/MoreLinq/FallbackIfEmpty.cs
+++ b/MoreLinq/FallbackIfEmpty.cs
@@ -46,7 +46,7 @@ namespace MoreLinq
         public static IEnumerable<T> FallbackIfEmpty<T>(this IEnumerable<T> source, T fallback)
         {
             if (source == null) throw new ArgumentNullException(nameof(source));
-            return FallbackIfEmptyImpl(source, 1, fallback, default!, default!, default!, null);
+            return FallbackIfEmptyImpl(source, 1, fallback, default, default, default, null);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


atifaziz (approved):
> Also, the signature of `FallbackIfEmptyImpl` doesn't seem to signal intent as clearly as making the fallback values nullable. This PR also brings some consistency, since the `fallback` collection is already nullable.

Completely agree. This is much better!

Thanks!

---
PS In #804, I introduced `Defaultable<>` that allows getting rid of `!` completely but I think that's just being sneaky and I might revert.

Issue #809, opened 2021-03-22
GroupAdjacent no longer works for null keys


atifaziz:
Using the [latest release (3.3.2)](https://github.com/morelinq/MoreLINQ/releases/tag/v3.3.2), the following code using `GroupAdjacent`:

```c#
var query =
    from g in Enumerable.Range(1, 5)
                        .SelectMany(x => Enumerable.Repeat((int?)x, x)
                                                   .Append(null))
                        .GroupAdjacent(x => x)
    select $"[{g.Key?.ToString() ?? "(null)"}] = [{string.Join(", ", g)}]";

foreach (var e in query)
    Console.WriteLine(e);
```

prints:

    [1] = [1]
    [(null)] = []
    [2] = [2, 2]
    [(null)] = []
    [3] = [3, 3, 3]
    [(null)] = []
    [4] = [4, 4, 4, 4]
    [(null)] = []
    [5] = [5, 5, 5, 5, 5]
    [(null)] = []

The same code against `master` (at b3268c56415ba19a7484b2bd7a8a483845d695ac at time of reporting) prints:

    [1] = [1]
    [2] = [2, 2]
    [3] = [3, 3, 3]
    [4] = [4, 4, 4, 4]
    [5] = [5, 5, 5, 5, 5]

The null groups are completely lost. This is a regression that seems to have creeped in with PR #582. It was spotted as part of extensions listed for review in #803.

PR #810, opened 2021-03-23
Fix "GroupAdjacent" null key handling bug

atifaziz:
This PR fixes the issue reported in #809.


viceroypenguin (approved):
Should you do a follow up to check every place that `{} x` is being used to ensure valid usage?

atifaziz:
> Should you do a follow up to check every place that `{} x` is being used to ensure valid usage?

That's the idea with #803 since it's not just about the type annotations in the public surface, but also within the body of the implementation. Unfortunately, not having a test previously for when a key is `null` means a regression could have been introduced with any refactoring, such as was done with #582.

PR #813, opened 2021-03-28
use badges as link

leandromoh:


atifaziz (commented):
Thanks, but [badges are already there on the home page](https://github.com/morelinq/morelinq.github.io/blob/3f07e9b0becd7a8713f28ea504c31dd4d8fc165f/index.html#L23-L29):

image

If you wish to add some of the new ones you're proposing here (like contributors & downloads) then I'd be happy to take a PR against [the repo for morelinq.github.io](https://github.com/morelinq/morelinq.github.io). I don't see much value in adding them to the README.

viceroypenguin:
I would suggest they be added to readme.md as well. I know many times when I look at projects, I'm browsing on GH, and the readme.md is the first thing I read, and I rarely go to the associated webpage. That's personal experience/opinion though.

leandromoh:
> I would suggest they be added to readme.md as well. I know many times when I look at projects, I'm browsing on GH, and the readme.md is the first thing I read, and I rarely go to the associated webpage.

Same here

PR #820, opened 2021-06-13
Add test to ensure Batch does not reuse internal array

leandromoh:
add test to ensure that if someone try to reuse the internal array in batch (break change) the unit test will not pass


fsateler (approved):
LGTM. Perhaps the name of the test could be a bit more explanatory, but I couldn't come up with a better one, so I'm marking as approved.

atifaziz (approved):
Thanks!

PR #833, opened 2021-11-20
add new batch overload (bucket factory)

leandromoh:
It is useful to have an overload where user can pass the bucket factory. Pratical example:
- user can customize to get arrays from ArrayPool
- user can customize to reuse the array when consume each bucket once, always foward

In both cases are desirable to have the option to custome Batch behaviour, expecially when dealing with large arrays (e.g. +100k elements) where allocate a new array for every new batch is expensive.

Once added this new overload, users are able to create their customizated `Batch` version in their own applications, like the bellow for example.

```c#
public IEnumerable<IEnumerable<T>> BatchFoward<T>(this IEnumerable<T> source, int size)
{
    int lastSize = 0;
    T[] temp = null;

    try
    {
        var result = source.Batch(size, x => x.Take(lastSize), count =>
        {
            lastSize = count;

            return temp ??= ArrayPool<T>.Shared.Rent(count);
        });

        foreach (var batch in result)
            yield return batch;
    }
    finally
    {
        if (temp is { })
            ArrayPool<T>.Shared.Return(temp);
    }
}
```

### DIFF
--- a/MoreLinq/Batch.cs
+++ b/MoreLinq/Batch.cs
@@ -81,9 +81,42 @@ namespace MoreLinq
         public static IEnumerable<TResult> Batch<TSource, TResult>(this IEnumerable<TSource> source, int size,
             Func<IEnumerable<TSource>, TResult> resultSelector)
         {
+            return Batch(source, size, resultSelector, size => new TSource[size]);
+        }
+
+        /// <summary>
+        /// Batches the source sequence into sized buckets and applies a projection to each bucket.
+        /// </summary>
+        /// <typeparam name="TSource">Type of elements in <paramref name="source"/> sequence.</typeparam>
+        /// <typeparam name="TResult">Type of result returned by <paramref name="resultSelector"/>.</typeparam>
+        /// <param name="source">The source sequence.</param>
+        /// <param name="size">Size of buckets.</param>
+        /// <param name="resultSelector">The projection to apply to each bucket.</param>
+        /// <param name="bucketFactory">A function that receive the requested size for the next bucket and return the array where elements will be placed.</param>
+        /// <returns>A sequence of projections on equally sized buckets containing elements of the source collection.</returns>
+        /// <para>
+        /// This operator uses deferred execution and streams its results
+        /// (buckets are streamed but their content buffered).</para>
+        /// <para>
+        /// <para>
+        /// When more than one bucket is streamed, all buckets except the last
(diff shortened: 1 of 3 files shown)


MoreLinq/Batch.cs
```diff
@@ -146,8 +179,9 @@ IEnumerable<TResult> Batch(int size)
                     // Return the last bucket with all remaining elements
                     if (bucket != null && count > 0)
                     {
-                        Array.Resize(ref bucket, count);
-                        yield return resultSelector(bucket);
+                        var newArray = bucketFactory(count);
+                        Array.Copy(bucket, 0, newArray, 0, count);
```
leandromoh, line 183: does the same as `Array.Resize(ref bucket, count)` but manually to use `bucketFactory`

atifaziz (commented):
I can certainly understand the argument for reducing memory pressure when working with very large batch sizes and thus arrays, but I think the proposed solution of simply adding an array factory argument to `Batch` is very tricky to get right for the caller. In the simples of cases, it will lead to incorrect usage and bugs because it requires side effects (like `lastSize` in your example and test), additional work like clipping each bucket (because the actual array length may be greater than the requested bucket size, especially when renting from a pool) and finally returning each array to the pool at the right time for the reuse to be effective through the iterations. You can see that even in the example you posted in [your opening comment](https://github.com/morelinq/MoreLINQ/pull/833#issue-1059028912), only the last bucket is returned to the pool whereas the rest leak. All this also make the composing with `Batch` very hard.

I think a sequence of buckets is probably the wrong return type for such an overload because it claims that each bucket is independent of another when that's not true, especially if the backing array is pooled. Each bucket will end up looking at the _same_ array. This can cause even more severe logical bugs.

I've experimented with what a design could look for a version of `Batch` using pooled arrays and will share it through another PR. Unfortunately, it's not a simple set of changes or tweaks so it's best to share, discuss and review the code in a PR.

atifaziz:
> I've experimented with what a design could look for a version of `Batch` using pooled arrays and will share it through another PR. Unfortunately, it's not a simple set of changes or tweaks so it's best to share, discuss and review the code in a PR.

See #856 and review welcome.

> I think a sequence of buckets is probably the wrong return type

I think I managed to workaround this to still return a sequence.

atifaziz:
This has been superseded by #856.

@leandromoh Happy to get your review in future follow-up issue or PR when you get the time.

PR #843, opened 2022-10-15
Add .NET 6 target while dropping .NET 5 & Core 2.1

atifaziz:
This PR adds the .NET 6 as a target while dropping retired .NET (Core) versions:

- .NET Core 2.1 end-of-life was August 21, 2021
- .NET 5.0 end-of-life was May 10, 2022

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "dotnet-t4": {
-      "version": "2.2.0",
+      "version": "2.3.0",
       "commands": [
         "t4"
       ]
--- a/MoreLinq.Test/MoreLinq.Test.csproj
+++ b/MoreLinq.Test/MoreLinq.Test.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <AssemblyTitle>MoreLinq.Test</AssemblyTitle>
-    <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp2.1;net451</TargetFrameworks>
+    <TargetFrameworks>net6.0;netcoreapp3.1;net451</TargetFrameworks>
     <DebugType>portable</DebugType>
     <AssemblyName>MoreLinq.Test</AssemblyName>
     <OutputType>Exe</OutputType>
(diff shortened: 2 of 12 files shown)


viceroypenguin:
FYI: If you wait a few more weeks, .net 7 will be out and you can add it as a target as well.

atifaziz:
> FYI: If you wait a few more weeks, .net 7 will be out and you can add it as a target as well.

I am well aware, but God knows what my schedule will look like in a few weeks, so refreshing right now to what's there while I have some time to spare. Plus, .NET 6 is LTS so should cover the next two years.

viceroypenguin:
If you don't have time to spare, then might I recommend you update the Usage section of the readme to invite people over to SuperLinq? I will be taking up the mantle of keeping it up to date with the latest versions of .net.

UweKeim:
This release (just pulled from NuGet) broke my [CS-Script](https://github.com/oleg-shilo/cs-script) files (I'm using the last non-.NET-Core version [v3.30.5.2](https://github.com/oleg-shilo/cs-script/releases/tag/v3.30.5.2)) with this (German, sorry) error message:

> System.IO.FileNotFoundException: Die Datei oder Assembly "System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" oder eine Abhängigkeit davon wurde nicht gefunden. Das System kann die angegebene Datei nicht finden.

Basically it tries to use "System.Runtime, Version=6.0.0.0" and complains that it cannot find the files.

I had to revert to Morelinq 3.3.2.

viceroypenguin:
This is actually related to #934. Looking at the CS-Script v3.30.5.2 source, it targets .NET Framework 4.6.0; which is [out of support](https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/) as of April 26, 2022. MoreLinq has moved minimum .NET Framework version support to 4.6.2 to match current Microsoft support.

atifaziz:
> MoreLinq has moved minimum .NET Framework version support to 4.6.2 to match current Microsoft support.

@UweKeim This is something that was called out as a breaking change in the [release notes](https://github.com/morelinq/MoreLINQ/releases/tag/v3.4.0).

PR #854, opened 2022-10-24
Make `ToDataTable` implementation null-safer

atifaziz:
[`Expression` property of `System.Linq.Expressions.MemberExpression`]( https://learn.microsoft.com/en-us/dotnet/api/system.linq.expressions.memberexpression.expression?view=net-6.0#system-linq-expressions-memberexpression-expression) is defined to be nullable. This PR ensures that the property is always access in a null-safe manner; in other words only when it's not `null`.


atifaziz:
This was superseded by PR #860.

PR #880, opened 2022-11-13
Add identity method to reduce JIT-ing

viceroypenguin:
Currently, the C# compiler does not de-dupe methods, including identity methods, such as `x => x`. This means each instance of `x => x` is compiled as a separate method; and each method must be JIT to asm separately, and must be done so for each type separately.

Adding an `Identity` method improves JIT significantly:
* Fewer methods to JIT
* Since these methods are generic, they have to be JITted separately for each type; Identity<T> can share the JITted versions across multiple call sites

Until .net7.0 + tiered PGO, these methods will not get inlined because they are accessed via delegate. This means they also take additional space in memory due to the translated code.

### DIFF
--- a/MoreLinq/Batch.cs
+++ b/MoreLinq/Batch.cs
@@ -50,7 +50,7 @@ namespace MoreLinq
 
         public static IEnumerable<IEnumerable<TSource>> Batch<TSource>(this IEnumerable<TSource> source, int size)
         {
-            return Batch(source, size, Identity<IEnumerable<TSource>>);
+            return Batch(source, size, IdFn<IEnumerable<TSource>>);
         }
 
         /// <summary>
--- a/MoreLinq/Batch.cs
+++ b/MoreLinq/Batch.cs
@@ -50,7 +50,7 @@ namespace MoreLinq
 
         public static IEnumerable<IEnumerable<TSource>> Batch<TSource>(this IEnumerable<TSource> source, int size)
         {
-            return Batch(source, size, IdFn<IEnumerable<TSource>>);
+            return Batch(source, size, IdFn);
         }
 
         /// <summary>
--- a/MoreLinq/Batch.cs
+++ b/MoreLinq/Batch.cs
@@ -50,7 +50,7 @@ namespace MoreLinq
 
         public static IEnumerable<IEnumerable<TSource>> Batch<TSource>(this IEnumerable<TSource> source, int size)
         {
-            return Batch(source, size, x => x);
+            return Batch(source, size, Identity<IEnumerable<TSource>>);
         }
 
         /// <summary>
--- a/MoreLinq/GroupAdjacent.cs
+++ b/MoreLinq/GroupAdjacent.cs
@@ -87,7 +87,7 @@ namespace MoreLinq
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (keySelector == null) throw new ArgumentNullException(nameof(keySelector));
 
-            return GroupAdjacent(source, keySelector, Identity<TSource>, comparer);
+            return GroupAdjacent(source, keySelector, IdFn, comparer);
         }
 
         /// <summary>
@@ -208,7 +208,7 @@ namespace MoreLinq
             // This should be removed once the target framework is bumped to something that supports covariance
             TResult ResultSelectorWrapper(TKey key, IList<TSource> group) => resultSelector(key, group);
 
-            return GroupAdjacentImpl(source, keySelector, Identity<TSource>, ResultSelectorWrapper,
+            return GroupAdjacentImpl(source, keySelector, IdFn, ResultSelectorWrapper,
                                      EqualityComparer<TKey>.Default);
         }
 
@@ -253,7 +253,7 @@ namespace MoreLinq
 
             // This should be removed once the target framework is bumped to something that supports covariance
             TResult ResultSelectorWrapper(TKey key, IList<TSource> group) => resultSelector(key, group);
-            return GroupAdjacentImpl(source, keySelector, Identity<TSource>, ResultSelectorWrapper,
+            return GroupAdjacentImpl(source, keySelector, IdFn, ResultSelectorWrapper,
                                      comparer ?? EqualityComparer<TKey>.Default);
         }
 
--- a/MoreLinq/GroupAdjacent.cs
+++ b/MoreLinq/GroupAdjacent.cs
@@ -87,7 +87,7 @@ namespace MoreLinq
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (keySelector == null) throw new ArgumentNullException(nameof(keySelector));
 
-            return GroupAdjacent(source, keySelector, e => e, comparer);
+            return GroupAdjacent(source, keySelector, Identity<TSource>, comparer);
         }
 
         /// <summary>
@@ -208,7 +208,7 @@ namespace MoreLinq
             // This should be removed once the target framework is bumped to something that supports covariance
             TResult ResultSelectorWrapper(TKey key, IList<TSource> group) => resultSelector(key, group);
 
-            return GroupAdjacentImpl(source, keySelector, i => i, ResultSelectorWrapper,
+            return GroupAdjacentImpl(source, keySelector, Identity<TSource>, ResultSelectorWrapper,
                                      EqualityComparer<TKey>.Default);
         }
 
@@ -253,7 +253,7 @@ namespace MoreLinq
 
             // This should be removed once the target framework is bumped to something that supports covariance
             TResult ResultSelectorWrapper(TKey key, IList<TSource> group) => resultSelector(key, group);
-            return GroupAdjacentImpl(source, keySelector, i => i, ResultSelectorWrapper,
+            return GroupAdjacentImpl(source, keySelector, Identity<TSource>, ResultSelectorWrapper,
                                      comparer ?? EqualityComparer<TKey>.Default);
         }
 
--- a/MoreLinq/Identity.cs
+++ b/MoreLinq/Identity.cs
@@ -1,0 +1,29 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2022 Turning Code, LLC. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq
+{
+    public static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Returns the identity function for a given type.
+        /// </summary>
+        /// <typeparam name="T">The type of identity function</typeparam>
+        /// <returns>A reference to the identity function</returns>
+        public static T Identity<T>(T x) => x;
+    }
+}
--- a/MoreLinq/OrderedMerge.cs
+++ b/MoreLinq/OrderedMerge.cs
@@ -70,7 +70,7 @@ namespace MoreLinq
             IEnumerable<T> second,
             IComparer<T>? comparer)
         {
-            return OrderedMerge(first, second, e => e, f => f, s => s, (a, _) => a, comparer);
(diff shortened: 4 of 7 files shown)


atifaziz (commented):
I don't see the value in adding this because it's not improving anything. It's less ergonomic/succinct (or more verbose) than just a lambda (`x => x`) and it's not removing duplication of any “methods” per the PR subject line. What's more, it doesn't work with anonymous types.

viceroypenguin:
> I don't see the value in adding this because it's not improving anything. It's less ergonomic/succinct (or more verbose) than just a lambda (`x => x`) and it's not removing duplication of any “methods” per the PR subject line. What's more, it doesn't work with anonymous types.

Selected Method:
```csharp
public static IEnumerable<T> OrderedMerge<T, TKey>(
    this IEnumerable<T> first,
    IEnumerable<T> second,
    Func<T, TKey> keySelector)
{
    return OrderedMerge(first, second, keySelector, Identity<T>, Identity<T>, (a, _) => a, null);
}
```

Decompile of `.OrderedMerge()` with lambdas:
```csharp
[Extension]
public static IEnumerable<T> OrderedMerge<[System.Runtime.CompilerServices.Nullable(2)] T, [System.Runtime.CompilerServices.Nullable(2)] TKey>(IEnumerable<T> first, IEnumerable<T> second, Func<T, TKey> keySelector)
{
	return OrderedMerge(first, second, keySelector, <>c__156<T, TKey>.<>9__156_0 ?? (<>c__156<T, TKey>.<>9__156_0 = new Func<T, T>(<>c__156<T, TKey>.<>9.<OrderedMerge>b__156_0)), <>c__156<T, TKey>.<>9__156_1 ?? (<>c__156<T, TKey>.<>9__156_1 = new Func<T, T>(<>c__156<T, TKey>.<>9.<OrderedMerge>b__156_1)), <>c__156<T, TKey>.<>9__156_2 ?? (<>c__156<T, TKey>.<>9__156_2 = new Func<T, T, T>(<>c__156<T, TKey>.<>9.<OrderedMerge>b__156_2)), null);
}
```

Decompile of `.OrderedMerge()` with `Identity<T>`:
```csharp
[Extension]
public static IEnumerable<T> OrderedMerge<[System.Runtime.CompilerServices.Nullable(2)] T, [System.Runtime.CompilerServices.Nullable(2)] TKey>(IEnumerable<T> first, IEnumerable<T> second, Func<T, TKey> keySelector)
{
	return OrderedMerge(first, second, keySelector, <OrderedMerge>O__157_0<T, TKey>.<0>__Identity ?? (<OrderedMerge>O__157_0<T, TKey>.<0>__Identity = new Func<T, T>(Identity)), <OrderedMerge>O__157_0<T, TKey>.<0>__Identity ?? (<OrderedMerge>O__157_0<T, TKey>.<0>__Identity = new Func<T, T>(Identity)), <>c__157<T, TKey>.<>9__157_0 ?? (<>c__157<T, TKey>.<>9__157_0 = new Func<T, T, T>(<>c__157<T, TKey>.<>9.<OrderedMerge>b__157_0)), null);
}
```

In the original version, the two identity lambdas are compiled separately (`<>c__156<T, TKey>.<>9.<OrderedMerge>b__156_0` and `<>c__156<T, TKey>.<>9.<OrderedMerge>b__156_1`). In the new version, only one method is compiled: `Identity<T>`. 

This improves JIT significantly:
* Fewer methods to JIT
* Since these methods are generic, they have to be JITted separately for each type; `Identity<T>` can share the JITted versions across multiple call sites

Until .net7.0 + tiered PGO, these methods will not get inlined because they are accessed via delegate. This means they also take additional space in memory due to the translated code. 

Also, it works fine with anonymous types - see LinqPAD attachment here:
image

MoreLinq/Identity.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace MoreLinq
+{
+    public static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Returns the identity function for a given type.
+        /// </summary>
+        /// <typeparam name="T">The type of identity function</typeparam>
+        /// <returns>A reference to the identity function</returns>
+        public static T Identity<T>(T x) => x;
```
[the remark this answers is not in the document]
viceroypenguin: Yes, the advantages of compiling the `Identity<T>` method once should be available to consumers if desired.

atifaziz:
@viceroypenguin Thanks for the [detailed explanation](https://github.com/morelinq/MoreLINQ/pull/880#issuecomment-1314520597). I understand now the savings you're trying to make with this and it would have great to have the justification in [the initial description](https://github.com/morelinq/MoreLINQ/pull/880#issue-1447141161). I'll come back to you on this, hopefully before the week is over.

MoreLinq/Batch.cs
```diff
@@ -50,7 +50,7 @@ static partial class MoreEnumerable
 
         public static IEnumerable<IEnumerable<TSource>> Batch<TSource>(this IEnumerable<TSource> source, int size)
         {
-            return Batch(source, size, x => x);
+            return Batch(source, size, Identity<IEnumerable<TSource>>);
```
atifaziz, line 53: Type argument is redundant:

```suggestion
            return Batch(source, size, Identity);
```

MoreLinq/GroupAdjacent.cs
```diff
@@ -87,7 +87,7 @@ static partial class MoreEnumerable
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (keySelector == null) throw new ArgumentNullException(nameof(keySelector));
 
-            return GroupAdjacent(source, keySelector, e => e, comparer);
+            return GroupAdjacent(source, keySelector, Identity<TSource>, comparer);
```
atifaziz, line 90: Type argument is redundant:

```suggestion
            return GroupAdjacent(source, keySelector, Identity, comparer);
```

MoreLinq/GroupAdjacent.cs
```diff
@@ -208,7 +208,7 @@ static partial class MoreEnumerable
             // This should be removed once the target framework is bumped to something that supports covariance
             TResult ResultSelectorWrapper(TKey key, IList<TSource> group) => resultSelector(key, group);
 
-            return GroupAdjacentImpl(source, keySelector, i => i, ResultSelectorWrapper,
+            return GroupAdjacentImpl(source, keySelector, Identity<TSource>, ResultSelectorWrapper,
```
atifaziz, line 211: Type argument is redundant:

```suggestion
            return GroupAdjacentImpl(source, keySelector, Identity, ResultSelectorWrapper,
```

MoreLinq/GroupAdjacent.cs
```diff
@@ -253,7 +253,7 @@ static partial class MoreEnumerable
 
             // This should be removed once the target framework is bumped to something that supports covariance
             TResult ResultSelectorWrapper(TKey key, IList<TSource> group) => resultSelector(key, group);
-            return GroupAdjacentImpl(source, keySelector, i => i, ResultSelectorWrapper,
+            return GroupAdjacentImpl(source, keySelector, Identity<TSource>, ResultSelectorWrapper,
```
atifaziz, line 256: Type argument is redundant:

```suggestion
            return GroupAdjacentImpl(source, keySelector, Identity, ResultSelectorWrapper,
```

MoreLinq/OrderedMerge.cs
```diff
@@ -70,7 +70,7 @@ static partial class MoreEnumerable
             IEnumerable<T> second,
             IComparer<T>? comparer)
         {
-            return OrderedMerge(first, second, e => e, f => f, s => s, (a, _) => a, comparer);
+            return OrderedMerge(first, second, Identity<T>, Identity<T>, Identity<T>, (a, _) => a, comparer);
```
atifaziz, line 73: Type argument is redundant:

```suggestion
            return OrderedMerge(first, second, Identity, Identity, Identity, (a, _) => a, comparer);
```

MoreLinq/OrderedMerge.cs
```diff
@@ -98,7 +98,7 @@ static partial class MoreEnumerable
             IEnumerable<T> second,
             Func<T, TKey> keySelector)
         {
-            return OrderedMerge(first, second, keySelector, a => a, b => b, (a, _) => a, null);
+            return OrderedMerge(first, second, keySelector, Identity<T>, Identity<T>, (a, _) => a, null);
```
atifaziz, line 101: Type argument is redundant:

```suggestion
            return OrderedMerge(first, second, keySelector, Identity, Identity, (a, _) => a, null);
```

MoreLinq/Rank.cs
```diff
@@ -32,7 +32,7 @@ public static partial class MoreEnumerable
 
         public static IEnumerable<int> Rank<TSource>(this IEnumerable<TSource> source)
         {
-            return source.RankBy(x => x);
+            return source.RankBy(Identity<TSource>);
```
atifaziz, line 35: Type argument is redundant:

```suggestion
            return source.RankBy(Identity);
```

MoreLinq/Rank.cs
```diff
@@ -45,7 +45,7 @@ public static IEnumerable<int> Rank<TSource>(this IEnumerable<TSource> source)
 
         public static IEnumerable<int> Rank<TSource>(this IEnumerable<TSource> source, IComparer<TSource> comparer)
         {
-            return source.RankBy(x => x, comparer);
+            return source.RankBy(Identity<TSource>, comparer);
```
atifaziz, line 48: Type argument is redundant:

```suggestion
            return source.RankBy(Identity, comparer);
```

MoreLinq/Identity.cs
```diff
@@ -0,0 +1,29 @@
...
+#endregion
+
+namespace MoreLinq
+{
+    public static partial class MoreEnumerable
+    {
+        /// <summary>
+        /// Returns the identity function for a given type.
+        /// </summary>
+        /// <typeparam name="T">The type of identity function</typeparam>
```
atifaziz, line 25: Period:

```suggestion
        /// <typeparam name="T">The type of identity function.</typeparam>
```

MoreLinq/Identity.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <returns>A reference to the identity function</returns>
```
atifaziz, line 26: Period + blank line:

```suggestion
        /// <returns>A reference to the identity function.</returns>

```

MoreLinq/Identity.cs
atifaziz, line 27: Make private:

```suggestion
        static T Identity<T>(T x) => x;
```

---
PS You could also skip the doc comments, which are not required for private members.

PSS As a private method, I would also strongly recommend renaming the method to `IdFn` or `IdFunc`.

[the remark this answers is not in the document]
atifaziz: The identity function is already a micro optimization and while it's fair to use within the library, I don't think offering it as a public API is worth committing to it. It would be the first stand-alone helper function that could open a can of worms. Ideally this is something that should be part of .NET proper. We can keep it private for now and always review the decision.

atifaziz:
> Also, it works fine with anonymous types - see LinqPAD attachment here:

My main concern was as a method group _within_ a query where you're expecting to use `Identity` (and not as an argument to the method as in your LINQPad example), but it works fine:

```c#
var map =
    Enumerable.Range(1, 10)
              .Select(x => new { X = x, Y = x * 2 })
              .ToDictionary(x => x.X, Identity);
```

I was somewhat misled during my initial review by the explicit type annotations like `Identity<TSource>` and `Identity<IEnumerable<TSource>>`, where you still specify the generic type parameter, but coming back to it, those are all redundant.

atifaziz (changes requested):
Thanks for adding this. I think you missed another opportunity to use identity function in `Split`?

viceroypenguin:
> I was somewhat misled during my initial review by the explicit type annotations like `Identity<TSource>` and `Identity<IEnumerable<TSource>>`, where you still specify the generic type parameter, but coming back to it, those are all redundant.

Ah, right. Totally missed that. Amazing how powerful the typing system is sometimes. :)

MoreLinq/Batch.cs
```diff
@@ -50,7 +50,7 @@ static partial class MoreEnumerable
 
         public static IEnumerable<IEnumerable<TSource>> Batch<TSource>(this IEnumerable<TSource> source, int size)
         {
-            return Batch(source, size, x => x);
+            return Batch(source, size, IdFn<IEnumerable<TSource>>);
```
atifaziz, line 53: Remove redundant type arguments:

```suggestion
            return Batch(source, size, IdFn);
```

atifaziz (approved):
Thanks, and nearly there! 🏁

MoreLinq/Batch.cs
viceroypenguin: My regex-foo failed me! :)

PR #881, opened 2022-11-13
Cleanup `Partition` Nullability

viceroypenguin:
This PR adds to https://github.com/morelinq/MoreLINQ/issues/803. It uses nullable `TKey` values in the implementation to more accurately express the code.

### DIFF
--- a/MoreLinq.Test/PartitionTest.cs
+++ b/MoreLinq.Test/PartitionTest.cs
@@ -15,6 +15,8 @@
 // limitations under the License.
 #endregion
 
+#nullable enable
+
 namespace MoreLinq.Test
 {
     using System;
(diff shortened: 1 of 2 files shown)


atifaziz (changes requested):
Like PR #882, this make unrelated changes (doc _and_ nullability revisions). Please submit one overall idea/feature per PR.

atifaziz (changes requested):
Thanks, but the bulk of changes seem to be unrelated (and I've highlighted a few):

- Formatting changes
- More eager argument validation
- Annotation of some arguments with parameter names.

>  uses nullable `TKey` values in the implementation to more accurately express the code.

This is the only change that should be in here, which is towards the end of `Partition.cs`.

I would avoid the temptation to bundle unrelated changes as it has the following downsides:

- The diff/patch is larger than it needs to be.
- As a result of the above point, it takes longer to review the PR.
- The PR subject line or the commit message won't reflect the actual change (lost in the noise).
- The final commit won't be atomic so, e.g., if someone wants to revert it in the future, it will undo more work than necessary.

atifaziz (approved):
Thank you!

PR #882, opened 2022-11-13
`ScanBy`: Update docs

viceroypenguin:
This PR adds to https://github.com/morelinq/MoreLINQ/issues/803. It updates documentation. `ScanBy` does not need other changes to address nullability documentation.

### DIFF
--- a/MoreLinq.Test/ScanByTest.cs
+++ b/MoreLinq.Test/ScanByTest.cs
@@ -15,6 +15,8 @@
 // limitations under the License.
 #endregion
 
+#nullable enable
+
 namespace MoreLinq.Test
 {
     using System;
@@ -48,10 +50,10 @@ namespace MoreLinq.Test
             };
 
             var result =
-                    source.ScanBy(
-                        item => item.First(),
-                        key => (Element: default(string), Key: key, State: key - 1),
-                        (state, key, item) => (item, char.ToUpperInvariant(key), state.State + 1));
+                source.ScanBy(
+                    item => item.First(),
+                    key => (Element: string.Empty, Key: key, State: key - 1),
+                    (state, key, item) => (item, char.ToUpperInvariant(key), state.State + 1));
 
             result.AssertSequenceEqual(
                KeyValuePair.Create('a', ("ana",     'A', 97)),
@@ -103,31 +105,31 @@ namespace MoreLinq.Test
             var result = source.ScanBy(c => c, _ => -1, (i, _, _) => i + 1);
 
             result.AssertSequenceEqual(
-                KeyValuePair.Create("foo"       , 0),
-                KeyValuePair.Create((string)null, 0),
-                KeyValuePair.Create("bar"       , 0),
-                KeyValuePair.Create("baz"       , 0),
-                KeyValuePair.Create((string)null, 1),
-                KeyValuePair.Create((string)null, 2),
-                KeyValuePair.Create("baz"       , 1),
-                KeyValuePair.Create("bar"       , 1),
-                KeyValuePair.Create((string)null, 3),
-                KeyValuePair.Create("foo"       , 1));
+                KeyValuePair.Create<string?, int>("foo", 0),
+                KeyValuePair.Create<string?, int>(null , 0),
+                KeyValuePair.Create<string?, int>("bar", 0),
+                KeyValuePair.Create<string?, int>("baz", 0),
+                KeyValuePair.Create<string?, int>(null , 1),
+                KeyValuePair.Create<string?, int>(null , 2),
+                KeyValuePair.Create<string?, int>("baz", 1),
+                KeyValuePair.Create<string?, int>("bar", 1),
+                KeyValuePair.Create<string?, int>(null , 3),
+                KeyValuePair.Create<string?, int>("foo", 1));
         }
 
         [Test]
         public void ScanByWithNullSeed()
         {
-            var nil = (object)null;
+            var nil = (object?)null;
             var source = new[] { "foo", null, "bar", null, "baz" };
             var result = source.ScanBy(c => c, _ => nil, (_, _, _) => nil);
 
             result.AssertSequenceEqual(
-                KeyValuePair.Create("foo"       , nil),
-                KeyValuePair.Create((string)null, nil),
-                KeyValuePair.Create("bar"       , nil),
-                KeyValuePair.Create((string)null, nil),
-                KeyValuePair.Create("baz"       , nil));
+                KeyValuePair.Create<string?, object?>("foo", nil),
+                KeyValuePair.Create<string?, object?>(null , nil),
+                KeyValuePair.Create<string?, object?>("bar", nil),
+                KeyValuePair.Create<string?, object?>(null , nil),
+                KeyValuePair.Create<string?, object?>("baz", nil));
         }
 
         [Test]
--- a/MoreLinq/ScanBy.cs
+++ b/MoreLinq/ScanBy.cs
@@ -23,26 +23,28 @@ namespace MoreLinq
     static partial class MoreEnumerable
     {
         /// <summary>
-        /// Applies an accumulator function over sequence element keys, returning the keys along with intermediate
-        /// accumulator states.
+        /// Applies an accumulator function over sequence element keys,
+        /// returning the keys along with intermediate accumulator states.
         /// </summary>
-        /// <typeparam name="TSource">Type of the elements of the source sequence.</typeparam>
+        /// <typeparam name="TSource">Type of the elements of the source
+        /// sequence.</typeparam>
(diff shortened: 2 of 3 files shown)


MoreLinq/ScanBy.cs
```diff
@@ -40,14 +40,17 @@ static partial class MoreEnumerable
         /// <returns>
         /// A sequence of keys paired with intermediate accumulator states.
         /// </returns>
-
+        /// <exception cref="ArgumentNullException"><paramref name="source"/> is null</exception>
```
atifaziz, line 43: Since it's the same exception listed in each case, you just want one `<exception>` block instead:

```c#
/// <exception cref="ArgumentNullException">
/// <paramref name="source"/>, <paramref name="keySelector"/>,
/// <paramref name="seedSelector"/> or <paramref name="accumulator"/> is
/// <see langword="keyword"/>.
/// </exception>
```

---
PS [See also how it's done for `Union`](https://github.com/dotnet/runtime/blob/v7.0.0/src/libraries/System.Linq/src/System/Linq/Union.cs#L52).

MoreLinq/ScanBy.cs
```diff
@@ -40,14 +40,17 @@ static partial class MoreEnumerable
         /// <returns>
         /// A sequence of keys paired with intermediate accumulator states.
         /// </returns>
-
```
atifaziz, line 43: Please keep the blank line for clarity.

MoreLinq/ScanBy.cs
```diff
@@ -40,14 +40,17 @@ static partial class MoreEnumerable
...
+        /// <exception cref="ArgumentNullException"><paramref name="seedSelector"/> is null</exception>
+        /// <exception cref="ArgumentNullException"><paramref name="accumulator"/> is null</exception>
         public static IEnumerable<KeyValuePair<TKey, TState>> ScanBy<TSource, TKey, TState>(
             this IEnumerable<TSource> source,
             Func<TSource, TKey> keySelector,
             Func<TKey, TState> seedSelector,
             Func<TState, TKey, TSource, TState> accumulator)
         {
-            return source.ScanBy(keySelector, seedSelector, accumulator, null);
+            return source.ScanBy(keySelector, seedSelector, accumulator, comparer: null);
```
atifaziz, line 53: Is the argument name being added for clarity?

MoreLinq/ScanBy.cs
```diff
@@ -72,7 +75,10 @@ static partial class MoreEnumerable
         /// <returns>
         /// A sequence of keys paired with intermediate accumulator states.
         /// </returns>
-
+        /// <exception cref="ArgumentNullException"><paramref name="source"/> is null</exception>
```
atifaziz, line 78: Ditto as [earlier comment](https://github.com/morelinq/MoreLINQ/pull/882/files#r1022080365).

atifaziz (changes requested):
Unfortunately, this PR makes two unrelated changes. Can we please have each in their own PR?

MoreLinq/ScanBy.cs
viceroypenguin: I would argue that it reduces clarity (by adding separation between the documentation and the method), but I'll defer to you.

MoreLinq/ScanBy.cs
viceroypenguin: Yes.

viceroypenguin:
> I think it reduces an allocation, but not the capture size? Either way, it would be best done as a sweeping change in one PR?

Sure.

MoreLinq.Test/ScanByTest.cs
```diff
@@ -48,10 +50,10 @@ public void ScanBy()
             };
 
             var result =
-                    source.ScanBy(
-                        item => item.First(),
-                        key => (Element: default(string), Key: key, State: key - 1),
-                        (state, key, item) => (item, char.ToUpperInvariant(key), state.State + 1));
+                source.ScanBy(
+                    item => item.First(),
+                    key => (Element: string.Empty, Key: key, State: key - 1),
```
viceroypenguin, line 55: Using `default(string?)` (rightfully) changes the element type to `string?`, which would then have to be handled below. `string.Empty` keeps type as `string` instead of `string?`.

MoreLinq.Test/ScanByTest.cs
```diff
@@ -103,31 +105,31 @@ public void ScanByWithSomeNullKeys()
...
-                KeyValuePair.Create((string)null, 0),
-                KeyValuePair.Create("bar"       , 0),
-                KeyValuePair.Create("baz"       , 0),
-                KeyValuePair.Create((string)null, 1),
-                KeyValuePair.Create((string)null, 2),
-                KeyValuePair.Create("baz"       , 1),
-                KeyValuePair.Create("bar"       , 1),
-                KeyValuePair.Create((string)null, 3),
-                KeyValuePair.Create("foo"       , 1));
+                KeyValuePair.Create<string?, int>("foo", 0),
```
viceroypenguin, line 108: Type explicit to improve nullability behavior since `KeyValuePair.Create()` enforces a nullability value without target-typing to the sequence type.

atifaziz: You could go with typing just the one needed:

```c#
            result.AssertSequenceEqual(
                KeyValuePair.Create((string?)"foo", 0),
                KeyValuePair.Create((string?)null , 0),
                KeyValuePair.Create((string?)"bar", 0),
                KeyValuePair.Create((string?)"baz", 0),
                KeyValuePair.Create((string?)null , 1),
                KeyValuePair.Create((string?)null , 2),
                KeyValuePair.Create((string?)"baz", 1),
                KeyValuePair.Create((string?)"bar", 1),
                KeyValuePair.Create((string?)null , 3),
                KeyValuePair.Create((string?)"foo", 1));
```

MoreLinq.Test/ScanByTest.cs
```diff
@@ -103,31 +105,31 @@ public void ScanByWithSomeNullKeys()
...
             var source = new[] { "foo", null, "bar", null, "baz" };
             var result = source.ScanBy(c => c, _ => nil, (_, _, _) => nil);
 
             result.AssertSequenceEqual(
-                KeyValuePair.Create("foo"       , nil),
-                KeyValuePair.Create((string)null, nil),
-                KeyValuePair.Create("bar"       , nil),
-                KeyValuePair.Create((string)null, nil),
-                KeyValuePair.Create("baz"       , nil));
+                KeyValuePair.Create<string?, object?>("foo", nil),
```
atifaziz, line 128: Ditto as [above](https://github.com/morelinq/MoreLINQ/pull/882/commits/8520a3444bbea465151b39558c0b16a732134d66#r1024593862):

```c#
            result.AssertSequenceEqual(
                KeyValuePair.Create((string?)"foo", nil),
                KeyValuePair.Create((string?)null , nil),
                KeyValuePair.Create((string?)"bar", nil),
                KeyValuePair.Create((string?)null , nil),
                KeyValuePair.Create((string?)"baz", nil));
```

MoreLinq.Test/ScanByTest.cs
viceroypenguin: I considered that, but I honestly like the other one better. Generally, to me, an explicit type cast like this feels like I may have done something wrong to need the cast, whereas target-typing using the explicit type of `.Create()` feels more natural. Just my opinion though.

atifaziz:
@viceroypenguin Any [follow-up](https://github.com/morelinq/MoreLINQ/pull/882#pullrequestreview-1179880632)?

viceroypenguin:
I'm not sure what your concern is. Per
> Unfortunately, this PR makes two unrelated changes. Can we please have each in their own PR?

the "reduces capture size" change was reverted.  This PR was also updated per your request to include
> Also, if ScanBy doesn't need changes with respect to #803 then let's enable nullable context on its unit tests, which is a good way to exercise its signature. 

There is an unanswered expressed opinion [here](https://github.com/morelinq/MoreLINQ/pull/882#discussion_r10246804720), but otherwise the only thing necessary for this PR is to resolve conflicts with latest.

atifaziz:
> I'm not sure what your concern is.

No concern. Just this:

> Unfortunately, this PR makes two unrelated changes. Can we please have each in their own PR?

Title of this PR reads says it updates the documentation (unrelated to #803), but it does more.

viceroypenguin:
If you resolve conflicts with mainline, the additional changes will work themselves out. I'll let you take care of that at your leisure.

atifaziz:
> If you resolve conflicts with mainline, the additional changes will work themselves out. I'll let you take care of that at your leisure.

This is your contribution. If you don't wish to or lost the motivation to see it through, including resolving conflicts and undoing unrelated changes, then that's fine. I'll go ahead and close it for now but happy to re-open if you'd like to help.

PR #883, opened 2022-11-14
Simplify Zip* Implementations

viceroypenguin:
This PR is for #803. It does two things: 1) It reimplements `EquiZip`, `ZipLongest`, and `ZipShortest` using simpler implementations; 2) It cleans up nullability constraints for those methods.


MoreLinq.Test/TestingSequence.cs
```diff
@@ -72,13 +72,15 @@ public IEnumerator<T> GetEnumerator()
                 Assert.That(_disposed, Is.False, "LINQ operators should not dispose a sequence more than once.");
                 _disposed = true;
             };
-            var ended = false;
             enumerator.MoveNextCalled += (_, moved) =>
             {
-                Assert.That(ended, Is.False, "LINQ operators should not continue iterating a sequence that has terminated.");
```
viceroypenguin, line 78: Per the [specification](https://learn.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.movenext?view=net-7.0#remarks), it is not a failure to call `.MoveNext()` after receiving a `false` response. The enumerator is simply expected to continue to return `false` for each following call.

MoreLinq.Test/ZipLongestTest.cs
```diff
@@ -58,21 +58,6 @@ public void ZipLongestIsLazy()
             bs.ZipLongest(bs, BreakingFunc.Of<int, int, int>());
         }
 
-        [Test]
-        public void ZipLongestDisposeSequencesEagerly()
```
viceroypenguin, line 62: Removed as simpler code does not dispose immediately, but instead keeps calling `.MoveNext()` expecting `false` return values. Is there a specific advantage to disposing early?

MoreLinq/ZipImpl.cs
```diff
@@ -1,97 +0,0 @@
...
-{
-    using System;
-    using System.Collections.Generic;
-    using System.Collections;
-
-    static partial class MoreEnumerable
-    {
-        delegate TResult Folder<in T, out TResult>(params T[] args);
-
-        static IEnumerable<TResult> ZipImpl<T1, T2, T3, T4, TResult>(
```
viceroypenguin, line 28: Removed as code is simpler when re-implemented for equi-, -shortest, and -longest.

PR #890, opened 2022-11-15
Replace null-forgiving operator uses with debug assertions

atifaziz:
This PR introduces `Debug.Assert` that's defined similarly to [`System.Diagnostics.Debug.Assert`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert?view=net-7.0#system-diagnostics-debug-assert(system-boolean-system-string)) to help eliminate the bulk uses of the null-forgiving operator (`!`). The assertion more explicitly states the illegal case, and if triggered, will also relay the offending condition in the message. It also eliminates needing to suppress [CS8602](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings#possible-dereference-of-null) and [CS8603](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings#possible-null-assigned-to-a-nonnullable-reference) in target earlier than .NET 6.

This PR also adds `Assume.NotNull`, that's a `Debug.Assert` in disguise, but which specifically asserts the not-null-reference case. The idea is to make these cases easily searchable in code.

### DIFF
--- a/MoreLinq.Test/FallbackIfEmptyTest.cs
+++ b/MoreLinq.Test/FallbackIfEmptyTest.cs
@@ -60,5 +60,13 @@ namespace MoreLinq.Test
             Assert.AreSame(source.FallbackIfEmpty(fallback), fallback);
             Assert.AreSame(source.FallbackIfEmpty(fallback.AsEnumerable()), fallback);
         }
+
+        [Test]
+        public void FallbackIfEmptyWithEmptyNullableSequence()
+        {
+            var source = Enumerable.Empty<int?>().Select(x => x);
+            var fallback = (int?)null;
+            source.FallbackIfEmpty(fallback).AssertSequenceEqual(fallback);
+        }
     }
 }
--- a/MoreLinq/Assume.cs
+++ b/MoreLinq/Assume.cs
@@ -1,0 +1,31 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2022 Atif Aziz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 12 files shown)


viceroypenguin:
While I am not opposed to `Debug.Assert(bool condition)`, I would _suggest_ adding a `static T AssertNotNull<T>([NotNull, AllowNull] T? object) where T : class`, which receives an object, throws if null, and returns the object unchanged otherwise. The nullability analyzer will treat the returned object as not-null, which means it can be used immediately.

Example use case:
```csharp
var g = Debug.AssertNotNull(_lastGrouping);
```

atifaziz:
> While I am not opposed to `Debug.Assert(bool condition)`, I would _suggest_ adding a `static T AssertNotNull<T>([NotNull, AllowNull] T? object) where T : class`, which receives an object, throws if null, and returns the object unchanged otherwise. The nullability analyzer will treat the returned object as not-null, which means it can be used immediately.
> 
> Example use case:
> 
> ```cs
> var g = Debug.AssertNotNull(_lastGrouping);
> ```

You can't do that because such assertion methods are _conditional methods_ that get compiled out of release builds and therefore they must return `void`. See also the [remarks section of the `ConditionalAttribute` docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?view=net-7.0#remarks):

> Applying [ConditionalAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?view=net-7.0) to a method indicates to compilers that a call to the method should not be compiled into Microsoft intermediate language (MSIL) unless the conditional compilation symbol that is associated with [ConditionalAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?view=net-7.0) is defined. _You will get a compilation error in Visual Studio if you apply this attribute to a method that does not return void.

viceroypenguin:
Ah, I missed the `[Conditional]`. I'll ponder this further.

atifaziz:
> which receives an object, throws if null, and returns the object unchanged otherwise

I've added this idea with 7e6fa33d1ea58325480998943cbfcf8cb9ca9a33 and updated the [initial description](https://github.com/morelinq/MoreLINQ/pull/890#issue-1450492936) with:

> This PR also adds `Assume.NotNull`, that's a `Debug.Assert` in disguise, but which specifically asserts the not-null-reference case. The idea is to make these cases easily searchable in code.

viceroypenguin:
Finished product LGTM

PR #894, opened 2022-11-19
Rewrite switch in `Memoize` as expression

viceroypenguin:
~This PR is for https://github.com/morelinq/MoreLINQ/issues/803.~ It updates to use the new switch expression for the operator.

### DIFF
--- a/MoreLinq/Experimental/Memoize.cs
+++ b/MoreLinq/Experimental/Memoize.cs
@@ -47,17 +47,15 @@ namespace MoreLinq.Experimental
         /// (though never simultaneously) may iterate over the sequence.
         /// </remarks>
 
-        public static IEnumerable<T> Memoize<T>(this IEnumerable<T> source)
-        {
-            switch (source)
+        public static IEnumerable<T> Memoize<T>(this IEnumerable<T> source) =>
+            source switch
             {
-                case null: throw new ArgumentNullException(nameof(source));
-                case ICollection<T>        : // ...
-                case IReadOnlyCollection<T>: // ...
-                case MemoizedEnumerable<T> : return source;
-                default: return new MemoizedEnumerable<T>(source);
-            }
-        }
+                null => throw new ArgumentNullException(nameof(source)),
+                ICollection<T> => source,
+                IReadOnlyCollection<T> => source,
+                MemoizedEnumerable<T> => source,
+                _ => new MemoizedEnumerable<T>(source),
+            };
     }
 
     sealed class MemoizedEnumerable<T> : IEnumerable<T>, IDisposable


atifaziz (approved):
Thanks!

MoreLinq/Experimental/Memoize.cs
```diff
@@ -47,17 +47,15 @@ static partial class ExperimentalEnumerable
...
-                case ICollection<T>        : // ...
-                case IReadOnlyCollection<T>: // ...
-                case MemoizedEnumerable<T> : return source;
-                default: return new MemoizedEnumerable<T>(source);
-            }
-        }
+                null => throw new ArgumentNullException(nameof(source)),
+                ICollection<T> => source,
+                IReadOnlyCollection<T> => source,
+                MemoizedEnumerable<T> => source,
```
atifaziz, line 56: Consolidate:

```suggestion
                ICollection<T> or IReadOnlyCollection<T> or MemoizedEnumerable<T> => source,
```

viceroypenguin: Considered - personal opinion is this is more readable. Up to you.

atifaziz: I can imagine, but versus duplication (albeit a simple one) of the right side?

viceroypenguin: Did a quick test on [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8lAbhvoYGUALbKAA4AZbMAB0AJQCuAOwwBLfDHa0AzE3JImpBgGEGAbxoMTDANoApeRgDiMGTCjywACgwBPATAgAzF/LkASkCAXWNTCOJ1Zi1mVQAeABUAPgYAWRh8CHkALxgk5LceeVwNBJSGXAgpKDAYQIYAXmTwiLbK6tqYSoB3azAeVvbTI2ph8YYZKQAbaabUjB4oCB7JmFWAQSgAcyklOQA5GemAUQQ6gQUIGRcZbCVfFyqauuC0IYnTAEldCFmYMBXGQFBjQBhfCQwbAAEwA8jJpu5fv9AfJriDmh0XjB3mNPhMAPrzBiLZarByrA4QDBffACaaZewYGDQs4XIEuQK4gC+Km5QA==) - the difference between the two versions is not significantly different in the IL. Then, the JIT identifies the `return` is the same and funnels success path to the same `ret` statement either way.

atifaziz: I have no doubts that the difference disappears in the JIT'd release code, but my point is about duplication of the right side of the expression in the source (what it eventually compiles down to could change with compiler/runtime versions and shouldn't be used as an absolute basis). For this case, it's super simple; it's always `source`. However, had the expression on the right side been even slightly more complex, I wonder if readability would have won over duplication. Anyway, this is not a blocker for merging. I just wanted to point out that it's not a strict translation of the `switch` statement that was there, just in case you'd agree that the parity should be maintained.

viceroypenguin: Agreed. Anything more complicated would definitely have changed the balance.

PR #896, opened 2022-11-19
Add null check for `Partition`

viceroypenguin:
This introduces a null-check in order to fail-fast on a `null` `resultSelector` value.


atifaziz (approved):
Thanks!

PR #897, opened 2022-11-19
Add parameter annotations for `Partition`

viceroypenguin:


viceroypenguin:
NB: Will need re-basing after #881 due to conflicts.

atifaziz (approved):
Thanks!

PR #898, opened 2022-11-20
Fix `Memoize` Implementation

### LINKED ISSUE #889: `.Memoize()` Concerns
I see two concerns with `.Memoize()`, that I would like to discuss before addressing:

* Currently, the `_error` field is use for both `.GetEnumerator()` errors and iteration errors. This is incorrect under the following scenario: 
  1. Obtain an iterator via `.GetEnumerator()`
  2. Encounter an error during iteration via `.MoveNext()`
  3. Attempt to obtain a new iterator.
  * The expectation is that I would be able to get an iterator, and that it would iterate to the exception
  * The current code shares the `_error` field, and so I would get an error in `.GetEnumerator()` and not be able to iterate to the error.
* Currently, the `MemoizedEnumerable` incorrectly implements the `IDispose` pattern
  * When an object is disposed, it should be prepared to be GC'd. All future attempts to use any aspect of the object should throw an `ObjectDisposedException`
  * The current code only throws an `ObjectDisposedException` when the object holds an open iteration after being disposed. The `.Dispose()` method is treated as a reset rather than an actual disposal; clearing the current cache and starting a new iteration as if `.Skip()` had been called on the original sequence.
  * If clearing the cache is a desired behavior, it should be done under a `.Clear()` method instead of `.Dispose()`. 

@leandromoh I am happy to update the code to correct these issues, but I would like to understand existing thought patterns if either of the above patterns is intended.

CC: @leandromoh

viceroypenguin:
This PR does three things: 1) Correct issues in `Memoize` implementation; 2) Update tests for `Memoize` and `Cartesian` to reflect updated expectations; 3) Update `TestingSequence` with better expectations for how a testing sequence should be used.

Fixes #889

### DIFF
--- a/MoreLinq.Test/TestingSequence.cs
+++ b/MoreLinq.Test/TestingSequence.cs
@@ -69,14 +69,11 @@ namespace MoreLinq.Test
             _disposed = false;
             enumerator.Disposed += delegate
             {
-                Assert.That(_disposed, Is.False, "LINQ operators should not dispose a sequence more than once.");
                 _disposed = true;
             };
-            var ended = false;
             enumerator.MoveNextCalled += (_, moved) =>
             {
-                Assert.That(ended, Is.False, "LINQ operators should not continue iterating a sequence that has terminated.");
-                ended = !moved;
+                Assert.That(_disposed, Is.False, "LINQ operators should not call MoveNext() on a disposed sequence.");
                 MoveNextCallCount++;
             };
             _sequence = null;
--- a/MoreLinq.Test/CartesianTest.cs
+++ b/MoreLinq.Test/CartesianTest.cs
@@ -154,10 +154,8 @@ namespace MoreLinq.Test
         [Test]
         public void TestEmptyCartesianEvaluation()
         {
-            using var sequence = Enumerable.Range(0, 5).AsTestingSequence();
-
-            var resultA = sequence.Cartesian(Enumerable.Empty<int>(), (a, b) => new { A = a, B = b });
-            var resultB = Enumerable.Empty<int>().Cartesian(sequence, (a, b) => new { A = a, B = b });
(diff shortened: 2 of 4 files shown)


MoreLinq.Test/TestingSequence.cs
```diff
@@ -69,14 +69,11 @@ public IEnumerator<T> GetEnumerator()
             _disposed = false;
             enumerator.Disposed += delegate
             {
-                Assert.That(_disposed, Is.False, "LINQ operators should not dispose a sequence more than once.");
                 _disposed = true;
             };
-            var ended = false;
             enumerator.MoveNextCalled += (_, moved) =>
             {
-                Assert.That(ended, Is.False, "LINQ operators should not continue iterating a sequence that has terminated.");
```
viceroypenguin, line 78: Per the [specification](https://learn.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.movenext?view=net-7.0#remarks), it is not a failure to call `.MoveNext()` after receiving a `false` response. The enumerator is simply expected to continue to return `false` for each following call.

MoreLinq.Test/TestingSequence.cs
```diff
@@ -69,14 +69,11 @@ public IEnumerator<T> GetEnumerator()
             _disposed = false;
             enumerator.Disposed += delegate
             {
-                Assert.That(_disposed, Is.False, "LINQ operators should not dispose a sequence more than once.");
```
viceroypenguin, line 72: Per the [specification](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose), it is not a failure to call `.Dispose()` multiple times. The `.Dispose()` method is expected to be idempotent, such that it is callable multiple times without throwing an exception.

atifaziz (commented):
There was no discussion in #889 and conclusion that there's any fixing needed, so I am going to hold off reviewing this for now until I have had time to think about the concerns you've raised.

viceroypenguin:
Upon further review, this is the incorrect path to address concerns anyway.

PR #902, opened 2022-11-25
Simplify `EquiZip` implementation

viceroypenguin:

This PR updates the implementation of `EquiZip` to be done separately from `.ZipImpl()`.

### DIFF
--- a/MoreLinq.Test/EquiZipTest.cs
+++ b/MoreLinq.Test/EquiZipTest.cs
@@ -17,7 +17,10 @@
 
 namespace MoreLinq.Test
 {
+    using System;
+    using System.Collections.Generic;
     using NUnit.Framework;
+    using NUnit.Framework.Interfaces;
     using Tuple = System.ValueTuple;
 
     [TestFixture]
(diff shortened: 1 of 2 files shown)


viceroypenguin:
NB: This may be a v4.0.0 PR?

atifaziz:

viceroypenguin:
Duplicate in the sense that both separate the implementations of the three zip functions, yes. However, #715 uses T4 to generate separate implementations of each method for argument count 2-4. Also, #715 and #902 have very different implementations. IMO #902 is simpler, because the control flow is more linear, rather than 4-way branch with multiple exit points.

Orace:
Some thoughts about this:
 - The used pattern probably wastes resources. It's implemented for 4 sources, and for other cases, it uses dummy sources and discards values with a lambda (resulting to memory allocation that can be avoided): https://github.com/morelinq/MoreLINQ/blob/5ff2ad4ea88cc2052eded807d87a421b9a301ba6/MoreLinq/EquiZip.cs#L71
   It will not scale well if we want to add overloads with more sources.
   
 - The control flow of #715 can be simplified a little bit to match #902 that offers more complete exception messages.

viceroypenguin:
@Orace Here's a copy of the [code gen](https://github.com/viceroypenguin/SuperLinq/blob/master/Generators/SuperLinq.Generator/EquiZip.sbntxt#L66-L96) for a simpler control flow if you want to steal it.

Closing this PR in favor of an updated #715.

PR #904, opened 2022-11-26
Simplify `ZipShortest` implementation

viceroypenguin:

This PR updates the implementation of `ZipShortest` to be done separately from `.ZipImpl()`.


atifaziz:

viceroypenguin:
Duplicate in the sense that both separate the implementations of the three zip functions, yes. However, #715 uses T4 to generate separate implementations of each method for argument count 2-4. #715 and #904 have essentially the same implementation, with some variances on how the implementation is written.

viceroypenguin:
Closing in favor of PR #715.

PR #905, opened 2022-11-26
Simplify `ZipLongest` implementation

viceroypenguin:

This PR updates the implementation of `ZipLongest` to be done separately from `.ZipImpl()`.

### DIFF
--- a/MoreLinq/ZipLongest.cs
+++ b/MoreLinq/ZipLongest.cs
@@ -19,6 +19,7 @@ namespace MoreLinq
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
 
     static partial class MoreEnumerable
     {
@@ -56,13 +57,13 @@ namespace MoreLinq
         public static IEnumerable<TResult> ZipLongest<TFirst, TSecond, TResult>(
             this IEnumerable<TFirst> first,
             IEnumerable<TSecond> second,
-            Func<TFirst, TSecond, TResult> resultSelector)
+            Func<TFirst?, TSecond?, TResult> resultSelector)
         {
             if (first == null) throw new ArgumentNullException(nameof(first));
             if (second == null) throw new ArgumentNullException(nameof(second));
             if (resultSelector == null) throw new ArgumentNullException(nameof(resultSelector));
 
-            return ZipImpl<TFirst, TSecond, object, object, TResult>(first, second, null, null, (a, b, _, _) => resultSelector(a, b), 1);
+            return ZipLongestImpl(first, second, Enumerable.Repeat(default(object?), int.MaxValue), Enumerable.Repeat(default(object?), int.MaxValue), (a, b, _, _) => resultSelector(a, b), 2);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


viceroypenguin:
NB: Depends on #901.

atifaziz:

viceroypenguin:
Duplicate in the sense that both separate the implementations of the three zip functions, yes. However, #715 uses T4 to generate separate implementations of each method for argument count 2-4. Also, #715 and #905 have very different implementations (subject to debate which is better).

* #715 keeps the `TryRead` method to play nice-nice with non-conforming enumerables and hide complexity
* #905 is ultimately simpler, but relies on enumerables to allow repeated `.MoveNext()` after enumerable is complete.

Orace:
Some thoughts about this:
 - The used pattern probably wastes resources. It's implemented for 4 sources, and for other cases, it uses dummy sources and discards values with a lambda: https://github.com/morelinq/MoreLINQ/blob/3d360392532db6cd2dcccdd803eb3d9bf98a142a/MoreLinq/ZipLongest.cs#L66
   It will not scale well if we want to add overloads with more sources.
 - Calls to `.MoveNext()` after an enumerable is complete should be avoided. Moreover this is tested with `TestingSequence` https://github.com/morelinq/MoreLINQ/blob/9c17aae35a5d8194776bc348028e78b36c023d6b/MoreLinq.Test/TestingSequence.cs#L93-L98

viceroypenguin:
> Some thoughts about this:
> 
> * The used pattern probably wastes resources. It's implemented for 4 sources, and for other cases, it uses dummy sources and discards values with a lambda (resulting to memory allocation that can be avoided): https://github.com/morelinq/MoreLINQ/blob/3d360392532db6cd2dcccdd803eb3d9bf98a142a/MoreLinq/ZipLongest.cs#L66
I did do some testing, and while the memory usage is negligible (400b vs 176b), it does cost an additional 30-50% to do it the #905 way. Probably has mostly to do with the extra `MoveNext` calls being done on the dummy enumerables. 🤔

>   It will not scale well if we want to add overloads with more sources.
Not sure that this is a concern. Asking for more than 4 enumerables to be zipped together seems a little excessive, and should probably be solved by the person requesting it.

> * Calls to `.MoveNext()` after an enumerable is complete should be avoided. Moreover this is tested with `TestingSequence` https://github.com/morelinq/MoreLINQ/blob/9c17aae35a5d8194776bc348028e78b36c023d6b/MoreLinq.Test/TestingSequence.cs#L93-L98
Already discussed and addressed in #901/#936. I won't repeat the discussion further here.

viceroypenguin:
Hey @Orace, I took your idea and managed to improve performance by [another 10-20%](https://github.com/viceroypenguin/SuperLinq/pull/162#issuecomment-1406006822). Feel free to take a look at the updated [`TryRead()`](https://github.com/viceroypenguin/SuperLinq/blob/master/Generators/SuperLinq.Generator/ZipLongest.sbntxt#L15-L23) and [code gen](https://github.com/viceroypenguin/SuperLinq/blob/master/Generators/SuperLinq.Generator/ZipLongest.sbntxt#L75-L87). It's Scriban instead of T4, but should be easy to translate, I hope?

Closing this PR in favor of an updated #715.

PR #907, opened 2022-12-09
Use random shared

leotsarev:
Advantages:
- Do not allocate Random & GlobalRandom instances on .NET 6
- Common Random implementation in Random() and other places


viceroypenguin (approved):

atifaziz (approved):
Thanks for this!

I've made some minor changes:

- Reduced the scope of conditional parts within `GlobalRandom`.
- Fixed the doc comment (anchor, grammar and formatting).
- Conditionally imported `System.Threading` on .NET 6+.
- Undid the move of `GlobalRandom` into a separate file since it's an unrelated change that can go into a follow-up PR (if absolutely necessary) and at which point the type could be un-nested too.

The net effect is a smaller diff.

PR #921, opened 2023-01-15
Retire our delegating `Comparer`

atifaziz:
This PR removes `Comparer` in favour of `Delegate.Comparer` from [Delegating](https://www.nuget.org/packages/Delegating) that provides the same functionality.


viceroypenguin (changes requested):
Is there a reason not to use `System.Collections.Generic.Comparer<T>.Create()`? (available since net4.5 per https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.comparer-1.create?view=net-7.0)

atifaziz:
> Is there a reason not to use `System.Collections.Generic.Comparer<T>.Create()`?

None. It's what happens when you work late. 😉 Thanks for the review! Fixed with c3dfb1996d46e7db773861d8da3d56075158c41c.

viceroypenguin (approved):

PR #924, opened 2023-01-15
Enable static analysis

atifaziz:
This PR enables static code analysis through the solution and addresses the issues/warnings/errors (see individual commits) highlighted by the .NET analyzers.

### DIFF
--- a/MoreLinq.Test/MoreLinq.Test.csproj
+++ b/MoreLinq.Test/MoreLinq.Test.csproj
@@ -62,6 +62,7 @@
       <Compile Include="..\MoreLinq\Debug.cs" Link="Debug.cs" />
       <Compile Include="..\MoreLinq\Disposable.cs" Link="Disposable.cs" />
       <Compile Include="..\MoreLinq\Reactive\Subject.cs" Link="Subject.cs" />
+      <Compile Include="Extensions.cs" />
       <Compile Include="Throws.cs" />
       <Compile Include="BreakingCollection.cs" />
       <Compile Include="BreakingAction.cs" />
--- a/MoreLinq/AssertCount.cs
+++ b/MoreLinq/AssertCount.cs
@@ -76,7 +76,7 @@ namespace MoreLinq
             var message = cmp < 0
                         ? "Sequence contains too few elements when exactly {0} were expected."
                         : "Sequence contains too many elements when exactly {0} were expected.";
-            return new SequenceException(string.Format(message, count.ToString("N0")));
+            return new SequenceException(string.Format(null, message, count.ToString("N0", null)));
         }
 
         #endif
--- a/MoreLinq/Experimental/Batch.cs
+++ b/MoreLinq/Experimental/Batch.cs
@@ -269,7 +269,6 @@ namespace MoreLinq.Experimental
             public override T this[int index]
             {
                 get => index >= 0 && index < Count ? _array[index] : throw new ArgumentOutOfRangeException(nameof(index));
-#pragma warning restore CA2201 // Do not raise reserved exception types
                 set => throw new NotSupportedException();
             }
 
--- a/MoreLinq/Experimental/Batch.cs
+++ b/MoreLinq/Experimental/Batch.cs
@@ -268,9 +268,10 @@ namespace MoreLinq.Experimental
 
             public override T this[int index]
             {
+#pragma warning disable CA2201 // Do not raise reserved exception types
                 get => index >= 0 && index < Count ? _array[index] : throw new IndexOutOfRangeException();
+#pragma warning restore CA2201 // Do not raise reserved exception types
                 set => throw new NotSupportedException();
-
             }
 
             public void Dispose()
--- a/MoreLinq/Fold.cs
+++ b/MoreLinq/Fold.cs
@@ -22,8 +22,6 @@ namespace MoreLinq
 
     static partial class MoreEnumerable
     {
-        const string folder = nameof(folder);
-
         static TResult FoldImpl<T, TResult>(IEnumerable<T> source, int count,
             Func<T, TResult>? folder1 = null,
             Func<T, T, TResult>? folder2 = null,
--- a/MoreLinq/Fold.cs
+++ b/MoreLinq/Fold.cs
@@ -22,6 +22,8 @@ namespace MoreLinq
 
     static partial class MoreEnumerable
     {
+        const string folder = nameof(folder);
+
         static TResult FoldImpl<T, TResult>(IEnumerable<T> source, int count,
             Func<T, TResult>? folder1 = null,
             Func<T, T, TResult>? folder2 = null,
@@ -60,7 +62,9 @@ namespace MoreLinq
                 || count == 16 && folder16 == null
                 )
             {                                                // ReSharper disable NotResolvedInText
+#pragma warning disable CA2208 // Instantiate argument exceptions correctly
                 throw new ArgumentNullException("folder");   // ReSharper restore NotResolvedInText
+#pragma warning restore CA2208 // Instantiate argument exceptions correctly
             }
 
             var elements = new T[count];
--- a/MoreLinq/Fold.cs
+++ b/MoreLinq/Fold.cs
@@ -100,7 +100,7 @@ namespace MoreLinq
             var message = cmp < 0
                         ? "Sequence contains too few elements when exactly {0} {1} expected."
                         : "Sequence contains too many elements when exactly {0} {1} expected.";
-            return new InvalidOperationException(string.Format(message, count.ToString("N0"), count == 1 ? "was" : "were"));
+            return new InvalidOperationException(string.Format(null, message, count.ToString("N0", null), count == 1 ? "was" : "were"));
         }
     }
 }
--- a/MoreLinq/From.cs
+++ b/MoreLinq/From.cs
@@ -38,7 +38,9 @@ namespace MoreLinq
         {
             return _(); IEnumerable<T> _()
             {
+#pragma warning disable CA1062 // Validate arguments of public methods
                 yield return function();
+#pragma warning restore CA1062 // Validate arguments of public methods
             }
         }
 
@@ -59,8 +61,10 @@ namespace MoreLinq
         {
             return _(); IEnumerable<T> _()
             {
+#pragma warning disable CA1062 // Validate arguments of public methods
                 yield return function1();
                 yield return function2();
+#pragma warning restore CA1062 // Validate arguments of public methods
             }
         }
 
@@ -82,9 +86,11 @@ namespace MoreLinq
         {
             return _(); IEnumerable<T> _()
             {
+#pragma warning disable CA1062 // Validate arguments of public methods
                 yield return function1();
                 yield return function2();
                 yield return function3();
+#pragma warning restore CA1062 // Validate arguments of public methods
             }
         }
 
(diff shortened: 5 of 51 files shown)


MoreLinq.Test/MoreLinq.Test.csproj
```diff
@@ -62,6 +62,7 @@
       <Compile Include="..\MoreLinq\Debug.cs" Link="Debug.cs" />
       <Compile Include="..\MoreLinq\Disposable.cs" Link="Disposable.cs" />
       <Compile Include="..\MoreLinq\Reactive\Subject.cs" Link="Subject.cs" />
+      <Compile Include="Extensions.cs" />
```
viceroypenguin, line 65: Why have `EnableDefaultCompileItems` turned off and then manually insert items? If there are items to exclude from compilation, better to do `<Compile Remove="xxx.cs"/>`. (Probably a separate PR to fix, but comment here to put on your todo list.)

MoreLinq/AssertCount.cs
```diff
@@ -76,7 +76,7 @@ static Exception OnAssertCountFailure(int cmp, int count)
             var message = cmp < 0
                         ? "Sequence contains too few elements when exactly {0} were expected."
                         : "Sequence contains too many elements when exactly {0} were expected.";
```
viceroypenguin, line 78:
```suggestion
            var message = cmp < 0
                        ? "Sequence contains too few elements when exactly {0:N0} were expected."
                        : "Sequence contains too many elements when exactly {0:N0} were expected.";
            return new SequenceException(string.Format(null, message, count));
```

MoreLinq/Experimental/Batch.cs
```diff
@@ -268,9 +268,10 @@ public bool UpdateWithNext()
 
             public override T this[int index]
             {
+#pragma warning disable CA2201 // Do not raise reserved exception types
                 get => index >= 0 && index < Count ? _array[index] : throw new IndexOutOfRangeException();
```
viceroypenguin, line 272: Per https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.item?view=net-7.0, should be `ArgumentOutOfRangeException` anyway. Possibly breaking change, but highly unlikely anyone is specifically catching for `IndexOutOfRangeException`.

MoreLinq/Fold.cs
```diff
@@ -96,7 +100,7 @@ static Exception OnFolderSourceSizeError(int cmp, int count)
             var message = cmp < 0
                         ? "Sequence contains too few elements when exactly {0} {1} expected."
                         : "Sequence contains too many elements when exactly {0} {1} expected.";
```
viceroypenguin, line 102: See comment on `AssertCount.cs`

MoreLinq/From.cs
```diff
@@ -38,7 +38,9 @@ public static IEnumerable<T> From<T>(Func<T> function)
         {
             return _(); IEnumerable<T> _()
             {
+#pragma warning disable CA1062 // Validate arguments of public methods
                 yield return function();
```
viceroypenguin, line 42: Is there a reason not to check that `function` is not null?

MoreLinq/PendNode.cs
```diff
@@ -91,7 +91,9 @@ public IEnumerator<T> GetEnumerator()
                                 case 1: concat2 = item.Value; break;
                                 case 2: concat3 = item.Value; break;
                                 case 3: concat4 = item.Value; break;
+#pragma warning disable CA2201 // Do not raise reserved exception types
                                 default: throw new IndexOutOfRangeException();
```
viceroypenguin, line 95: `ArgumentOutOfRangeException`

MoreLinq/ToArrayByIndex.cs
```diff
@@ -126,16 +126,22 @@ static partial class MoreEnumerable
...
-                var i = indexSelector(e);
-                if (i < 0)
-                    throw new IndexOutOfRangeException();
-                lastIndex = Math.Max(i, lastIndex);
-                Indexed().Add(new KeyValuePair<int, T>(i, e));
+                switch (indexSelector(e))
+                {
+                    case < 0:
+#pragma warning disable CA2201 // Do not raise reserved exception types
+                        throw new IndexOutOfRangeException();
```
viceroypenguin, line 133: `ArgumentOutOfRangeException`

viceroypenguin (approved):

MoreLinq/AssertCount.cs
atifaziz: This could potentially cause an additional allocation (potentially because it could vary with runtime): first to box `count` and then to format it into a string.

viceroypenguin: int.ToString() has a minimum allocation of 32b, so boxing is actually cheaper. That said, if you're worried about allocation here, then interpolated string is actually best, because the builder is on the stack and types are used without boxing.

MoreLinq/ToArrayByIndex.cs
atifaziz: This is by-design. It's not an invalid argument of `ToArrayByIndex`, but an invalid index for the target array.

MoreLinq/ToArrayByIndex.cs
```diff
@@ -242,8 +248,14 @@ static partial class MoreEnumerable
             foreach (var e in source)
             {
                 var i = indexSelector(e);
+
                 if (i < 0 || i > array.Length)
+                {
+#pragma warning disable CA2201 // Do not raise reserved exception types
                     throw new IndexOutOfRangeException();
```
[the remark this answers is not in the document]
atifaziz: This is by-design. It's not an invalid argument of `ToArrayByIndex`, but an invalid index for the target array.

MoreLinq/Experimental/Batch.cs
atifaziz: Too bad that this didn't get caught during the review of PR #856, but it's not too late to change since this overload of `Batch` hasn't been released yet.

---
PS The summary of [`IndexOutOfRangeException`](https://learn.microsoft.com/en-us/dotnet/api/system.indexoutofrangeexception) really needs to be changed to avoid mentioning _collection_:

> The exception that is thrown when an attempt is made to access an element of an array or collection with an index that is outside its bounds.

The examples in the documentation don't help much with mixed uses:

> Typically, an [IndexOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.indexoutofrangeexception?view=net-7.0) exception is thrown as a result of developer error. Instead of handling the exception, you should diagnose the cause of the error and correct your code. The most common causes of the error are:

>
> - Forgetting that the upper bound of a collection or a zero-based array is one less than its number of members or elements, as the following example illustrates.

>
>   ```c#
>   using System;
>   using System.Collections.Generic;
>   
>   public class Example
>   {
>      public static void Main()
>      {
>         List<Char> characters = new List<Char>();
>         characters.InsertRange(0, new Char[] { 'a', 'b', 'c', 'd', 'e', 'f' } );
>         for (int ctr = 0; ctr <= characters.Count; ctr++)
>            Console.Write("'{0}'    ", characters[ctr]);
>      }
>   }
>   // The example displays the following output:
>   //    'a'    'b'    'c'    'd'    'e'    'f'
>   //    Unhandled Exception:
>   //    System.ArgumentOutOfRangeException:
>   //    Index was out of range. Must be non-negative and less than the size of the collection.
>   //    Parameter name: index
>   //       at Example.Main()
>   ```

MoreLinq/From.cs
atifaziz: See the discussion in PR #333 for background on this.

MoreLinq/Experimental/Batch.cs
atifaziz: Follow-up in PR #927.

MoreLinq/Fold.cs
atifaziz: [Same answer](https://github.com/morelinq/MoreLINQ/pull/924#discussion_r1070661519) as for [comment on `AssertCount.cs`](https://github.com/morelinq/MoreLINQ/pull/924#discussion_r1070624334).

MoreLinq/ToArrayByIndex.cs
viceroypenguin: In that case, I would recommend intentionally not making the check and allowing the array to throw the exception on the invalid index. The array indexer will repeat the check anyway, and will provide a better exception message than we can provide. Mark a comment that guard-checking is intentionally deferred to the array index instead?

MoreLinq/AssertCount.cs
atifaziz: > then interpolated string is actually best, because the builder is on the stack and types are used without boxing.

Yes, but this applies to more recent versions and I think as recent as only .NET 7.

> if you're worried about allocation here

I'm not worried about allocation. Your suggestion didn't provide any context or motivation so I didn't understand what's being improve with respect to the subject of this PR, but it was making one allocation more and duplicating the formatting specification. Since it's unrelated, can we take the suggestion and discussion in a separate PR?

MoreLinq/PendNode.cs
atifaziz: This case cannot occur due to a caller-supplied argument so `ArgumentOutOfRangeException` doesn't seem to be the right call here. In fact, arriving here would constitute a bug in the implementation. Ideally, we could have used [`UnreachableException`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.unreachableexception), but that just got added in .NET 7.

MoreLinq/AssertCount.cs
viceroypenguin: * With regards to my original comment: I commented because having a `string.Format()` _and_ an `int.ToString()` looks bad. I would prefer moving the `N0` into the format specifier because it's simpler to state that "we're formatting the string to introduce this value and here's how the value should be formatted". Also, it reduces the number of places to specify `null` `IFormatProvider`.
* You replied that it would cause an additional allocation. My response vis-a-vis allocation was strictly in response to yours. 
* In this case, in lower versions of the framework, the compiler would turn it into simple `string.Join()` most likely, so my arguments re: allocation are probably moot for lower fxs anyway. That said, under net6+, interpolated would be lowest allocation.
* If you want to leave alone and move to a separate PR, I'm good with that.

MoreLinq/PendNode.cs
viceroypenguin: `NotSupportedException` with a message of "Bug in MoreLinq implementation." (needs better wording, but you get the idea) would probably be a better expression here then. `IndexOutOfRange` looks out of place and doesn't explain the intent of this `throw`. 

I'm good with it being done on another PR though.

MoreLinq/ToArrayByIndex.cs
atifaziz: The array assignment happens much later and there's not much point in wasting time and resources until then.

MoreLinq/PendNode.cs
atifaziz: I've added a private version of `UnreachableException` with ac34b3055ea18b8e83435de2238716fa1521b9bf and thrown that here instead.

viceroypenguin (approved):
With formatting changes addressed in #929, approved.

MoreLinq/AssertCount.cs
atifaziz: > …in lower versions of the framework, the compiler would turn it into simple `string.Join()` most likely, …

You mean `String.Format`, right?

> * That said, under net6+, interpolated would be lowest allocation.

When I tested yesterday, I didn't get the low-allocating string interpolation under .NET 6, only .NET 7. Perhaps I made some mistake in my quick test in LINQPad, but on retrying, it works for .NET 6.

Anyway, I addressed this with string interpolation and removed all duplication in PR #929.

PR #928, opened 2023-01-16
Additional tests for `TagFirstLast`

Orace:
Add 

- TagFirstLastEvaluatesSourceLazily ensure that TagFirstLast doesn't evaluate the whole source uselessly.
- TagFirstLastWithSourceSequenceOfZero for 100% code coverage in #643

### DIFF
--- a/MoreLinq.Test/TagFirstLastTest.cs
+++ b/MoreLinq.Test/TagFirstLastTest.cs
@@ -23,6 +23,15 @@ namespace MoreLinq.Test
     public class TagFirstLastTest
     {
         [Test]
+        public void TagFirstLastEvaluatesSourceLazily()
+        {
+            var source = MoreEnumerable.From(() => 123, () => 456, () => throw new TestException());
+            source.TagFirstLast((item, isFirst, isLast) => new { Item = item, IsFirst = isFirst, IsLast = isLast })
+                  .Take(1)
+                  .Consume();
+        }
+
+        [Test]
         public void TagFirstLastIsLazy()
         {
             new BreakingSequence<object>().TagFirstLast(BreakingFunc.Of<object, bool, bool, object>());
--- a/MoreLinq.Test/TagFirstLastTest.cs
+++ b/MoreLinq.Test/TagFirstLastTest.cs
@@ -38,6 +38,14 @@ namespace MoreLinq.Test
         }
 
         [Test]
+        public void TagFirstLastWithSourceSequenceOfZero()
+        {
+            var source = new int[0];
+            var sut = source.TagFirstLast((item, isFirst, isLast) => new { Item = item, IsFirst = isFirst, IsLast = isLast });
+            Assert.That(sut, Is.Empty);
+        }
+
+        [Test]
         public void TagFirstLastWithSourceSequenceOfOne()
         {
             var source = new[] { 123 };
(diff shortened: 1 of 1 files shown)


MoreLinq.Test/TagFirstLastTest.cs
```diff
@@ -22,12 +22,29 @@ namespace MoreLinq.Test
...
         [Test]
         public void TagFirstLastIsLazy()
         {
             new BreakingSequence<object>().TagFirstLast(BreakingFunc.Of<object, bool, bool, object>());
         }
 
+        [Test]
+        public void TagFirstLastWithSourceSequenceOfZero()
+        {
+            var source = new int[0];
```
viceroypenguin, line 43:
```suggestion
            var source = EmptyArray<int>.Value;
```

viceroypenguin (changes requested):

MoreLinq.Test/TagFirstLastTest.cs
Orace: Sadly this doesn't compile when .NET Framework 4.5.1 is targeted.

viceroypenguin (approved):

MoreLinq.Test/TagFirstLastTest.cs
Orace: My bad, I didn't fully read your first comment.

MoreLinq.Test/TagFirstLastTest.cs
```diff
@@ -22,12 +22,29 @@ namespace MoreLinq.Test
...
         public void TagFirstLastIsLazy()
         {
             new BreakingSequence<object>().TagFirstLast(BreakingFunc.Of<object, bool, bool, object>());
         }
 
+        [Test]
+        public void TagFirstLastWithSourceSequenceOfZero()
+        {
+            var source = Enumerable.Empty<int>();
+            var sut = source.TagFirstLast((item, isFirst, isLast) => new { Item = item, IsFirst = isFirst, IsLast = isLast });
```
atifaziz, line 44: The lambda will never be called so we could make it more strict and clear like so:

```suggestion
            var sut = source.TagFirstLast(BreakingFunc.Of<int, bool, bool, int>());
```

MoreLinq.Test/TagFirstLastTest.cs
```diff
@@ -22,12 +22,29 @@ namespace MoreLinq.Test
     [TestFixture]
     public class TagFirstLastTest
     {
+        [Test]
+        public void TagFirstLastEvaluatesSourceLazily()
+        {
+            var source = MoreEnumerable.From(() => 123, () => 456, () => throw new TestException());
```
atifaziz, line 28: Use `BreakingFunc` instead of throwing `TestException`:

```suggestion
            var source = MoreEnumerable.From(() => 123, () => 456, BreakingFunc.Of<int>());
```

MoreLinq.Test/TagFirstLastTest.cs
```diff
@@ -22,12 +22,29 @@ namespace MoreLinq.Test
     [TestFixture]
     public class TagFirstLastTest
     {
+        [Test]
+        public void TagFirstLastEvaluatesSourceLazily()
```
atifaziz, line 26: Rename the test to be clear about why `source` has at least two elements.

```suggestion
        public void TagFirstLastDoesOneLookAhead()
```

atifaziz (changes requested):
Thanks for adding these tests. Just a couple of suggestions and then we're good to merge this!

atifaziz (approved):
Thanks for adding these additional tests.

PR #942, opened 2023-01-23
Improve list optimisation in `CountDown`

atifaziz:
`CountDown` has a path optimised for when the `source` parameter is a list, but it was poorly written respect to performance. It accessed the list `Count` property 3 times, twice during each iteration of the loop. Each access is a virtual dispatch so it seems to adds up, making it perform _worse_ than when the source is a sequence (zero net optimisation). This PR fixes that by accessing `Count` once, storing it in a local and then using the local throughout. This improves the performance by 40%.

## Before this PR

|            Method |               Source |     Mean |    Error |   StdDev |   Median | Allocated |
|------------------ |--------------------- |---------:|---------:|---------:|---------:|----------:|
| MoreLinqCountDown |       Int32[1000000] | 21.31 ms | 0.426 ms | 1.209 ms | 21.02 ms |     119 B |
| MoreLinqCountDown | Syste(...)rator [36] | 19.30 ms | 0.496 ms | 1.440 ms | 18.54 ms |     179 B |

See [`benchmarks-before.log.txt`](https://github.com/morelinq/MoreLINQ/files/10483667/benchmarks-before.log.txt) for full details.

## After this PR

|            Method |               Source |     Mean |    Error |   StdDev |   Median | Allocated |
|------------------ |--------------------- |---------:|---------:|---------:|---------:|----------:|
| MoreLinqCountDown |       Int32[1000000] | 12.83 ms | 0.318 ms | 0.934 ms | 12.63 ms |     100 B |
| MoreLinqCountDown | Syste(...)rator [36] | 20.65 ms | 0.631 ms | 1.851 ms | 20.01 ms |     179 B |

See [`benchmarks-after.log.txt`](https://github.com/morelinq/MoreLINQ/files/10483669/benchmarks-after.log.txt) for full details.

## Additional Information

    BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22621.1105)
    Intel Core i7-1065G7 CPU 1.30GHz, 1 CPU, 8 logical and 4 physical cores
    .NET SDK=7.0.102
      [Host] : .NET 7.0.2 (7.0.222.60605), X86 RyuJIT AVX2

See also:

- [`Benchmark.linq`](http://share.linqpad.net/d63764.linq) for benchmarking code for [LINQPad](https://www.linqpad.net/).
- PR #472 for some related discussion.

### DIFF
--- a/MoreLinq/CountDown.cs
+++ b/MoreLinq/CountDown.cs
@@ -65,11 +65,11 @@ namespace MoreLinq
 
             IEnumerable<TResult> IterateList(IListLike<T> list)
             {
-                var count = list.Count;
-                var countdown = Math.Min(count, count);
+                var listCount = list.Count;
+                var countdown = Math.Min(count, listCount);
 
-                for (var i = 0; i < count; i++)
-                    yield return resultSelector(list[i], count - i <= count ? --countdown : null);
+                for (var i = 0; i < listCount; i++)
+                    yield return resultSelector(list[i], listCount - i <= count ? --countdown : null);
             }
 
             IEnumerable<TResult> IterateCollection(int i)
--- a/MoreLinq/CountDown.cs
+++ b/MoreLinq/CountDown.cs
@@ -65,15 +65,11 @@ namespace MoreLinq
 
             IEnumerable<TResult> IterateList(IListLike<T> list)
             {
-                var countdown = Math.Min(count, list.Count);
+                var count = list.Count;
+                var countdown = Math.Min(count, count);
 
-                for (var i = 0; i < list.Count; i++)
-                {
-                    var cd = list.Count - i <= count
-                           ? --countdown
-                           : (int?) null;
-                    yield return resultSelector(list[i], cd);
-                }
+                for (var i = 0; i < count; i++)
+                    yield return resultSelector(list[i], count - i <= count ? --countdown : null);
             }
 
             IEnumerable<TResult> IterateCollection(int i)


MoreLinq/CountDown.cs
```diff
@@ -65,15 +65,11 @@ static partial class MoreEnumerable
 
             IEnumerable<TResult> IterateList(IListLike<T> list)
             {
-                var countdown = Math.Min(count, list.Count);
+                var count = list.Count;
```
viceroypenguin, line 68: This variable hides `CountDown` parameter `count`.

MoreLinq/CountDown.cs
```diff
@@ -65,15 +65,11 @@ static partial class MoreEnumerable
...
+                var countdown = Math.Min(count, count);
```
viceroypenguin, line 69: This is effectively a no-op, since both parameters are the same.

viceroypenguin (changes requested):

viceroypenguin:
Two things:
1. I would rerun the performance benchmark after making the changes to fix the bugs above. The missed `count` value could have a performance impact on the behavior.
2. If possible, I would copy the before/after into LinqPAD for benchmarking. The reason is because two separate benchmarks, even on the same machine, are not always comparable (other processes, etc.). It is more accurate to have a `Before()` method and an `After()` method in LinqPAD and benchmark them against each other in the same run.

MoreLinq/CountDown.cs
atifaziz: Yep, looks like I made a silly mistake and the [tests broke](https://ci.appveyor.com/project/raboof/morelinq/builds/46021323/job/c7ad76k2b4qypcv0#L448) in the CI build. On the plus side, good things tests caught it. 😅

MoreLinq/CountDown.cs
atifaziz: Yep, working late again. 😆 Fixed with da13c3f0c25786c033807f120cbbf9c1c01d5176.

viceroypenguin (approved):

atifaziz:
> 1. I would rerun the performance benchmark after making the changes to fix the bugs above. The missed `count` value could have a performance impact on the behavior.

45% improvement:

|            Method |               Source |     Mean |    Error |   StdDev |   Median | Allocated |
|------------------ |--------------------- |---------:|---------:|---------:|---------:|----------:|
| MoreLinqCountDown |       Int32[1000000] | 11.27 ms | 0.316 ms | 0.923 ms | 11.02 ms |     100 B |
| MoreLinqCountDown | Syste(...)rator [36] | 20.61 ms | 0.707 ms | 2.041 ms | 20.58 ms |     179 B |


> 2. If possible, I would copy the before/after into LinqPAD for benchmarking. The reason is because two separate benchmarks, even on the same machine, are not always comparable (other processes, etc.). It is more accurate to have a `Before()` method and an `After()` method in LinqPAD and benchmark them against each other in the same run.

Yes, I am aware and admit taking a shortcut there. The difference was large enough and comparable in the many runs I did. What's more important is the comparison between list and sequence processing in the same ran than the before and after.

PR #945, opened 2023-01-25
Hide extensions that conflict with newer .NET versions

### LINKED ISSUE #565: Ease usage in next major version by renaming methods conflicting with Enumerable methods

viceroypenguin:
Properly implement solution for hiding extension methods on different frameworks by removing the `this` modifier. 

* Supports binary compatibility by still compiling duplicated methods. Since extension methods are implemented via syntactic sugar, as long as the methods still exist, the `this` modifier can be removed and callers will still be able to reach the method.
* Does not address issues with `MinBy`/`MaxBy` per other discussions.

---
* Replaces: #739
* Replaces: #749
* Fixes: #565
* Fixes: #738


viceroypenguin:
PS: @atifaziz  You'll have to decide how you want to handle `System.Linq` being referenced in unit tests, since they are required for multi-fx support for some test implementations.

atifaziz:
> You'll have to decide how you want to handle `System.Linq` being referenced in unit tests, since they are required for multi-fx support for some test implementations.

See 5e1d88ae089e9a5a66c8ee6d94ecabb14313e20f.

atifaziz:
Dropping a note that unfortunately .NET and MoreLINQ signatures differ for `ExceptBy`:

MoreLINQ:

- [`ExceptBy<TSource, TKey>(IEnumerable<TSource>, IEnumerable<TSource>, Func<TSource, TKey>)`](https://morelinq.github.io/3.4/ref/api/html/M_MoreLinq_MoreEnumerable_ExceptBy__2.htm)
- [`ExceptBy<TSource, TKey>(IEnumerable<TSource>, IEnumerable<TSource>, Func<TSource, TKey>, IEqualityComparer<TKey>)`](https://morelinq.github.io/3.4/ref/api/html/M_MoreLinq_MoreEnumerable_ExceptBy__2_1.htm)

.NET:

- [`ExceptBy<TSource,TKey>(IEnumerable<TSource>, IEnumerable<TKey>, Func<TSource,TKey>)`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.exceptby?view=net-8.0#system-linq-enumerable-exceptby-2(system-collections-generic-ienumerable((-0))-system-collections-generic-ienumerable((-1))-system-func((-0-1))))
- [`ExceptBy<TSource,TKey>(IEnumerable<TSource>, IEnumerable<TKey>, Func<TSource,TKey>, IEqualityComparer<TKey>)`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.exceptby?view=net-8.0#system-linq-enumerable-exceptby-2(system-collections-generic-ienumerable((-0))-system-collections-generic-ienumerable((-1))-system-func((-0-1))-system-collections-generic-iequalitycomparer((-1))))

Most notable, the second parameter type differs: `IEnumerable<TSource>` vs. `IEnumerable<TKey>`

atifaziz:
> Ah, then we don't need to hide `ExceptBy`

Reverted with 7d341224385450784498acc5e3bc255210fe82a4.

MoreLinq/Append.cs
viceroypenguin, line 1: Removing `Append` and `Prepend` entirely will break binary compatibility for these methods, and affect any consumers who are accessing them via `AppendExtension.Append()`. Admittedly, these are unlikely given how long `Append` and `Prepend` have existed in the framework, but the possibility remains.

atifaziz: I've been thinking about that and I think we should take the breaking change now with a major version update and add it back if people report binary compatibility problems that they can't address because it's in some third-party component. The worst that can happen is that people running into this specific issue can't upgrade to 4.0 immediately. Fair?

atifaziz: I've reverted this for now with cf55babce69e6cc601ba6696d80da8c3c5e54a96 as it belongs in a separate PR as this is just about hiding some extensions depending on the target.

PR #946, opened 2023-01-26
Fix collection-optimized paths to be at iteration-time

### LINKED ISSUE #943: Some operators incorrectly cache count of a collection source
In some places (see `CountDown+IterateCollection`), `TryGetCollectionCount` is used incorrectly. The `Count` of the collection is evaluated at the time the enumerable is created, rather than when it is enumerated. However, the length of the collection could change in between the time the enumeration is created and the time it is enumerated. As such, the value of collection `Count` should be obtained at the time of enumeration. `TryGetCollectionCount` can be used to determine that a `source` is a collection for the purposes of determining which method to call; but if done, the value should not be relied on.

List of potentially affected operators (added by @atifaziz):

- [x] `AssertCount`
- [x] `Batch`
- [x] `CountDown`
- [x] `FallbackIfEmpty`
- [x] `PadStart`
- [x] `SkipLast`
- [x] `TakeLast`

atifaziz:
This PR fixes #943.

It fixes the optimised path in the following operators for when the source is a collection:

- `AssertCount`
- `CountDown`
- `FallbackIfEmpty`
- `PadStart`
- `SkipLast`
- `TakeLast`

All of the above operators have deferred execution semantics, but they were binding to the source collection count when invoked and not when the resulting sequence was iterated. Consequently, if the source collection was updated before iteration, the (cached) count and the resulting sequence would be incorrect.

Tests have been added to avoid regressions in the future.

Some tests were removed because the optimisations they were exercising couldn't be maintained anymore. For example, if `AssertCount` could determine that the source was a collection and had the right count, then it would return the source reference as-is. Now, the count is asserted when iteration start.

### DIFF
--- a/MoreLinq.Test/AssertCountTest.cs
+++ b/MoreLinq.Test/AssertCountTest.cs
@@ -18,6 +18,7 @@
 namespace MoreLinq.Test
 {
     using System;
+    using System.Collections.Generic;
     using NUnit.Framework;
 
     [TestFixture]
@@ -113,13 +114,6 @@ namespace MoreLinq.Test
             _ = new BreakingCollection<int>(new int[5]).AssertCount(0);
         }
 
-        [Test]
-        public void AssertCountWithMatchingCollectionCount()
-        {
-            var xs = new[] { 123, 456, 789 };
-            Assert.That(xs, Is.SameAs(xs.AssertCount(3)));
-        }
-
         [TestCase(3, 2, "Sequence contains too many elements when exactly 2 were expected.")]
         [TestCase(3, 4, "Sequence contains too few elements when exactly 4 were expected.")]
         public void AssertCountWithMismatchingCollectionCount(int sourceCount, int count, string message)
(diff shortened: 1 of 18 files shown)


viceroypenguin (approved):
The only concern I'd have is that in some cases, there is a performance advantage to have in .net6+ by using `IEnumerable.TryGetNonEnumeratedCount()`. 

The differences are:
* `TryGetNonEnumeratedCount` tests `ICollection<>`, an internal `IIListProvider<>`, and `ICollection`
* `TryGetNonEnumeratedCount` does not test `IReadOnlyCollection<>`

The cases where this difference applies are:
* `IReadOnlyCollection<>`
* Certain `IEnumerable`s returned by `System.Linq.Enumerable`, which implement `IIListProvider<>`, but not `ICollection<>`.

atifaziz:
> The only concern I'd have is that in some cases, there is a performance advantage to have in .net6+ by using `IEnumerable.TryGetNonEnumeratedCount()`.

That's an improvement rather than a concern 😉 and happy to take a PR on it, but the focus in this PR was to fix the bug as a priority.

PR #947, opened 2023-01-26
Improve Error Checking Code

### LINKED ISSUE #903: Update Guard statements
In most operators, we have guard statements that are executed immediately, while the bulk of the operator is implemented in a separate method (see #837 for discussion on why this pattern is used). Currently, the guard statements cannot be inlined to the calling method. This is because `throw` statements are used directly, and `throw` statements prevent the JIT from inlining any method. 

We should move guard statements to separate methods, which would allow them to be inlined in release builds. In some cases, this would be that due to available context information to the JIT, they can be eliminated entirely and the final JIT code only has a direct call to the internal operator code.

Personal recommendation: add a reference to [`CommunityToolkit.Diagnostics`](https://www.nuget.org/packages/CommunityToolkit.Diagnostics/), and update all guard statements to these. This would allow us to rely on a shared implementation of this boilerplate code and if others do as well, then that is mutual code shared.

Alternative recommendation: Add our own `Guard` and `ThrowHelper` methods and update all guard statements to these. This would duplicate code that could be shared, but not add a dependency on another library.

viceroypenguin:
This PR migrates code contracts for methods to `CommunityToolkit.Diagnostics` support methods. This has several advantages:
* Improves chance that error-checking host method will be inlined, possibly improving performance
* Reduces possibility human error by specifying parameter name only once
* Reduces boilerplate code.

Fixes #903 

Tasks:
* [x] `ArgumentNullException`
* [x] `ArgumentOutOfRangeException`
* [x] `ArgumentException`
* [x] `InvalidOperationException` (??)

### DIFF
--- a/MoreLinq.Test/Extensions.cs
+++ b/MoreLinq.Test/Extensions.cs
@@ -17,6 +17,7 @@
 
 namespace MoreLinq.Test
 {
+    using CommunityToolkit.Diagnostics;
     using System;
     using System.Diagnostics;
     using System.Globalization;
@@ -39,7 +40,7 @@ namespace MoreLinq.Test
         [DebuggerStepThrough]
         public static string ToInvariantString<T>(this T formattable, string? format) where T : IFormattable
         {
-            if (formattable is null) throw new ArgumentNullException(nameof(formattable));
+            Guard.IsNotNull(formattable);
             return formattable.ToString(format, CultureInfo.InvariantCulture);
         }
     }
--- a/MoreLinq/Append.cs
+++ b/MoreLinq/Append.cs
@@ -17,7 +17,7 @@
 
 namespace MoreLinq
 {
-    using CommunityToolkit.Diagnostics;
+	using CommunityToolkit.Diagnostics;
     using System.Collections.Generic;
 
     static partial class MoreEnumerable
--- a/MoreLinq/Append.cs
+++ b/MoreLinq/Append.cs
@@ -17,7 +17,7 @@
 
 namespace MoreLinq
 {
-    using System;
+    using CommunityToolkit.Diagnostics;
     using System.Collections.Generic;
 
     static partial class MoreEnumerable
@@ -37,7 +37,7 @@ namespace MoreLinq
         public static IEnumerable<T> Append<T>(this IEnumerable<T> head, T tail)
 #endif
         {
-            if (head == null) throw new ArgumentNullException(nameof(head));
+            Guard.IsNotNull(head);
             return head is PendNode<T> node
                 ? node.Concat(tail)
                 : PendNode<T>.WithSource(head).Concat(tail);
--- a/MoreLinq/Fold.cs
+++ b/MoreLinq/Fold.cs
@@ -31,7 +31,7 @@ namespace MoreLinq
             {
                 elements[i] = i < count ? item : throw LengthError(1);
                 i++;
-            }
+        }
 
             if (i < count)
                 throw LengthError(-1);
@@ -39,6 +39,6 @@ namespace MoreLinq
             return elements;
 
             InvalidOperationException LengthError(int cmp) => new(FormatSequenceLengthErrorMessage(cmp, count));
-        }
     }
+}
 }
--- a/MoreLinq/Fold.g.cs
+++ b/MoreLinq/Fold.g.cs
@@ -27,6 +27,7 @@
 
 namespace MoreLinq
 {
+	using CommunityToolkit.Diagnostics;
     using System;
     using System.Collections.Generic;
 
--- a/MoreLinq/Fold.g.cs
+++ b/MoreLinq/Fold.g.cs
@@ -27,7 +27,7 @@
 
 namespace MoreLinq
 {
-	using CommunityToolkit.Diagnostics;
+    using CommunityToolkit.Diagnostics;
     using System;
     using System.Collections.Generic;
 
--- a/MoreLinq/Fold.g.tt
+++ b/MoreLinq/Fold.g.tt
@@ -32,7 +32,7 @@
 
 namespace MoreLinq
 {
-	using CommunityToolkit.Diagnostics;
+    using CommunityToolkit.Diagnostics;
     using System;
     using System.Collections.Generic;
 
--- a/MoreLinq/Fold.g.tt
+++ b/MoreLinq/Fold.g.tt
@@ -32,6 +32,7 @@
 
 namespace MoreLinq
 {
+	using CommunityToolkit.Diagnostics;
     using System;
     using System.Collections.Generic;
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -86,3 +86,7 @@ dotnet_diagnostic.IDE0055.severity = suggestion
 
 # IDE0046: Convert to conditional expression
 dotnet_diagnostic.IDE0046.severity = suggestion
+
+# Inform the CA1062 analyzer that Guard.IsNotNull will do null-check for us
+# CA1062: Validate arguments of public methods
+dotnet_code_quality.CA1062.null_check_validation_methods = IsNotNull
--- a/MoreLinq.Test/NullArgumentTest.cs
+++ b/MoreLinq.Test/NullArgumentTest.cs
@@ -17,6 +17,7 @@
 
 namespace MoreLinq.Test
 {
+    using CommunityToolkit.Diagnostics;
     using System;
     using System.Collections;
     using System.Collections.Generic;
@@ -56,9 +57,20 @@ namespace MoreLinq.Test
                 Assert.That(e, Is.InstanceOf<ArgumentNullException>().With.Property(nameof(ArgumentNullException.ParamName)).EqualTo(paramName));
                 Debug.Assert(e is not null);
                 var stackTrace = new StackTrace(e, false);
-                var stackFrame = stackTrace.GetFrames().First();
-                var actualType = stackFrame.GetMethod()?.DeclaringType;
-                Assert.That(actualType, Is.SameAs(typeof(MoreEnumerable)));
+                var stackFrames = stackTrace.GetFrames();
+
+                // `Guard.IsNotNull()` adds additional stack frames
+                // so find first non-CommunityToolkit.Diagnostics
+                for (var i = 0; i < stackFrames.Length; i++)
+                {
+                    var actualType = stackFrames[i]?.GetMethod()?.DeclaringType;
+                    if (actualType?.Namespace != "CommunityToolkit.Diagnostics")
+                    {
+                        // top _non_-`Guard` should be `MoreEnumerable`
+                        Assert.That(actualType, Is.SameAs(typeof(MoreEnumerable)));
+                        break;
+                    }
+                }
             });
 
         static IEnumerable<ITestCaseData> GetCanBeNullTestCases() =>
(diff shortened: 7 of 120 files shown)


atifaziz (changes requested):
Thanks for picking up #903 and your effort here!

I noticed one sweeping so could I start by asking you to fix misalignments due to tabs? You can find these with:

    git grep -P '\t' -- *.cs *.tt *.csproj

I'll get round to a more detailed review after. Thanks!

MoreLinq/Fold.cs
```diff
@@ -31,14 +31,14 @@ static T[] Fold<T>(this IEnumerable<T> source, int count)
             {
                 elements[i] = i < count ? item : throw LengthError(1);
                 i++;
-            }
+        }
```
atifaziz, line 34: Revert:

```suggestion
            }
```

MoreLinq/Fold.cs
```diff
@@ -31,14 +31,14 @@ static T[] Fold<T>(this IEnumerable<T> source, int count)
...
             if (i < count)
                 throw LengthError(-1);
 
             return elements;
 
             InvalidOperationException LengthError(int cmp) => new(FormatSequenceLengthErrorMessage(cmp, count));
-        }
     }
 }
+}
```
atifaziz, line 44: Something seems to have gone off with the indentation here.

MoreLinq/Fold.g.tt
```diff
@@ -32,6 +32,7 @@
 
 namespace MoreLinq
 {
+    using CommunityToolkit.Diagnostics;
```
atifaziz, line 35: Remove unused import:

```suggestion
```

MoreLinq/Fold.g.cs
```diff
@@ -27,6 +27,7 @@
 
 namespace MoreLinq
 {
+    using CommunityToolkit.Diagnostics;
```
atifaziz, line 30: Remove unused import:

```suggestion
```

MoreLinq/Append.cs
```diff
@@ -17,7 +17,7 @@
 
 namespace MoreLinq
 {
-    using System;
+	using CommunityToolkit.Diagnostics;
```
atifaziz, line 20: Convert tab to 4 spaces:

```suggestion
    using CommunityToolkit.Diagnostics;
```

atifaziz (commented):
Thanks for the follow-up. There are naturally tons of changed files so it's going to take me a few sittings, but I managed get through a quarter of the 120 files and so here are some interim/initial thoughts…

I noticed that we're only using a tiny fraction of the guards from the **[CommunityToolkit]**

- `Guard.IsBetween`
- `Guard.IsBetweenOrEqualTo`
- `Guard.IsGreaterThan`
- `Guard.IsGreaterThanOrEqualTo`
- `Guard.IsLessThan`
- `Guard.IsLessThanOrEqualTo`
- `Guard.IsNotNull`

I'm therefore leaning towards it being simpler to just embed these into the project instead of taking an additional dependency. A downside might be eventual divergence with time, but I don't expect these to change much and it would also enable some flexibility for tailoring. For example, I wonder if [the `Guard` methods shouldn't be changed to _return_ the valid argument, for some future-readiness](https://twitter.com/raboof/status/1683944201275473921).

Do we really need a throw helper for `InvalidOperationException`? I think you had doubts about the value of this too? I don't think it helps the JIT case. Consistency might be one argument, but a bare `throw` still reads simpler. I can understand doing this consistently for arguments, but invalid use seems like an outlier.

[CommunityToolkit]: https://www.nuget.org/packages/CommunityToolkit.Diagnostics/8.2.2

viceroypenguin:
> I'm therefore leaning towards it being simpler to just embed these into the project instead of taking an additional dependency. A downside might be eventual divergence with time, but I don't expect these to change much and it would also enable some flexibility for tailoring. 

If you want to take the time to copy them from CommunityToolkit and implement them yourself, I won't disagree. That said, a) I'm far too lazy to redo what someone else has already done, and b) I think this goes against principles of working with libraries in the first place.

Namely if _every_ library copies these methods, then now these methods are implemented in each library, increasing dll size and JIT size since they all have to be JIT'd separately, especially for generic methods which need to be genericized separately for each implementation. 

Using a single library (CommunityToolkit for cross-TFM, BCL for net8+ which has many of these now implemented) means that this code is shared across all consumers. 

> For example, I wonder if [the Guard methods shouldn't be changed to return the valid argument, for some future-readiness](https://twitter.com/raboof/status/1683944201275473921).

I used to think that too, when I started with the Throw library (from amantinband) and switched to CommunityToolkit. However, I've found as I've used `Guard.Xxx` more, I now prefer the non-return version because it encourages better code readability, specifically, that contractual obligations (not-null, not-null/empty/whitespace, numeric range) are specified and checked at the start of the method.

> Do we really need a throw helper for `InvalidOperationException`? I think you had doubts about the value of this too? 

The advantage of this specifically is two-fold:
1. the calling code reduces JIT code as the `throw` creates a fairly large footprint
2. calling code is no-longer allowed to be inlined (inlining is turned off for any method that includes a `throw` anywhere in the method). This may not be as big of a deal for a `throw` in the middle of an iterator method, but definitely has an impact on methods that simply return a value. Obviously all of the outer methods that MoreLinq uses are more likely to be inlined, switching to `Guard.Xxx`, which means many of them may turn into direct calls to the `Core` method if the JIT can prove that the values are not-null, etc.
   
   Whether this matters in the cases of the various `ThrowHelper` calls is less certain, but definitely `TryAsCollectionLike()` and `AggregateRight()` will benefit from it. For me, I've grown used to it, so I now actually prefer `ThrowHelper.` instead of `throw new`, but I'll concede that to be a usage/personal preference.

atifaziz (changes requested):
I did another review session and posted some comments. As I did the review, I had some more overall thoughts about this.

We're only using a handful of `Guard` and `ThrowHelper` methods from the CommunityToolkit and most of them are plain one-liners. What's more, practically all of the ones we need have found their way into .NET 8. Some of the complexity that comes from the [friendly type string formatting](https://github.com/CommunityToolkit/dotnet/blob/v8.2.2/src/CommunityToolkit.Diagnostics/Extensions/TypeExtensions.cs) that CommunityToolkit does in the error messages doesn't seem to be warranted. This is not about redoing or not leveraging a library, but assessing whether taking a whole dependency is worth 7 dead simple methods that are landing in .NET anyway. The considerations and criteria are often different for an application than an F/OSS library.

I think we should leverage the methods added in .NET instead of relying on the CommunityToolkit. For targets where the helpers are unavailable, we can copy over the 1-2 line implementations from .NET and light them up conditionally. These will only need to be maintained until those targets have been dropped, although I don't expect there'll any maintenance to do since they can't be expected to change much at all with time.

In related issue, you proposed two options:

> Personal recommendation: add a reference to [`CommunityToolkit.Diagnostics`](https://www.nuget.org/packages/CommunityToolkit.Diagnostics/), and update all guard statements to these. This would allow us to rely on a shared implementation of this boilerplate code and if others do as well, then that is mutual code shared.
>
> Alternative recommendation: Add our own `Guard` and `ThrowHelper` methods and update all guard statements to these. This would duplicate code that could be shared, but not add a dependency on another library.

I'm still leaning on the second one, but saying that let's go with what's there & coming with .NET.

---
PS The whole business of helping JIT inline the argument validation is interesting for methods that are expected to be used on hot paths, but since that's where one generally avoids LINQ anyway, the overall benefits are marginal. Moreover, tiered compilation and PGO may never inline in the first place so one may not even be getting the benefits one might expect!

MoreLinq.Test/Extensions.cs
```diff
@@ -17,6 +17,7 @@
 
 namespace MoreLinq.Test
 {
+    using CommunityToolkit.Diagnostics;
```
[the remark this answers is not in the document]
viceroypenguin: auto-add code-fix. usually it respects that, but alas...

viceroypenguin:
> For targets where the helpers are unavailable, we can copy over the 1-2 line implementations from .NET and light them up conditionally. These will only need to be maintained until those targets have been dropped, although I don't expect there'll any maintenance to do since they can't be expected to change much at all with time.

This is not a bad suggestion. I'll have to think about how to implement in a way that a) isn't too complicated and b) doesn't require tfm based evaluations at the call site. 

What are your thoughts about using global usings to hide the complexity. Could do something like:
```xml
<ItemGroup Condition="$(TargetFramework) != 'net8.0'">
  <Using Include="MoreLinq.Exceptions.ArgumentNullException" Static="true"/>
</ItemGroup>
<ItemGroup Condition="$(TargetFramework) == 'net8.0'">
  <Using Include="System.ArgumentNullException" Static="true"/>
</ItemGroup>
```
```cs
public static IEnumerable<TSource> Operator<TSource>(this IEnumerable<TSource> source)
{
  ThrowIfNull(source);
  // ...
}
```

atifaziz:
> What are your thoughts about using global usings to hide the complexity.

@viceroypenguin I was thinking something along the very same lines! However, instead of statically importing methods, I'd like to try and maintain the type name qualification. Taking the example of one of the `Pad` overloads, ideally the code looks as you'd write it in .NET 8:

```c#
public static IEnumerable<TSource> Pad<TSource>(this IEnumerable<TSource> source, int width,
                                                Func<int, TSource> paddingSelector)
{
    ArgumentNullException.ThrowIfNull(source);
    ArgumentNullException.ThrowIfNull(paddingSelector);
    ArgumentOutOfRangeException.ThrowIfLessThan(width, 0);

    return PadImpl(source, width, default, paddingSelector);
}
```

The targets that won't have those methods on types could have masking definitions in the `MoreLinq` namespace (which could be made to take precedence), like so:

```c#
#if EXCEPTION_HELPERS

namespace MoreLinq
{
    using System.Diagnostics.CodeAnalysis;
    using System.Runtime.CompilerServices;

    static class ArgumentNullException
    {
        public static void ThrowIfNull([NotNull] object? argument,
                                       [CallerArgumentExpression(nameof(argument))]
                                       string? paramName = null)
        {
            if (argument is null)
                Throw(paramName);
        }

        [DoesNotReturn]
        static void Throw(string? paramName) =>
            throw new System.ArgumentNullException(paramName);
    }
}

#endif
```

A conditional symbol like `EXCEPTION_HELPERS` could then be defined in the project file based on the target framework. This is just a rough idea from the head and a little bit of experimentation may be needed to find the sweet spot.

This does mean however that you can't just write `throw new ArgumentNullException()` anymore, but that may not be a bad thing!

I would also limit this just to the main library project. I don't think we need to extend this approach, for example, to the test project.

What would be the cherry on top is to add an integration test that exercises the inlining (although as I said previously, tiered compilation means you don't always get the expected optimisations). I have some ideas for how this could be pulled off.

atifaziz:
@viceroypenguin As a pulse check, do you plan to follow-up here based on latest discussion? I'm just asking to triage open issues and PRs.

viceroypenguin:
Yes, but will probably be a while. That's a bit of more work than I have time for right now.

viceroypenguin:
@atifaziz Question on which path you want to go here. 

Currently, the following code does not work:
```cs
#if !NET6_0_OR_GREATER
global using ArgumentNullException = MoreLinq.Exceptions.ArgumentNullException;
#endif
```

The reason is that each `namespace MoreLinq { using System; }` creates a tighter binding for `ArgumentNullException` to `System` than to the global alias. 

I don't have this problem in SuperLinq, because I migrated to implicit usings and file-scoped namespaces. Since these are the recommended patterns, I'd recommend we do a PR to do these migrations. 

Alternatively, each file can include 
```cs
#if !NET6_0_OR_GREATER
global using ArgumentNullException = MoreLinq.Exceptions.ArgumentNullException;
#endif
```
at the top of the file, below the `using System;`. I think this path is a huge detriment to the usability of each file.

Your call, though.

PR #955, opened 2023-01-29
Remove testing on .NET Core 3.1 that's no longer supported

atifaziz:
This PR removes testing on [.NET Core 3.1](https://dotnet.microsoft.com/en-us/download/dotnet/3.1) that's no longer supported since last month (December 13, 2022).


viceroypenguin:
Even though 3.1 is out of support, many companies are remaining on it; I would keep it in the active testing set for the time being.

atifaziz:
> Even though 3.1 is out of support, many companies are remaining on it; I would keep it in the active testing set for the time being.

We can always revert if there is an uptick in bugs being reported (over this year) where we're causing accidental regressions. If those bug reports never come then this change helps reduce maintenance, build and test times.

PR #959, opened 2023-01-30
Make use of `message` in `Debug.Assert`

i3arnon:
I assume the argument expression in `message` was originally meant to be used.


atifaziz (approved):
Thanks for fixing this oversight!

Issue #676, opened 2019-11-09
Perform the tests against IEnumerable, ICollection, IList, ...


Orace:
Some methods like SkipLast:

https://github.com/morelinq/MoreLINQ/blob/c3516e335230f2a598ea0ab8f570b3d1b45b0970/MoreLinq/SkipLast.cs#L43-L48

Use different strategies if the received `IEnumerable` parameters can be cast to `ICollection` or `IList`.
But the tests may not be done against all those types.
In such case we have to perform the tests with all relevant object type.

First we have to list those methods:

- [ ] SkipLast
- ...

Orace:
This is related to #670

atifaziz:
> In such case we have to perform the tests with all relevant object type.

Yes, I agree. Please go ahead with a PR if you would like to help with this. Thanks!

PR #960, opened 2023-01-30
Add new `SourceKind`: `Collection` and `List` and use it for `EndsWith` tests.

Orace:
Fix for #676/EndsWith.

This is WIP, I'm not happy on how all `Sequence`/`Collection`/`List` cases are built.  
I didn't find yet a nice way to mix `TestCase` and `Values` NUnit attributes.


atifaziz:
> This is WIP

I've converted this PR to a draft.

PR #961, opened 2023-02-01
Simplify `Fold` Implementation

viceroypenguin:
This PR simplifies the implementation of `Fold`. The existing `FoldImpl` is unnecessary and onerous to maintain. Since it receives 18 arguments, it must check which one should be non-`null` and then it must select that one for `Fold`ing. This a) creates additional branches and code bloat, and b) adds an additional method that must be maintained. 

It is better to have each overload of `Fold` do what it needs to do directly. This avoids branching and the `Assume.NotNull()` check, and simplifies the code significantly.

### DIFF
--- a/MoreLinq/Fold.g.tt
+++ b/MoreLinq/Fold.g.tt
@@ -34,7 +34,6 @@ namespace MoreLinq
             select new
             {
                 Ts            = string.Join(", ", Enumerable.Repeat("T", i)),
-                Count         = i,
                 CountElements = istr + " " + (i == 1 ? "element" : "elements"),
                 CountArg      = istr,
                 Elements      = string.Join(", ", from j in Enumerable.Range(0, i)
@@ -64,7 +63,7 @@ namespace MoreLinq
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (folder == null) throw new ArgumentNullException(nameof(folder));
 
-            var elements = source.Fold(<#= e.Count #>);
+            var elements = source.Fold(<#= e.CountArg #>);
             return folder(<#= e.Elements #>);
         }
 
--- a/MoreLinq/Fold.g.tt
+++ b/MoreLinq/Fold.g.tt
@@ -64,10 +64,7 @@ namespace MoreLinq
             if (source == null) throw new ArgumentNullException(nameof(source));
             if (folder == null) throw new ArgumentNullException(nameof(folder));
 
-            var elements = new T[<#= e.Count #>];
-            foreach (var e in AssertCountImpl(source.Index(), <#= e.Count #>, OnFolderSourceSizeErrorSelector))
-                elements[e.Key] = e.Value;
-
+            var elements = source.Fold(<#= e.Count #>);
             return folder(<#= e.Elements #>);
         }
 
--- a/MoreLinq/Fold.g.tt
+++ b/MoreLinq/Fold.g.tt
@@ -38,7 +38,7 @@ namespace MoreLinq
                 CountElements = istr + " " + (i == 1 ? "element" : "elements"),
                 CountArg      = istr,
                 FolderArgs    = "folder" + istr + ": folder",
-
+                Elements      = string.Join(", ", Enumerable.Range(0, i).Select(j => $"elements[{j}]")),
             };
 
         foreach (var e in overloads) { #>
@@ -61,7 +61,14 @@ namespace MoreLinq
 
         public static TResult Fold<T, TResult>(this IEnumerable<T> source, Func<<#= e.Ts #>, TResult> folder)
         {
-            return FoldImpl(source, <#= e.CountArg #>, <#= e.FolderArgs #>);
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (folder == null) throw new ArgumentNullException(nameof(folder));
+
+            var elements = new T[<#= e.Count #>];
+            foreach (var e in AssertCountImpl(source.Index(), <#= e.Count #>, OnFolderSourceSizeErrorSelector))
+                elements[e.Key] = e.Value;
+
+            return folder(<#= e.Elements #>);
         }
 
 <#      } #>
--- a/MoreLinq/Fold.g.tt
+++ b/MoreLinq/Fold.g.tt
@@ -37,8 +37,8 @@ namespace MoreLinq
                 Count         = i,
                 CountElements = istr + " " + (i == 1 ? "element" : "elements"),
                 CountArg      = istr,
-                Elements      = string.Join(", ", Enumerable.Range(0, i)
-                                                            .Select(j => FormattableString.Invariant($"elements[{j}]"))),
+                Elements      = string.Join(", ", from j in Enumerable.Range(0, i)
+                                                  select FormattableString.Invariant($"elements[{j}]")),
             };
 
         foreach (var e in overloads) { #>
--- a/MoreLinq/Fold.g.tt
+++ b/MoreLinq/Fold.g.tt
@@ -37,8 +37,8 @@ namespace MoreLinq
                 Count         = i,
                 CountElements = istr + " " + (i == 1 ? "element" : "elements"),
                 CountArg      = istr,
-                FolderArgs    = "folder" + istr + ": folder",
-                Elements      = string.Join(", ", Enumerable.Range(0, i).Select(j => $"elements[{j}]")),
+                Elements      = string.Join(", ", Enumerable.Range(0, i)
+                    .Select(j => FormattableString.Invariant($"elements[{j}]"))),
             };
 
         foreach (var e in overloads) { #>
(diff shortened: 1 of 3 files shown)


MoreLinq/Fold.g.tt
```diff
@@ -38,7 +38,7 @@ namespace MoreLinq
                 CountElements = istr + " " + (i == 1 ? "element" : "elements"),
                 CountArg      = istr,
                 FolderArgs    = "folder" + istr + ": folder",
-
+                Elements      = string.Join(", ", Enumerable.Range(0, i).Select(j => $"elements[{j}]")),
```
atifaziz, line 41: Use culturally invariant formatting:

```suggestion
                Elements      = string.Join(", ", from j in Enumerable.Range(0, i)
                                                  select j.ToString(CultureInfo.InvariantCulture) into j
                                                  select $"elements[{j}]"),
```

MoreLinq/Fold.g.tt
```diff
@@ -61,7 +61,14 @@ namespace MoreLinq
...
         public static TResult Fold<T, TResult>(this IEnumerable<T> source, Func<<#= e.Ts #>, TResult> folder)
         {
-            return FoldImpl(source, <#= e.CountArg #>, <#= e.FolderArgs #>);
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (folder == null) throw new ArgumentNullException(nameof(folder));
+
+            var elements = new T[<#= e.Count #>];
+            foreach (var e in AssertCountImpl(source.Index(), <#= e.Count #>, OnFolderSourceSizeErrorSelector))
+                elements[e.Key] = e.Value;
+
```
atifaziz, line 70: Since `Fold` is a composition of `AssertCount` + `Index`, we might as well compose with `ToArrayByIndex` too to replace the for-loop:

```suggestion
            const int count = <#= e.CountArg #>;
            var indexed = AssertCountImpl(source.Index(), count, OnFolderSourceSizeErrorSelector);
            var elements = indexed.ToArrayByIndex(count, e => e.Key, e => e.Value);
```

MoreLinq/Fold.g.tt
```diff
@@ -61,7 +61,14 @@ namespace MoreLinq
 
         public static TResult Fold<T, TResult>(this IEnumerable<T> source, Func<<#= e.Ts #>, TResult> folder)
         {
-            return FoldImpl(source, <#= e.CountArg #>, <#= e.FolderArgs #>);
```
atifaziz, line 64: Since `FolderArg` is no longer used, it can be removed.

atifaziz (approved):
Minor comments other looks good to merge, thanks!

MoreLinq/Fold.g.tt
viceroypenguin: I don't like adding another operator unnecessarily - for-loop is probably better for now. I do have a better idea that we can use later. See #963.

MoreLinq/Fold.g.tt
viceroypenguin: Removed.

MoreLinq/Fold.g.tt
viceroypenguin: Addressed, using `FormattableString.Invariant` (cleaner)

MoreLinq/Fold.g.tt
```diff
@@ -37,8 +37,8 @@ namespace MoreLinq
                 Count         = i,
                 CountElements = istr + " " + (i == 1 ? "element" : "elements"),
                 CountArg      = istr,
-                FolderArgs    = "folder" + istr + ": folder",
-
+                Elements      = string.Join(", ", Enumerable.Range(0, i)
+                    .Select(j => FormattableString.Invariant($"elements[{j}]"))),
```
atifaziz, line 41: Please fix the code formatting by aligning the chained methods:

```suggestion
                Elements      = string.Join(", ", Enumerable.Range(0, i)
                                                            .Select(j => FormattableString.Invariant($"elements[{j}]"))),
```

And, as a nit, preferably using the query syntax:

```suggestion
                Elements      = string.Join(", ", from j in Enumerable.Range(0, i)
                                                  select FormattableString.Invariant($"elements[{j}]")),
```

Also, since this is only ever used once, I would even suggest to in-line the expression.

---
> Addressed, using `FormattableString.Invariant` (cleaner)

PS Just saying “cleaner” doesn't offer much, but in the absence of any technical/objective argument, what it is saying is that the [original suggestion](https://github.com/morelinq/MoreLINQ/pull/961#discussion_r1096072797) was somehow quick and dirty. I'd either entirely refrain or qualify with a technical/objective argument for the reviewers and future readers/maintainers.

MoreLinq/Fold.g.tt
atifaziz: > I don't like…

It's best to refrain from expressing personal likes/dislikes (brings subjectivity that's often difficult to resolve) and make as objective an argument as possible.

> adding another operator unnecessarily

This is not about necessity; it's about two things:

1. Consistency: that is, either implement an operator entirely in terms of others (if possible) or as a stand-alone. Otherwise it appears to be a random decision as to where the line is being drawn.
2. [Code coverage](https://github.com/morelinq/MoreLINQ/pull/961#issuecomment-1412807872): if an operator is implemented entirely in terms of others then chances are very high that you can maintain high code coverage with very little effort (just basic tests).

In general, [make it work, make it right then make it fast](https://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast) is a good rule of thumb. Writing an operator, entirely if possible, in terms of others helps to make it work quickly so that one can spend the bulk of time making it right (API design and determining any edge cases that may need handling). Eventually, one can make it fast, sometimes even simpler, by removing the compositions. By this last step (which, bear in mind, can come much later, as in a future release), you hopefully have enough unit tests to avoid introducing regressions as it's refactored.

> for-loop is probably better for now.
> I do have a better idea that we can use later. See #963.

Again, just saying “better” here isn't offering much. It would help others to provide more insight into why you think a for-loop is (technically/objectively) better and why #963 is a better idea. Do we have to wait until #963 is added? Do we know it will be added for sure? Is using `ToArrayByIndex` really problematic meanwhile?

atifaziz (changes requested):

MoreLinq/Fold.g.tt
viceroypenguin: The problem I have with `ToArrayByIndex` in this situation is that it adds too many layers to get to the goal _in this case_. Specifically, as suggested, it would be two calls to `Index`, adding all of these double-indexed items (now `ValueTuple<int, ValueTuple<int, T>>`) to a `List`, then looping that list to store it into an array.

I considered `foreach` to be better because it was a single pass of the data set (albeit small) without any of the additional `Index` overhead. 

Keep in mind that this is already unit tested, so the whole point of this PR is faster and/or simpler code, not original writing. It already works, and we already have proof of non-regression. 

I actually came up with an even better solution that I haven't had time to backport to this PR: `var elements = source.AssertCount(n).ToList();`. Doesn't require #963, simple, and avoids unnecessary overhead. If you're not too upset with this, then I'll update next week when I get a chance.

atifaziz: > The problem I have with `ToArrayByIndex` in this situation is that it adds too many layers to get to the goal _in this case_. Specifically, as suggested, it would be two calls to `Index`, adding all of these double-indexed items (now `ValueTuple<int, ValueTuple<int, T>>`) to a `List`, then looping that list to store it into an array.

I'm a bit confused by how you're arriving at these conclusions. There's one call to `Index` that gives you an index paired with its element, a `KeyValuePair<int, T>`, which is later used to populate an appropriately dimensioned array. I don't see where a `List` gets involved. The downside with the `source.AssertCount(n).ToList()` approach is that you end up with several unnecessary allocations: one for [the wrapper list](https://github.com/dotnet/runtime/blob/v7.0.2/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs), one for [the internal array](https://github.com/dotnet/runtime/blob/v7.0.2/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L25) and then possibly several more as the internal array is grown to accommodate more items (because a list starts with a [default capacity of 4](https://github.com/dotnet/runtime/blob/v7.0.2/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L23)) when the final length is known in advance. With `ToArrayByIndex`, you get an array that is sized to the exact needed element count from the start and its elements are populated in a loop. In other words, this doesn't look awfully complex:

```c#
var elements =
    source.AssertCount(count, OnFolderSourceSizeError)
          .Index()
          .ToArrayByIndex(count, e => e.Key, e => e.Value);
```

> then I'll update next week when I get a chance.

Just in the interest of time (because I have some today), I've gone ahead and made an executive decisions/commits. Considering the goal of this PR was to simplify the implementation, which didn't initially use `ToArrayByIndex` either, we don't have to debate this further. What I have done additionally with 1a93ab344dd213d2abdcb3168ec364fda0fdf365 is move common code out of the template.

atifaziz (approved):
Thanks!

MoreLinq/Fold.g.tt
atifaziz: Fixed with 9d6b44c929eb536c36aa18ae5b0373f92c0a125f and 22edea258682c77f4cc4af470dfd93be088038cd.

PR #964, opened 2023-02-07
Add `CopyTo` operator

### LINKED ISSUE #963: Add `CopyTo` Operator
Add a `CopyTo` operator to replicate the behavior of `List<>.CopyTo` or `Array.CopyTo`, with enumerable semantics.

NB: See https://github.com/viceroypenguin/SuperLinq/blob/master/Source/SuperLinq/CopyTo.cs for implementation

viceroypenguin:
This PR adds a `CopyTo` operator, which operates similar to `Array.CopyTo` or `List.CopyTo`. This is useful for copying data from a stream to a pre-existing array or `IList<>`, such as an array rented from an `ArrayPool<>`. 

To match semantics with existing methods, this method will throw an exception if there is not enough room in the specified destination for the input data to be stored. If the consumer does not care about excess data, they can apply the `Take` operator before calling `CopyTo` to avoid the exception.

One difference between this operator and the collection methods is that this version returns an `int`. With the collection methods, it is easy to know pre or post how much data should be copied, because the `Count`/`Length` property is available. For `IEnumerable<>` this information is not as readily available, so `CopyTo` returns how many elements were stored in the destination.

Fixes #963

### DIFF
--- a/MoreLinq/CopyTo.cs
+++ b/MoreLinq/CopyTo.cs
@@ -1,0 +1,172 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2023 Turning Code, LLC. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#endregion
+
+namespace MoreLinq
+{
+    using System;
+    using System.Collections.Generic;
+
+    static partial class MoreEnumerable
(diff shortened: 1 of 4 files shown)


MoreLinq/CopyTo.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (list == null) throw new ArgumentNullException(nameof(list));
+            if (index < 0) throw new ArgumentOutOfRangeException(nameof(index));
+
+            if (list is TSource[] array)
+            {
+                return CopyTo(source, array, index);
+            }
+            else
```
DanielSantaRosaZup, line 148: 💡You don't need the else statement here, because there is no more instruction after it.

MoreLinq/CopyTo.cs
```diff
@@ -0,0 +1,172 @@
...
+                    && source.TryAsCollectionLike() is CollectionLike<TSource> c)
+                {
+                    l.EnsureCapacity(c.Count + index);
+                }
+#endif
+
+                var i = index;
+                foreach (var el in source)
+                {
+                    if (i < list.Count)
```
DanielSantaRosaZup, line 161: 💡 Sometimes you use { } and sometimes you don't use it in if statement. Adopt a single code style.

Issue #906, opened 2022-11-26
Avoid closures due to implementations in local functions


viceroypenguin:
Currently, most operators are implemented as such:
```csharp
public static IEnumerator<TSource> Operator<TSource>(this IEnumerable<TSource> source, ...)
{
   // guard statements
   return _();

   _()
   {
      yield return source.First();
      yield break;
   }
}
```

A better implementation would be the following:
```csharp
public static IEnumerator<TSource> Operator<TSource>(this IEnumerable<TSource> source, ...)
{
   // guard statements
   return _(source, ...);

   static _(IEnumerable<TSource> source, ...)
   {
      yield return source.First();
      yield break;
   }
}
```

With the current implementation, the host method must create a closure class before even doing guard statements, copy parameters, and then call a method on that closure class. This means that there are two closure classes: one for the parameters, and one for the enumerator.

Switching to passing the parameters into the iterator method means that only one closure class has to be created, the one for the enumerator; the parameters will be held by that class. This should simplify and improve the IL, the JIT, and the output JIT-ted code.

atifaziz:
Thanks for bringing this up and shame about the closure because it means reverting all of #290. The local functions really made things pleasant (for all the reasons listed in #290), but I agree that inducing a closure is not worth the cost. In fact, if the local function is going to be static, then it might as well be a private method again.

julealgon:
> if the local function is going to be static, then it might as well be a private method again.

@atifaziz the function being local has other benefits which are significant, such as being scoped only to that function instead of being callable by the entire class.

I don't think you should make them `private` methods again for that reason.

@viceroypenguin do you mind if I ask a related question here? Would this also apply to an `async` method that is split due to validation concerns like this?

```csharp
public Task DoStuffAsync(string argument, CancellationToken cancellationToken)
{
    ArgumentNullException.ThrowIfNull(argument);
    return DoStuffAsync();

    async Task DoStuffAsync()
    {
        await SomethingElseAsync(cancellationToken);
    }
}
```

Would it have similar benefits here to make the local function `static` if possible?

viceroypenguin:
@julealgon Yes, it does make a difference for `async` methods. See two comparisons here:
[...]

PR #965, opened 2023-02-08
Fix collection-optimized paths for `Batch`

viceroypenguin:
The `Batch` operator contains similar problems to those addressed in issue #943/PR #946. This PR removes all collection optimizations, since they are all based on caching the collection count at initial call instead of run-time.

### DIFF
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -134,17 +134,35 @@ namespace MoreLinq.Test
             Assert.That(batches, Is.Empty);
         }
 
-        [Test]
-        public void BatchUsesCollectionCountAtIterationTime()
+        [TestCase(SourceKind.BreakingList)]
+        [TestCase(SourceKind.BreakingReadOnlyList)]
+        [TestCase(SourceKind.BreakingCollection)]
+        public void BatchCollectionSmallerThanSizeUsesCollectionCountAtIterationTime(SourceKind kind)
+        {
+            var list = new List<int> { 1, 2 };
+            var result = list.ToSourceKind(kind, copy: false).Batch(3);
+            list.Add(3);
+
+            result.AssertSequenceEqual(new[] { 1, 2, 3 });
+        }
+
+        [TestCase(SourceKind.BreakingList)]
+        [TestCase(SourceKind.BreakingReadOnlyList)]
+        [TestCase(SourceKind.BreakingCollection)]
+        public void BatchUsesCollectionCountAtIterationTime(SourceKind kind)
         {
             var list = new List<int>(Enumerable.Range(1, 3));
+            var result = list.ToSourceKind(kind, copy: false).Batch(3);
 
-            var result = list.Batch(3);
+            // should use `CopyTo`
             result.AssertCount(1).Consume();
 
             list.Add(4);
-            result.AssertCount(2).Consume();
-            Assert.Pass();
+
+            // should fail trying to enumerate
+            Assert.That(
+                () => result.AssertCount(2).Consume(),
+                Throws.TypeOf<BreakException>());
         }
     }
 }
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -137,32 +137,17 @@ namespace MoreLinq.Test
         [TestCase(SourceKind.BreakingList)]
         [TestCase(SourceKind.BreakingReadOnlyList)]
         [TestCase(SourceKind.BreakingCollection)]
-        public void BatchCollectionSmallerThanSizeUsesCollectionCountAtIterationTime(SourceKind kind)
-        {
-            var list = new List<int> { 1, 2 };
-            var result = list.ToSourceKind(kind, copy: false).Batch(3);
-            list.Add(3);
-
-            result.AssertSequenceEqual(new[] { 1, 2, 3 });
-        }
-
-        [TestCase(SourceKind.BreakingList)]
-        [TestCase(SourceKind.BreakingReadOnlyList)]
-        [TestCase(SourceKind.BreakingCollection)]
         public void BatchUsesCollectionCountAtIterationTime(SourceKind kind)
         {
-            var list = new List<int>(Enumerable.Range(1, 3));
-            var result = list.ToSourceKind(kind, copy: false).Batch(3);
+            var list = new List<int> { 1, 2 };
+            var result = list.AsSourceKind(kind).Batch(3);
 
-            // should use `CopyTo`
-            result.AssertCount(1).Consume();
+            list.Add(3);
+            result.AssertSequenceEqual(new[] { 1, 2, 3 });
 
             list.Add(4);
-
             // should fail trying to enumerate
-            Assert.That(
-                () => result.AssertCount(2).Consume(),
-                Throws.TypeOf<BreakException>());
+            Assert.That(result.Consume, Throws.TypeOf<BreakException>());
         }
     }
 }
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -133,6 +133,16 @@ namespace MoreLinq.Test
             var batches = Enumerable.Empty<int>().ToSourceKind(kind).Batch(100);
             Assert.That(batches, Is.Empty);
         }
+
+        [Test]
+        public void BatchUsesCollectionCountAtIterationTime()
+        {
+            var stack = new List<int>(Enumerable.Range(1, 3));
+            var result = stack.Batch(3);
+            stack.Add(4);
+            result.AssertCount(2).Consume();
+            Assert.Pass();
+        }
     }
 }
 
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -137,9 +137,12 @@ namespace MoreLinq.Test
         [Test]
         public void BatchUsesCollectionCountAtIterationTime()
         {
-            var stack = new List<int>(Enumerable.Range(1, 3));
-            var result = stack.Batch(3);
-            stack.Add(4);
+            var list = new List<int>(Enumerable.Range(1, 3));
+
+            var result = list.Batch(3);
+            result.AssertCount(1).Consume();
+
+            list.Add(4);
             result.AssertCount(2).Consume();
             Assert.Pass();
         }
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -146,7 +146,7 @@ namespace MoreLinq.Test
             result.AssertSequenceEqual(new[] { 1, 2, 3 });
 
             list.Add(4);
-            // should fail trying to enumerate
+            // should fail trying to enumerate because count is now greater than the batch size
             Assert.That(result.Consume, Throws.TypeOf<BreakException>());
         }
     }
--- a/MoreLinq/Batch.cs
+++ b/MoreLinq/Batch.cs
@@ -92,18 +92,18 @@ namespace MoreLinq
                 {
                     case ICollection<TSource> { Count: 0 }:
                     {
-                        yield break;
+                        break;
                     }
                     case ICollection<TSource> collection when collection.Count <= size:
                     {
                         var bucket = new TSource[collection.Count];
                         collection.CopyTo(bucket, 0);
                         yield return resultSelector(bucket);
-                        yield break;
+                        break;
                     }
                     case IReadOnlyCollection<TSource> { Count: 0 }:
                     {
-                        yield break;
+                        break;
                     }
                     case IReadOnlyList<TSource> list when list.Count <= size:
                     {
@@ -111,7 +111,7 @@ namespace MoreLinq
                         for (var i = 0; i < list.Count; i++)
                             bucket[i] = list[i];
                         yield return resultSelector(bucket);
-                        yield break;
+                        break;
                     }
                     case IReadOnlyCollection<TSource> collection when collection.Count <= size:
                     {
@@ -144,6 +144,7 @@ namespace MoreLinq
                             Array.Resize(ref bucket, count);
                             yield return resultSelector(bucket);
                         }
+
                         break;
                     }
                 }
--- a/MoreLinq/Batch.cs
+++ b/MoreLinq/Batch.cs
@@ -19,7 +19,6 @@ namespace MoreLinq
 {
     using System;
     using System.Collections.Generic;
-    using System.Linq;
 
     static partial class MoreEnumerable
     {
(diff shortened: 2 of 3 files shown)


atifaziz (changes requested):
Good catch and thanks for submitting the PR!

Two things:

1. There are [failing tests](https://ci.appveyor.com/project/raboof/morelinq/builds/46181278/job/og7qof29m65t6wlg#L403) that need to be adjusted.
2. The PR removes all optimisations whereas they should have been moved into the iterator method for similar savings at iteration-time.

viceroypenguin:
1. Oops. Wilco.
2. Will restore for now, and issue separate PR to remove them after this PR is pulled in. Discussion to be tabled until then.

atifaziz:
> and issue separate PR to remove them after this PR is pulled in

What's the reason behind removing them at all?

MoreLinq.Test/BatchTest.cs
```diff
@@ -133,6 +133,16 @@ public void BatchEmptySource(SourceKind kind)
...
         }
+
+        [Test]
+        public void BatchUsesCollectionCountAtIterationTime()
+        {
+            var stack = new Stack<int>(Enumerable.Range(1, 3));
+            var result = stack.Batch(3);
+            stack.Push(4);
+            result.AssertCount(2).Consume();
+            Assert.Pass();
```
atifaziz, line 144: This test doesn't look right. It doesn't exercise any of the optimisations for the various collection/list types. Those optimisations only kick in in two cases:

- The collection/list is empty.
- The collection/list count is smaller than the request batch size.

The test has a stack of 4 elements, which is greater than the request batch size of 3.

What you want here is something like the `BatchCollectionSmallerThanSize` test.

atifaziz (changes requested):
Thanks for the follow-up, but I am afraid the test and the implementation don't look right. See my comments.

viceroypenguin:
> Thanks for the follow-up, but I am afraid the test and the implementation don't look right. See my comments.

Bah! Too many branches on too many projects. :)

MoreLinq.Test/BatchTest.cs
viceroypenguin: Didn't realize that `Stack<>` doesn't implement `ICollection<T>`. Don't care about doing full test, just need to validate that the block of code is moved to the proper spot. Switched to `List<>`, which tests appropriately.

MoreLinq/Batch.cs
```diff
@@ -87,69 +86,65 @@ public static IEnumerable<IEnumerable<TSource>> Batch<TSource>(this IEnumerable<
...
+                            count = 0;
+                        }
+
+                        // Return the last bucket with all remaining elements
+                        if (count > 0)
+                        {
+                            Array.Resize(ref bucket, count);
+                            yield return resultSelector(bucket);
+                        }
+                        break;
```
atifaziz, line 147: nit: blank line for clarity:

```suggestion

                        break;
```

MoreLinq.Test/BatchTest.cs
atifaziz: > Don't care about doing full test, …

I don't know how to read that you “don't care”.

> just need to validate that the block of code is moved to the proper spot.

It's not about the physical movement of the code, but the observable behaviour that happens for a number of types at a certain time.

That aside, the test doesn't exercise the right path. It hits the `default` case in the `switch` within the iterator method. If you use a `BreakingList<>`, you'll see the test break:

```c#
[Test]
public void BatchUsesCollectionCountAtIterationTime()
{
    var list = Enumerable.Range(1, 3).ToList();
    var collection = new BreakingCollection<int>(list);
    var result = collection.Batch(3);
    list.Add(4);
    result.AssertCount(2).Consume();
    Assert.Pass();
}
```

Again:

> What you want here is something like the `BatchCollectionSmallerThanSize` test.

atifaziz (changes requested):

MoreLinq.Test/BatchTest.cs
```diff
@@ -133,6 +133,16 @@ public void BatchEmptySource(SourceKind kind)
...
         }
+
+        [Test]
+        public void BatchUsesCollectionCountAtIterationTime()
+        {
+            var stack = new List<int>(Enumerable.Range(1, 3));
+            var result = stack.Batch(3);
+            stack.Add(4);
+            result.AssertCount(2).Consume();
+            Assert.Pass();
```
atifaziz, line 144: [Follow-up](https://github.com/morelinq/MoreLINQ/pull/965#discussion_r1102691206) suggestion:

```suggestion
        [TestCase(SourceKind.BreakingList)]
        [TestCase(SourceKind.BreakingReadOnlyList)]
        [TestCase(SourceKind.BreakingCollection)]
        public void BatchCollectionSmallerThanSizeUsesCollectionCountAtIterationTime(SourceKind kind)
        {
            var list = new List<int> { 1, 2 };
            var result = list.ToSourceKind(kind).Batch(3);
            list.Add(3);
            result.AssertSequenceEqual(new[] { 1, 2, 3 });
        }
```

MoreLinq/Batch.cs
```diff
@@ -87,69 +86,65 @@ public static IEnumerable<IEnumerable<TSource>> Batch<TSource>(this IEnumerable<
             if (size <= 0) throw new ArgumentOutOfRangeException(nameof(size));
             if (resultSelector == null) throw new ArgumentNullException(nameof(resultSelector));
 
-            switch (source)
+            return Impl(); IEnumerable<TResult> Impl()
```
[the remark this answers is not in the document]
viceroypenguin: Used per updated convention set in #944 (and other operators prior to #944). Either a) `_()` is desired, and `ExceptBy`, `Interleave`, `InsertBy`, `EndsWith`, `Await`, and `SortedMerge` should be updated to reflect (and any difficulties with discards should be addressed), or b) a PR should be done to update all operators to `Impl`. 

Either way, currently, both `_` and `Impl` are present enough to be considered convention. Would recommend that this remains as `Impl`, and an issue opened to return at a later date to update all operators to use `Impl`.

MoreLinq.Test/BatchTest.cs
viceroypenguin: Disagree. 

First, the test that you provided `BatchCollectionSmallerThanSizeUsesCollectionCountAtIterationTime` does not run correctly. All three classes (`BreakingCollection`, `BreakingList`, and `BreakingReadOnlyList`) copy their values internally at creation. Any changes to `list` are not propagated to these objects, so the test is done effectively on an unchanged list. `list.Add(3)` is a no-op.

Second, I have copied the code from the test here, to add additional conversation documentation. The important thing that are tracking is not the collection size itself. It is which branch of the `switch` are we using. Previously, the branching is done at the first call-site, so we receive an iterator that will always use `.CopyTo`, no matter whether the size has changed. The size of the array is queried at runtime, so that behavior is not in question here (i.e. if the collection size changes, it will use the appropriate collection size at runtime); however, if the size gets larger than the batch size, then the iterator _used_ should change to one that will split the collection correctly (aka the `default` branch). 

Under this PR, `Batch` correctly evaluates the `switch` at runtime, so when the size of the collection changes, the choice of iteration changes as well. _This_ is the behavior under test here. As such, it is not as important to test every branch under consideration; what's important is to test _that_ the branch selected changes.

Now, it may be important to also test that the collection count is obtained at runtime. In this case, we will need to add an option to `BreakingList` so that we can change the data inside of it after initialization. I'm happy to discuss the possibility of doing this. However, the test as written is correct, and tests the behavior that I intend to test; as such, it will remain.

```csharp
var list = new List<int>(Enumerable.Range(1, 3));

// verify that source <= batch size returns a single element
// since List<> implements ICollection<>, this should use
// `CopyTo` to generate the array.
var result = list.Batch(3);
result.AssertCount(1).Consume();

list.Add(4);

// verify that source > batch size returns two elements
// this test fails prior to PR #965;
// previously, `Batch` returned immediately the iterator
// method for `ICollection<> when Count <= size`, so
// it would return a single element with an array of size 4.
// corrected, `Batch` waits until iteration to decide which
// branch to use.
result.AssertCount(2).Consume();
Assert.Pass();
```

MoreLinq.Test/BatchTest.cs
viceroypenguin: Rejected per [comment](https://github.com/morelinq/MoreLINQ/pull/965#discussion_r1103908079)

MoreLinq.Test/BatchTest.cs
viceroypenguin: Apologies - had more time to debug today.

The issue is that `.ToSourceKind` always does a `.ToList()`, which makes a copy. Your original test in this comment is valid, but the test as written in the code suggestion was not, due to the copy being made. 

Updated `.ToSourceKind` to accept a `bool copy`, which determines whether to make a copy or not. This way we can specify that we are not intending to copy, but to use the original `List<>` provided. This seemed the simplest way to update, though there are probably better ways.

Added your test, which correctly asserts that the collection count is obtained at runtime. Kept and modified my test to use `ToSourceKind` to validate that the branching is done at runtime.

atifaziz: > The issue is that `.ToSourceKind` always does a `.ToList()`, which makes a copy. Your original test in this comment is valid, but the test as written in the code suggestion was not, due to the copy being made.

That's correct. I was in a hurry and only shared the changes to the test and figured you'd draw the logical conclusion about the changes needed to `ToSourceKind` from there. I could have made the guesswork easier by pointing it out.

Rather than add another flag to disable copying and do safe checks internally, why not go with distinct and strong-typed cases for each that read clearer:

```c#
        internal static IEnumerable<T> AsSourceKind<T>(this List<T> input, SourceKind sourceKind) =>
#pragma warning disable IDE0072 // Add missing cases
            sourceKind switch
#pragma warning restore IDE0072 // Add missing cases
            {
                SourceKind.BreakingList => new BreakingList<T>(input),
                SourceKind.BreakingReadOnlyList => new BreakingReadOnlyList<T>(input),
                SourceKind.BreakingCollection => new BreakingCollection<T>(input),
                SourceKind.BreakingReadOnlyCollection => new BreakingReadOnlyCollection<T>(input),
                _ => throw new ArgumentException(null, nameof(sourceKind))
            };

        internal static IEnumerable<T> ToSourceKind<T>(this IEnumerable<T> input, SourceKind sourceKind) =>
#pragma warning disable IDE0072 // Add missing cases
            sourceKind switch
#pragma warning restore IDE0072 // Add missing cases
            {
                SourceKind.Sequence => input.Select(x => x),
                var kind => input.ToList().AsSourceKind(kind)
            };
```

Here's the same as an overall patch that's ready for application (so there's no guesswork needed):

```diff
diff --git a/MoreLinq.Test/BatchTest.cs b/MoreLinq.Test/BatchTest.cs
index 51fb3360..4083cafb 100644
--- a/MoreLinq.Test/BatchTest.cs
+++ b/MoreLinq.Test/BatchTest.cs
@@ -140,7 +140,7 @@ public void BatchEmptySource(SourceKind kind)
         public void BatchCollectionSmallerThanSizeUsesCollectionCountAtIterationTime(SourceKind kind)
         {
             var list = new List<int> { 1, 2 };
-            var result = list.ToSourceKind(kind, copy: false).Batch(3);
+            var result = list.AsSourceKind(kind).Batch(3);
             list.Add(3);
 
             result.AssertSequenceEqual(new[] { 1, 2, 3 });
@@ -152,7 +152,7 @@ public void BatchCollectionSmallerThanSizeUsesCollectionCountAtIterationTime(Sou
         public void BatchUsesCollectionCountAtIterationTime(SourceKind kind)
         {
             var list = new List<int>(Enumerable.Range(1, 3));
-            var result = list.ToSourceKind(kind, copy: false).Batch(3);
+            var result = list.AsSourceKind(kind).Batch(3);
 
             // should use `CopyTo`
             result.AssertCount(1).Consume();
diff --git a/MoreLinq.Test/TestExtensions.cs b/MoreLinq.Test/TestExtensions.cs
index 95355de3..60f43889 100644
--- a/MoreLinq.Test/TestExtensions.cs
+++ b/MoreLinq.Test/TestExtensions.cs
@@ -87,18 +87,25 @@ internal static IEnumerable<string> GenerateSplits(this string str, params char[
                 yield return split;
         }
 
-        internal static IEnumerable<T> ToSourceKind<T>(this IEnumerable<T> input, SourceKind sourceKind, bool copy = true) =>
+        internal static IEnumerable<T> AsSourceKind<T>(this List<T> input, SourceKind sourceKind) =>
+#pragma warning disable IDE0072 // Add missing cases
             sourceKind switch
+#pragma warning restore IDE0072 // Add missing cases
             {
-                SourceKind.Sequence => copy ? input.Select(x => x) : throw new InvalidOperationException("Invalid to keep original as sequence."),
-                SourceKind.BreakingList => new BreakingList<T>(copy ? input.ToList() : input.SafeCast<List<T>>()),
-                SourceKind.BreakingReadOnlyList => new BreakingReadOnlyList<T>(copy ? input.ToList() : input.SafeCast<List<T>>()),
-                SourceKind.BreakingCollection => new BreakingCollection<T>(copy ? input.ToList() : input.SafeCast<List<T>>()),
-                SourceKind.BreakingReadOnlyCollection => new BreakingReadOnlyCollection<T>(copy ? input.ToList() : input.SafeCast<List<T>>()),
+                SourceKind.BreakingList => new BreakingList<T>(input),
+                SourceKind.BreakingReadOnlyList => new BreakingReadOnlyList<T>(input),
+                SourceKind.BreakingCollection => new BreakingCollection<T>(input),
+                SourceKind.BreakingReadOnlyCollection => new BreakingReadOnlyCollection<T>(input),
                 _ => throw new ArgumentException(null, nameof(sourceKind))
             };
 
-        internal static T SafeCast<T>(this object obj) where T : class =>
-            obj is T t ? t : throw new InvalidOperationException($"Must pass an object of type `{typeof(T)}`.");
+        internal static IEnumerable<T> ToSourceKind<T>(this IEnumerable<T> input, SourceKind sourceKind) =>
+#pragma warning disable IDE0072 // Add missing cases
+            sourceKind switch
+#pragma warning restore IDE0072 // Add missing cases
+            {
+                SourceKind.Sequence => input.Select(x => x),
+                var kind => input.ToList().AsSourceKind(kind)
+            };
     }
 }
```

MoreLinq/Batch.cs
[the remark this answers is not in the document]
atifaziz: Yes, the code base is inconsistent at the moment. `Impl` was used where absolutely necessary due to C# discards quirk-mode. The _majority_ convention was `_` before #944 when an iterator could be a mere local function. In general, the `_` convention for an iterator will need a bigger reconsideration with #906 when even `Impl` won't be a good choice since there won't be much motivation left to keep the iterator as a local function (it could be just lifted out as a stand-alone static method again and then potentially need a unique name).

There is no need to go with `Impl` here so the idea was to keep the majority convention for now.

MoreLinq.Test/BatchTest.cs
```diff
@@ -133,6 +133,37 @@ public void BatchEmptySource(SourceKind kind)
             var batches = Enumerable.Empty<int>().ToSourceKind(kind).Batch(100);
             Assert.That(batches, Is.Empty);
         }
+
+        [TestCase(SourceKind.BreakingList)]
+        [TestCase(SourceKind.BreakingReadOnlyList)]
+        [TestCase(SourceKind.BreakingCollection)]
+        public void BatchCollectionSmallerThanSizeUsesCollectionCountAtIterationTime(SourceKind kind)
```
atifaziz, line 140: The two could be combined into a effectively single test?

```c#
[TestCase(SourceKind.BreakingList)]
[TestCase(SourceKind.BreakingReadOnlyList)]
[TestCase(SourceKind.BreakingCollection)]
public void BatchUsesCollectionCountAtIterationTime(SourceKind kind)
{
    var list = new List<int> { 1, 2 };
    var result = list.ToSourceKind(kind, copy: false).Batch(3);

    list.Add(3);
    Assert.That(result.Consume, Throws.Nothing);

    list.Add(4);
    Assert.That(result.Consume, Throws.TypeOf<BreakException>());
}
```

atifaziz:
@viceroypenguin Any chance we can merge this soon?

atifaziz (approved):
Thanks for this!

PR #980, opened 2023-03-03
Add public API tracking

atifaziz:
This PR adds tracking of the library's publicly shipped API using **[Microsoft.CodeAnalysis.PublicApiAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.PublicApiAnalyzers/3.3.4)**. This adds controls to make sure that no new API or changes leak accidentally.


MoreLinq/Aggregate.g.cs
```diff
@@ -15,6 +15,16 @@
 // limitations under the License.
 #endregion
 
+#nullable enable // required for auto-generated sources (see below why)
```
viceroypenguin, line 18: Would argue these should be done in a separate PR, either before or after this one.

MoreLinq/Extensions.ToDataTable.g.cs
```diff
@@ -27,6 +27,8 @@
 //
 // Source: https://github.com/dotnet/roslyn/blob/70e158ba6c2c99bd3c3fc0754af0dbf82a6d353d/docs/features/nullable-reference-types.md#generated-code
 
+#pragma warning disable RS0041 // Public members should not use oblivious types
```
viceroypenguin, line 30: Would love a comment on why RS0041 is necessary when `#nullable enable` has been done above.

MoreLinq/Aggregate.g.cs
atifaziz: Agree, but for now these were needed to make the CI build for this PR pass. I'll be moving these changes out.

atifaziz: Tracked as bug #981.

MoreLinq/Extensions.ToDataTable.g.cs
atifaziz: What's the specific point of confusion in your mind?

viceroypenguin: It is not obvious on first read which types referenced below are oblivious, and as such, trigger RS0041, requiring it to be disabled. Would guess `DataTable` (as opposed to `Expression<>`, `Func<>`, or `T`/`TTable`, but that is by no means certain or apparent on a naive read. Since this `#pragma` is at the top of the file to deal with it, and is specifically commented vis-a-vis oblivious types, there should be a reference here (line above or on this line, "here" is not specific) for easy read to identify the failure that causes this.

viceroypenguin: `Flatten` is a bit more obvious because `IEnumerable` is a more recognizable non-nullable aware type, but it may be good to add a comment there for specificity as well.

atifaziz: Yes, I agree that it's not obvious. In fact, this suppression of warning is not needed for `ToDataTable`. It's there because it's applied to all output generated by the **MoreLinq.ExtensionsGenerator** project and the real reason it was added/needed was for `Flatten`. In these generated extension wrappers, a file-level suppression of RS0041 doesn't seem to be a real issue because any real problem that it would be helping to flag would have been dealt with in the underlying **MoreLinq** project sources.

atifaziz: There is no `ToDataTable.g.cs` that's generated. It's `Extensions.ToDataTable.g.cs`, which contains the wrapper extension for non-conflicting imports. It's done like for all other _extension_ methods of `MoreEnumerable`.

viceroypenguin: Ah, missed that. 👍🏼

PR #986, opened 2023-03-12
Skip build when testing on CI

### LINKED ISSUE #95: Change ArgumentNullException to early return.
Originally reported on Google Code with ID 95

```
Extension methods can operate on a null instances (http://stackoverflow.com/questions/847209/in-c-what-happens-when-you-call-an-extension-method-on-a-null-object).

Rather than requiring the user always validate that they have a non-null enumerable
before applying an operation, it is nice when the extension method supports null instances.

An example of where this is not the case in MoreLinq is ForEach:
https://code.google.com/p/morelinq/source/browse/MoreLinq/ForEach.cs

Notice that ForEach throws an ArgumentNullException if source is null.  However, I
believe that it would be entirely reasonable to simple return early instead.  This
would allow the user to call ForEach on the enumerable without having to run it through
something like (myEnumerable ?? Enumerable.Empty<MyType>()).

The ForEach is just an example, it would be nice if the entire MoreLinq library followed
this pattern and treated null enumerations the same as empty enumerations.  e.g., Concat
would just return an enumerable with one item if the source was null, the same as you
would get if the source was empty.
```

Reported by `micah@zoltu.net` on 2014-10-17 00:29:21

atifaziz:
This PR fixes #95 by skipping the build step in `test.cmd` and `test.sh` when an environment variable named `SKIP_TEST_BUILD` is defined and its value is anything but `false`. During a CI build `SKIP_TEST_BUILD` is set to `true`.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,7 @@ image:
 environment:
   DOTNET_CLI_TELEMETRY_OPTOUT: true
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+  SKIP_TEST_BUILD: true
 for:
 -
   matrix:
--- a/test.cmd
+++ b/test.cmd
@@ -5,8 +5,9 @@ popd & exit /b %ERRORLEVEL%
 
 :main
 setlocal
-call build ^
-  && call :clean ^
+if not defined SKIP_TEST_BUILD set SKIP_TEST_BUILD=false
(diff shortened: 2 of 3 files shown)


atifaziz:
This change brings the overall CI build time for all jobs down by roughly 6 minutes; that is 20 instead of 26 minutes.

PR #998, opened 2023-04-08
[Proposal] Trim Extension

jamkin:
This is an incomplete PR to propose a new extension method `Trim`. The same can be achieved with an inefficient combination of `SkipWhile`, `Skip`, `Reverse` and `Count`. This is more flexible and efficient. It can also provide a more efficient way than the commonly used `.Trim` method for `string`s, which has to allocate a whole new string.

Looking for any feedback before adding tests, generating the public/shipped API and so forth.

### DIFF
--- a/MoreLinq/Trim.cs
+++ b/MoreLinq/Trim.cs
@@ -1,0 +1,115 @@
+#region License and Terms
+// MoreLINQ - Extensions to LINQ to Objects
+// Copyright (c) 2023 Jamin King. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


atifaziz (commented):
> It can also provide a more efficient way than the commonly used `.Trim` method for `string`s, which has to allocate a whole new string.

If you want to avoid allocations with strings, you can use [`MemoryExtensions.Trim`](https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.trim#system-memoryextensions-trim(system-readonlyspan((system-char)))), as in `str.AsSpan().Trim()` where `str` is a `string`.

atifaziz:
I'm going to assume that this has been abandoned based on [my earlier comment](https://github.com/morelinq/MoreLINQ/pull/998#pullrequestreview-1376754174), but happy to re-open if there's follow-up and arguments for pursuing this further.

Issue #358, opened 2017-10-16
Choose (elements that can be projected)


atifaziz:
`Choose` will apply a function to _optionally_ project a result from each element of a sequence and returns a sequence of results that could be projected successfully.

## Prototype

```c#
public static IEnumerable<TResult> Choose<T, TResult>(this IEnumerable<T> source, Func<T, (bool IsSome, TResult Value)> chooser) =>
    from item in source
    select chooser(item) into option
    where option.IsSome
    select option.Value;
```

## Example

```c#
var xs =
    "O,l,2,3,4,S,6,7,B,9"
        .Split(',')
        .Choose(s => (int.TryParse(s, out var n), n));

Console.WriteLine(string.Join(", ", xs)); // 2, 3, 4, 6, 7, 9
```

## Motivation

You can't use the query syntax due to CS8201:

> Out variable and pattern variable declarations are not allowed within a query clause.

This doesn't work today:

```c#
var q =
    from s in "O,l,2,3,4,S,6,7,B,9".Split(',')
    // CS8201: Out variable and pattern variable declarations are
    //         not allowed within a query clause.
    select int.TryParse(s, out var n)
         ? (IsSome: true, Value: n)
         : (IsSome: false, Value: 0) into n
    where n.IsSome
    select n.Value;
```

The alternative using method chaining, while works, remains nonetheless verbose:

```c#
var xs =
    "O,l,2,3,4,S,6,7,B,9"
        .Split(',')
        .Select(s => int.TryParse(s, out var n)
                   ? (IsSome: true, Value: n)
                   : (IsSome: false, Value: 0))
        .Where(n => n.IsSome)
        .Select(n => n.Value);
```

The `Select` &rarr; `Where` &rarr; `Select` composition does not capture the intention as clearly as `Choose`.

Although not a huge benefit, `Choose` allows immediate projection (in limited cases) to an anonymous type:

```c#
var points =
    "O,l,2,3,4,S,6,7,B,9"
        .Split(',')
        .Choose(s => (int.TryParse(s, out var x), new { X = x, Y = x * 2 }));
```

`Choose` does not require labeling the returned tuple elements as it's clear from the context.

fsateler:
[...]

PR #1001, opened 2023-05-12
Adds HasDuplicates

julienasp:
Add one new extension methods:

-HasDuplicates, an quick an efficient way to find out if you have any duplicates.

### DIFF
--- a/MoreLinq.Test/MoreLinq.Test.csproj
+++ b/MoreLinq.Test/MoreLinq.Test.csproj
@@ -26,12 +26,12 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="NUnit" Version="3.13.3" />
-    <PackageReference Include="NUnit.Analyzers" Version="3.8.0">
+    <PackageReference Include="NUnit" Version="3.14.0" />
+    <PackageReference Include="NUnit.Analyzers" Version="3.9.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="NUnitLite" Version="3.13.3" />
+    <PackageReference Include="NUnitLite" Version="3.14.0" />
     <PackageReference Include="PolySharp" Version="1.13.2">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
@@ -39,7 +39,7 @@
     <PackageReference Include="System.Reactive" Version="6.0.0" />
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'net471'">


julienasp:
@atifaziz @leandromoh @Orace if anyone is available for a quick review :)

MoreLinq.Test/HasDuplicatesTest.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    [TestFixture]
+    public class HasDuplicatesTest
+    {
+        [Test]
+        public void When_Asking_For_Duplicates_On_IEnumerable_Without_Duplicates_Then_False_Is_Returned()
+        {
+            var stringArray = new[]
+            {
+            "FirstElement",
```
viceroypenguin, line 14: nit: incorrect indentation - array initialization elements should be one-tab further than the braces.

MoreLinq/HasDuplicates.cs
```diff
@@ -0,0 +1,27 @@
...
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        ///   Checks if an enumerable of T contains duplicates
+        /// </summary>
+        /// <param name="source">The enumerable of T to validate.</param>
+        /// <param name="comparer">The comparer used for equality. If none is provided, <see cref="EqualityComparer{T}.Default"/> is used</param>
+        /// <typeparam name="T">The item type</typeparam>
+        /// <returns>true if any element of the source enumerable is duplicated, false otherwise</returns>
+        public static bool HasDuplicates<T>(this IEnumerable<T> source, IEqualityComparer<T>? comparer = null)
```
viceroypenguin, line 16: This should follow the pattern of other operators in both System.Linq and MoreLinq where there are two overloads, one with `IEqualityComparer<T>? comparer` and one without. The one without can call the one with using `null` as a parameter.

MoreLinq/ToReadOnlyCollection.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// <summary>
+        ///     Converts an IEnumerable to a ReadOnlyCollection.
+        /// </summary>
+        /// <typeparam name="T">Type of the items in the source enumeration.</typeparam>
+        /// <param name="source">The IEnumerable nullable collection of T to be converted.</param>
+        /// <returns>A ReadOnlyCollection of T as the source.</returns>
+        public static ReadOnlyCollection<T> ToReadOnlyCollection<T>(this IEnumerable<T> source) =>
+            source is null
+                ? throw new ArgumentNullException(nameof(source))
+                : new ReadOnlyCollection<T>(source as IList<T> ?? source.ToList());
```
viceroypenguin, line 19: This operator should not keep a reference to the original `source` after the method ends. On the off chance the consumer provides a list, they may modify the list after they provide it, which would violate the intent of the `ReadOnlyCollection`. 

Because of this ambiguity, I would recommend not implementing this operator in a library at all - the semantics of the use of `source` are not obvious and not everyone will expect the same behavior.

viceroypenguin (commented):
At the very least, this should be two separate PRs. `HasDuplicates` as an operator has nothing to do with `ToReadOnlyCollection`, so combining them on a single PR is definitely improper.

As noted, I would probably not implement `ToReadOnlyCollection` as an operator anyway, due to various semantic ambiguities.

MoreLinq/ToReadOnlyCollection.cs
julienasp: good point
im changing the implementation to prevent that original source reference leak.

would that change your recommandation? I find myself using ReadOnlyCollection alot, and i would love to be able to use it via MoreLinq

viceroypenguin: Not really, for me at least. The ambiguity is there regardless. In any case, remove it from this PR and create a new PR for a more specific conversation.

julienasp:
> At the very least, this should be two separate PRs. `HasDuplicates` as an operator has nothing to do with `ToReadOnlyCollection`, **so combining them on a single PR is definitely improper.**

I can understand why, you would prefer 2 PR, i'll keep that in mind for my future contributions. Thanks!

MoreLinq/ToReadOnlyCollection.cs
julienasp: Done

atifaziz (commented):
@julienasp Thanks for your contribution, but I am a little hesitant to add this because you could argue that MoreLINQ already gives you the operators to efficiently determine if a sequence has duplicates, using [`ScanBy`](https://morelinq.github.io/3.4/ref/api/html/Overload_MoreLinq_MoreEnumerable_ScanBy.htm) together with `Any`. Here's how:

```c#
public static bool HasDuplicates<T>(this IEnumerable<T> source, IEqualityComparer<T>? comparer) =>
    source.ScanBy(e => e, _ => 0, (count, _, _) => count + 1, comparer)
          .Any(e => e is (_, > 1));
```

Here's a breakdown of how this works:

```c#
var strings = new[]
{
    "FirstElement",
    "DUPLICATED_STRING",
    "ThirdElement",
    "duplicated_string",
};

var ds =
    strings.ScanBy(e => e, _ => 0, (count, _, _) => count + 1,
                   StringComparer.OrdinalIgnoreCase);

foreach (var (s, c) in ds)
    Console.WriteLine($"{s} = {c}");

Console.WriteLine();

var hasDuplicates = ds.Any(e => e is (_, > 1));
Console.WriteLine($"Duplicates? {hasDuplicates}");
```

So `ScanBy` is used to count observations per key (`(count, _, _) => count + 1`), starting with zero (`_ => 0`) and where each element serves itself as the key (`e => e`). `Any` is then used to determine when the count exceeds 1.

The above code will print:

    FirstElement = 1
    DUPLICATED_STRING = 1
    ThirdElement = 1
    duplicated_string = 2
    
    Duplicates? True

Note that `Any` will stop consuming `ScanBy` results as soon as the condition is met, which means that the original source is not entirely iterated if not needed. It will be consumed up to the first element that is a duplicate, which is pretty much what your implementation does too and so practically equally efficient.

julienasp:
> ```cs
> source.ScanBy(e => e, e => 0, (count, _, _) => count + 1, comparer)
>           .Any(e => e is (_, > 1))
> ```

I can understand your concern, but isn't that the MoreLINQ mission to make it simple working with collections? And using DRY? 
the ScanBy solution works, and is efficient but lacks lisibility and abstraction. 

You need to decipher the statement to understand that it does a HasDuplicates

julienasp:
> An even simpler version that should be easy to decipher:

Still lacks abstraction and breaks DRY as soon as you use it more than once.

I mean the goal is to add Abstraction over commonly used LINQ pattern, helping with DRY and adding explicit scenarios in the linq ( extension ) API

a quick google search:

- https://stackoverflow.com/questions/5080538/c-sharp-determine-duplicate-in-list (9 years ago)
- https://stackoverflow.com/questions/16197290/checking-for-duplicates-in-a-list-of-objects-c-sharp (10 years ago)
- https://stackoverflow.com/questions/18547354/c-sharp-linq-find-duplicates-in-list (12 years ago)

tons of upvotes, tons of way of doing it... some more efficient than others... and yet people are still wondering HOW to do it, because .NET has not yet abstracted it into a pretty abstraction like HasDuplicates.

The goal should **not** be to figure out how to do a HasDuplicates. It should be to find which nice library like moreLINQ can help me improve my productivity by skipping those google search all together ;) 

If you do not agree with the above, you can close that PR.

Have a great day and thanks for the feedback

MoreLinq.Test/HasDuplicatesTest.cs
```diff
@@ -0,0 +1,99 @@
...
+                "ThirdElement"
+            };
+
+            var hasDuplicates = stringArray.HasDuplicates(null);
+
+            Assert.That(hasDuplicates, Is.False);
+        }
+
+        [Test]
+        public void When_Asking_For_Duplicates_On_IEnumerable_With_Custom_Always_False_Comparer_Then_False_Is_Returned()
```
leandromoh, line 71: should be nice to add a test for opposite scenario: where duplicates exists, but the result is false because you passed a comparer that always returns false. 

also, I missed more 2 important tests:
- check that extension method is lazy (almost every morelinq method has this one)
- check that current implementation does not enumerate unnecessary elements (if you find a duplicate for example in the 3rd element, so you must not iterate the rest of elements)

MoreLinq/HasDuplicates.cs
```diff
@@ -0,0 +1,44 @@
...
+    using System.Linq;
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        ///   Checks if an enumerable of T contains duplicates
+        /// </summary>
+        /// <param name="source">The enumerable of T to validate.</param>
+        /// <typeparam name="T">The item type</typeparam>
+        /// <returns>true if any element of the source enumerable is duplicated, false otherwise</returns>
```
leandromoh, line 14: there is no need to explicit that is an "enumerable of T", just "sequence" is enough. Take a look in other summaries to check this and other patterns used in documentation.

MoreLinq.Test/HasDuplicatesTest.cs
julienasp: Good points! i'll add those after work :) 

Thanks for the feedback

MoreLinq/HasDuplicates.cs
```diff
@@ -0,0 +1,44 @@
...
+        public static bool HasDuplicates<T>(this IEnumerable<T> source)
+        {
+            if (source is null)
+            {
+                throw new ArgumentNullException(nameof(source));
+            }
+
+            var enumeratedElements = new HashSet<T>();
+            return source.Any(element => !enumeratedElements.Add(element));
```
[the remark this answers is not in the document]
julienasp: i'll push something very similar, i think you'll like it!

MoreLinq.Test/HasDuplicatesTest.cs
julienasp: in my case, method isn't lazy since it returns a bool, but i still added a test for that!

README.md
```diff
@@ -306,6 +306,10 @@ selector function.
 
 This method has 6 overloads.
 
+### HasDuplicates
+
+Checks if an enumerable of T contains duplicates.
+
```
Orace, line 312: Documentation of other methods use `sequence` ; ie: "Check if a sequence contains duplicates elements".

Orace:
An overload with a selector (like the one first, min, max, etc have) is easy to implements and are a nice addition. Like
`s.HasDuplicate(e => e.Id)` instead of `s.Select(e => e.Id).HasDuplicate()`.

README.md
julienasp: good catch!

julienasp:
> Overloads with a selector (like the one first, min, max, etc have) are easy to implement and are a nice addition. I mean `s.HasDuplicate(e => e.Id)` instead of `s.Select(e => e.Id).HasDuplicate()`.

i added 2 overloads to enable projection

MoreLinq/PublicAPI/net6.0/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,5 @@
 #nullable enable
+static MoreLinq.MoreEnumerable.HasDuplicates<T>(this System.Collections.Generic.IEnumerable<T>! source) -> bool
+static MoreLinq.MoreEnumerable.HasDuplicates<T>(this System.Collections.Generic.IEnumerable<T>! source, System.Collections.Generic.IEqualityComparer<T>? comparer) -> bool
```
julienasp, line 3: Seems like my definition isn't correctly done
image

MoreLinq/HasDuplicates.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        ///   Checks if sequence contains duplicates
+        /// </summary>
+        /// <param name="source">The sequence to check.</param>
+        /// <typeparam name="T">The type of the elements in the source sequence</typeparam>
+        /// <returns>true if any element of the sequence is duplicated, false otherwise</returns>
+        public static bool HasDuplicates<T>(this IEnumerable<T> source) => source?.HasDuplicates(EqualityComparer<T>.Default) ?? throw new ArgumentNullException(nameof(source));
```
viceroypenguin, line 15: nit: everything after `=>` should be on the next line, for screen width concerns.

MoreLinq/HasDuplicates.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        ///   Checks if sequence contains duplicates, using the specified element equality comparer
+        /// </summary>
+        /// <param name="source">The sequence to check.</param>
+        /// <param name="comparer">The equality comparer to use to determine whether or not keys are equal.
+        /// If null, the default equality comparer for <c>TSource</c> is used.</param>
+        /// <typeparam name="T">The type of the elements in the source sequence</typeparam>
+        /// <returns>true if any element of the sequence is duplicated, false otherwise</returns>
+        public static bool HasDuplicates<T>(this IEnumerable<T> source, IEqualityComparer<T>? comparer) => source?.HasDuplicates(x => x, comparer) ?? throw new ArgumentNullException(nameof(source));
```
viceroypenguin, line 35: Should use `IdFn` instead of `x => x`; for JIT/perf/etc. reasons.

MoreLinq/PublicAPI/net6.0/PublicAPI.Unshipped.txt
viceroypenguin: You're missing the versions that get added in Extensions.g.cs. You'll have another version of each function, which is in the extension class.

viceroypenguin (changes requested):
You also need to run `build.cmd` from the root, so that the `Extensions.g.cs` file can be updated; and the changes in that file must be part of this PR as well.

MoreLinq/HasDuplicates.cs
julienasp: nice, didn't know about that! I'll swap it thanks!

atifaziz:
I just wanted to drop a short note to say that I don't mean to give this any silent treatment. I have been travelling and will be AFK for most of the next two weeks. If I do get a moment or two, I will try and follow-up. If something like `HasDuplicates` hasn't landed in .NET for over a decade (based on the quoted ages of the StackOverflow questions), then hopefully a few more weeks won't kill anyone. 😅

Orace:
*Announcement* (I will delete this post on request): `HasDuplicates` was added to [EnumerationQuest v1.1.0](https://github.com/Orace/EnumerationQuest).

viceroypenguin:
> @viceroypenguin, it actually is `GetHasDuplicates` / `AndHasDuplicates`.
> 
> ```cs
> (hasDuplicate, min, max) = e.GetHasDuplicates().AndMin().AndMax();
> (min, max, hasDuplicate) = e.GetMin().AndMax().AndHasDuplicates();
> ```

@Orace Ah, I just saw `HasDuplicates` as a method in one of the files, so I jumped from that to the idea that the `IEnumerable<>` extension would be `HasDuplicates` as well. Apologies! :)

MoreLinq/Extensions.g.cs
```diff
@@ -2906,6 +2906,57 @@ public static partial class GroupAdjacentExtension
 
     }
 
+    /// <summary><c>HasDuplicates</c> extension.</summary>
+
+    [GeneratedCode("MoreLinq.ExtensionsGenerator", "1.0.0.0")]
+    public static partial class HasDuplicatesExtension
+    {
+        /// <summary>
+        ///   Checks if sequence contains duplicates
```
Orace, line 2915: Shouldn't it be : `/// Checks if a sequence contains duplicates`

MoreLinq.Test/HasDuplicatesTest.cs
```diff
@@ -0,0 +1,231 @@
...
+            public int GetHashCode(string obj) => 0;
+        }
+
+        sealed class DummyStringAlwaysFalseComparer : IEqualityComparer<string>
+        {
+            public bool Equals(string? x, string? y) => false;
+
+            public int GetHashCode(string obj) => 0;
+        }
+    }
```
Orace, line 230: Tests with a `null` source are missing.

viceroypenguin: FYI `null` checks are covered for all operators via `NullArgumentTest`, which uses reflection to generate and run tests.

julienasp: I had some of those at the start, but I removed them after noticing that NullArgumentTest was doing all that already

MoreLinq/HasDuplicates.cs
```diff
@@ -0,0 +1,59 @@
...
+    static partial class MoreEnumerable
+    {
+        /// <summary>
+        ///   Checks if sequence contains duplicates
+        /// </summary>
+        /// <param name="source">The sequence to check.</param>
+        /// <typeparam name="T">The type of the elements in the source sequence</typeparam>
+        /// <returns>true if any element of the sequence is duplicated, false otherwise</returns>
+        public static bool HasDuplicates<T>(this IEnumerable<T> source)
+            => source?.HasDuplicates(EqualityComparer<T>.Default) ?? throw new ArgumentNullException(nameof(source));
```
leandromoh, line 16:
```suggestion
            => HasDuplicates(source, EqualityComparer<T>.Default);
```

MoreLinq/HasDuplicates.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <typeparam name="TSource">Type of the source sequence</typeparam>
+        /// <typeparam name="TKey">Type of the projected element</typeparam>
+        /// <returns>true if any element of the sequence is duplicated, false otherwise</returns>
+        public static bool HasDuplicates<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? comparer)
+        {
+            if (source is null) throw new ArgumentNullException(nameof(source));
+            if (keySelector is null) throw new ArgumentNullException(nameof(keySelector));
+
+            var enumeratedElements = new HashSet<TKey>(comparer);
+            return source.Any(element => !enumeratedElements.Add(keySelector(element)));
```
leandromoh, line 56: maybe should be better to reimplement it with imperative way using a foreach.. not using `Any` and lambdas would use less resources, what is a great thing for all users of a shared library.

MoreLinq/HasDuplicates.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        ///   Checks if sequence contains duplicates, using the specified element equality comparer
+        /// </summary>
+        /// <param name="source">The sequence to check.</param>
+        /// <param name="keySelector">Projection for determining "distinctness"</param>
+        /// <typeparam name="TSource">Type of the source sequence</typeparam>
+        /// <typeparam name="TKey">Type of the projected element</typeparam>
+        /// <returns>true if any element of the sequence is duplicated, false otherwise</returns>
+        public static bool HasDuplicates<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector)
+            => source?.HasDuplicates(keySelector, EqualityComparer<TKey>.Default) ?? throw new ArgumentNullException(nameof(source));
```
leandromoh, line 38:
```suggestion
            => HasDuplicates(source, keySelector, EqualityComparer<TKey>.Default);
```

Orace (approved):
LGTM

julienasp:
> I just wanted to drop a short note to say that I don't mean to give this any silent treatment. I have been travelling and will be AFK for most of the next two weeks. If I do get a moment or two, I will try and follow-up. If something like `HasDuplicates` hasn't landed in .NET for over a decade (based on the quoted ages of the StackOverflow questions), then hopefully a few more weeks won't kill anyone. 😅

if you have time :)

atifaziz:
> if you have time :)

I'll try to find some this week.

julienasp:
> > if you have time :)
> 
> I'll try to find some this week.

friendly reminder 📆

julienasp:
> > > if you have time :)
> > 
> > 
> > I'll try to find some this week.
> 
> friendly reminder 📆

Monthly reminder 👯‍♂️

julienasp:
> > > > if you have time :)
> > > 
> > > 
> > > I'll try to find some this week.
> > 
> > 
> > friendly reminder 📆
> 
> Monthly reminder 👯‍♂️

Monthly reminder 🗓️

atifaziz (changes requested):
Hi @julienasp. Thanks for the friendly and regular reminders, and especially for your patience while I needed to attend to other things before reverting to this.

> LINQ to Objects is missing a few desirable features.
>
> This project enhances LINQ to Objects with extra methods, in a manner which keeps to the spirit of LINQ.
>
> …, but isn't that the MoreLINQ mission to make it simple working with collections?

That's not the mission. It's to provide general and generic algorithms that operate on sequences and compose well with the rest of the ecosystem of operators to compute a desired result. The more general an operator is, the more _reusable_ it is, and so yes:

> And using DRY?

_that is_ the DRY bit. If you have to repeatedly use some lambdas to vary and bend the algorithm to your actual need then that's the _acceptable balance_ of repetition.

> I mean the goal [is/should be] to add Abstraction over commonly used LINQ pattern, helping with DRY and adding explicit scenarios in the linq ( extension ) API

Abstractions can exist at several levels, but adding explicit scenarios contradicts that.

The first and foremost goal that remains is to encode unique sequence operators that encapsulate side-effects and which can be composed together. Sometimes, there's more than one way to skin a cat and as we've added more general operators to the library, we've found that some older ones could be implemented in terms of newer and more general ones. The rule of thumb has been that if something takes 3 or more compositions and very broadly applicable then it may be worth adding. [`Choose`] is one such example, which is essentially [`Select` + `Where` + `Select`]([url](https://github.com/morelinq/MoreLINQ/issues/358)). We have to also consider the cost of maintaining each addition long after the original contributor has moved on.

[`Choose`]: https://morelinq.github.io/4.0/ref/api/html/M_MoreLinq_MoreEnumerable_Choose__2.htm

> the ScanBy solution works, and is efficient but lacks lisibility and abstraction.
> 
> You need to decipher the statement to understand that it does a HasDuplicates

I understand, but I think it's fair to ask the user to add a simple helper method to their project if it helps them to increase readability and avoid repetition. If you want to avoid doing that across several projects, I'd be happy to see someone maintain such small helpers as an add-on to MoreLINQ.

> The goal should **not** be to figure out how to do a HasDuplicates.

The goal of MoreLINQ is to offer basic building blocks and sometimes you have to figure out how to combine those blocks to get the result you're after. The potential problem with adding it to MoreLINQ is scope creep. If we add `HasDuplicates` then the next person who wants a variation on it will ask for another version to be added. For example, beyond just knowing whether there are duplicates or not, one may want to:

- list all duplicates
- list a sample of first _N_ duplicates
- list of duplicates together with their index/position in the sequence

Imagine if everyone wanted to add a version that reads more abstract and readable to them then soon you could get a spaghetti of extensions that you need to try and maintain coherently.

That all said, I think we can strike a compromise here by shooting for the most general case of streaming duplicates in a sequence. Then you can combine it with `Any()` to get the effect of `HasDuplicates` (albeit at the cost of a small allocation) while also leaving the door open for others to have their variations (like combining with `Take` to get _N_ samples or with `Index` to capture positions, etc.). If you're willing to reshape this into an extension called `Duplicates` then I'd be happy to consider this for the next release.

Otherwise, you can vote and argue on issue dotnet/runtime#30582 for the same ask for .NET.

> "GroupBy + Count" was a naive alternative for CountBy, but when we added it the rational was tha same: to provide an efficient (therefore not naive) implementation and well tested abstraction for the user to consume.

@leandromoh Actually, it was simply added as a parallel to [`Seq.countBy` in F#](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-seqmodule.html#countBy) (see #108). There was no other rationale or thinking behind it besides it would allow C# developers to have the same functionality without taking a dependency on F# libraries.

> When we added [IndexBy](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/IndexBy.cs) (which is basically a wrapper for ScanBy) the rational was tha same: to hide the hard part to reason about and make a simple and well tested abstraction for the user to consume.

@leandromoh I just want to make sure that we recall things correctly here. We have always tried to see if something can be implemented more generally, and in that vein, `IndexBy` came _before_ `ScanBy`. In fact, `ScanBy` was born out of generalising `IndexBy` during the PR review. Once the generalization was added, `IndexBy` was written in its terms. I am going through the same thought process here.

atifaziz:
> _Announcement_ (I will delete this post on request): `HasDuplicates` was added to [EnumerationQuest v1.1.0](https://github.com/Orace/EnumerationQuest).

@Orace Lately there's been lots of advertising going on in the repo. One more won't hurt 😅 so don't worry about deleting it.

julienasp:
> Hi @julienasp. Thanks for the friendly and regular reminders, and especially for your patience while I needed to attend to other things before reverting to this.
> 
> > LINQ to Objects is missing a few desirable features.
> > This project enhances LINQ to Objects with extra methods, in a manner which keeps to the spirit of LINQ.
> > …, but isn't that the MoreLINQ mission to make it simple working with collections?
> 
> That's not the mission. It's to provide general and generic algorithms that operate on sequences and compose well with the rest of the ecosystem of operators to compute a desired result. The more general an operator is, the more _reusable_ it is, and so yes:
> 
> > And using DRY?
> 
> _that is_ the DRY bit. If you have to repeatedly use some lambdas to vary and bend the algorithm to your actual need then that's the _acceptable balance_ of repetition.
> 
> > I mean the goal [is/should be] to add Abstraction over commonly used LINQ pattern, helping with DRY and adding explicit scenarios in the linq ( extension ) API
> 
> Abstractions can exist at several levels, but adding explicit scenarios contradicts that.
> 
> The first and foremost goal that remains is to encode unique sequence operators that encapsulate side-effects and which can be composed together. Sometimes, there's more than one way to skin a cat and as we've added more general operators to the library, we've found that some older one could be implemented in terms of newer and more general ones. The rule of thumb has been that if something takes 3 or more compositions and very broadly applicable then it may be worth adding. `Choose` is one such example, which is essentially `Select` + `Where` + `Select`. We have to also consider the cost of maintaining each addition long after the original contributor has moved on.
> 
> > the ScanBy solution works, and is efficient but lacks lisibility and abstraction.
> > You need to decipher the statement to understand that it does a HasDuplicates
> 
> I understand, but I think it's fair to ask the user to add a simple helper method to their project if it helps them to increase readability and avoid repetition. If you want to avoid doing that across several projects, I'd be happy to see someone maintain such small helpers as an add-on to MoreLINQ.
> 
> > The goal should **not** be to figure out how to do a HasDuplicates.
> 
> The goal of MoreLINQ is to offer basic building blocks and sometimes you have to figure out how to combine those blocks to get the result you're after. The potential problem with adding it to MoreLINQ is scope creep. If we add `HasDuplicates` then the next person who wants a variation on it will ask for another version to be added. For example, beyond just knowing whether there are duplicates or not, one may want to:
> 
> * list of all duplicates
> * list a sample of first _N_ duplicates
> * list of duplicates together with their index/position in the sequence
> 
> Imagine if everyone wanted to add a version that reads more abstract and readable added then soon you could get a spaghetti of extensions that you need to try and maintain coherently.
> 
> That all said, I think we can strike a compromise here by shooting for the most general case of streaming duplicates in a sequence. Then you can combine it with `Any()` to get the effect of `HasDuplicates` (albeit at the cost of a small allocation) while also leaving the door open for others to have their variations (like combining with `Take` to get _N_ samples or with `Index` to capture positions, etc.). If you're willing to reshape this into an extension called `Duplicates` then I'd be happy to consider this for the next release.
> 
> Otherwise, you can vote and argue on issue [dotnet/runtime#30582](https://github.com/dotnet/runtime/issues/30582) for the same into .NET.
> 
> > "GroupBy + Count" was a naive alternative for CountBy, but when we added it the rational was tha same: to provide an efficient (therefore not naive) implementation and well tested abstraction for the user to consume.
> 
> @leandromoh Actually, it was simply added as a parallel to [`Seq.countBy` in F#](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-seqmodule.html#countBy) (see #108). There was no other rationale or thinking behind it besides it would allow C# developers to have the same functionality without taking a dependency on F# libraries.
> 
> > When we added [IndexBy](https://github.com/morelinq/MoreLINQ/blob/master/MoreLinq/IndexBy.cs) (which is basically a wrapper for ScanBy) the rational was tha same: to hide the hard part to reason about and make a simple and well tested abstraction for the user to consume.
> 
> @leandromoh I just want to make sure that we recall things correctly here. We have always tried to see if something can be implemented more generally, and in that vein, `IndexBy` came _before_ `ScanBy`. In fact, `ScanBy` was born out of generalising `IndexBy` during the PR review. Once the generalization was added, `IndexBy` was written in its terms. I am going through the same thought process here.

Thanks for the very detailed reply!

I really like your idea, so a Duplicates extension that returns all duplicates based on a predicate, and after that if they want to knows if there's any duplicates they can pair it with IsEmpty or Any

```
collection.Duplicates(x => x.Property).Any()
```
I'll work on that this week.
Thanks again!

viceroypenguin:
@atifaziz I'm sorry, but I'm going to disagree with you here. 

`.Duplicates()` does not seem to be a valuable operator, as it does not add anything (other than naming) that couldn't be done by calling `.IndexBy(predicate).Where(x => x.Value > 1).Select(x => x.Key).DistinctBy(predicate)`, `.CountBy(predicate).Where(x => x.Value > 1)` (depending on whether the writer prefers succinctness or performance/only consuming as many elements as necessary). 

A `.Duplicates()` operator seems overly focused and not generally useful enough, as compared to other operators already present to handle counting. 

However, `.HasDuplicates()` even though it is more focused, is more specifically and commonly useful (again, as evidenced by the number of people looking to know if there are any duplicates in a sequence). 

Note that in [this comment](https://github.com/dotnet/runtime/issues/30582#issuecomment-734443509), the .net bcl person responsible for System.Linq prefers to use `.CountBy()` as a means of implementing `.Duplicates()`, and `.CountBy()` is now available in the bcl, along with `.AggregateBy()`. 

For the comments arguing about consuming the entire sequence, I don't believe that this is a justifiable reason to include `.Duplicates()` when it is trivially implemented using `.IndexBy()` or `.ScanBy()`.

julienasp:
Created a new PR [here](https://github.com/morelinq/MoreLINQ/pull/1037)

but its NullArgumentTest are giving me trouble, i haven't figured it out yet.

atifaziz:
> @atifaziz I'm sorry, but I'm going to disagree with you here.

@viceroypenguin You're entitled to disagree and your mileage may vary. I'm happy with neither adding `Duplicates` nor `HasDuplicates` because, [as I said, you can achieve the same with `ScanBy` and `ScanBy` + `Any`](https://github.com/morelinq/MoreLINQ/pull/1001#pullrequestreview-1431088445) the few times you need to do detect (and list) duplicates. In case `ScanBy` reads tricky for others, I'd leave a comment in the source code and move on, but that's me. Frankly, I'd be even happier if people directed more energy towards issue dotnet/runtime#30582 so they can get it as part of .NET proper. 😅

Since @julienasp [seems content](https://github.com/morelinq/MoreLINQ/pull/1001#issuecomment-1806885967) with the compromise of adding `Duplicates` instead, I'll let you debate with him and convince him otherwise. You have also added `HasDuplicates` to your [personal fork](https://github.com/morelinq/MoreLINQ/pull/1001#issuecomment-1557064233), so I don't understand what pain you're wanting to help us avoid down the road by preventing a more general and streaming version from being added here.

I can imagine the streaming version being potentially helpful to do a check on data for duplicates (thinking millions of records) before submitting an expensive computation that could, for example, cost money on the cloud. A streaming version can stop as soon as a duplicate has been detected _and_ report it. For those working with small data or are okay with a Boolean answer, they can just compose with `Any`. I don't see a reason to choose one over the other when you can have both.

> Note that in [this comment](https://github.com/dotnet/runtime/issues/30582#issuecomment-734443509), the .net bcl person responsible for System.Linq prefers to use `.CountBy()` as a means of implementing `.Duplicates()`, and `.CountBy()` is now available in the bcl, along with `.AggregateBy()`.

Unless I misread or missed something, I don't believe he ever expressed his preference in that comment. He just pointed out how people have been using `Seq.countBy` for the purpose of duplicates detection, which is fine if full consumption of the sequence is not a concern. Also, preferences are subjective.

Ultimately, this seems to be about names like `HasDuplicates`/`Duplicates` just being more discoverable for those who may be largely unfamiliar with the operators and how to use them to achieve the same.

atifaziz:
Superseded by #1037.

PR #1006, opened 2023-06-25
Fix bug in `FillForward` when filler is `null`

### LINKED ISSUE #999: `FillForward` bug when filler is `null`
There is a slight bug in `FillForward`; though it is very obscure and unlikely to be relevant to most users.

Consider the following code:
```csharp
var result = new int?[] { null, 200, null, }
	.FillForward(x => x == 200);

// expected
result == [ null, null, null ]

// actual
result == [ null, 200, null ]
```

The first value of `null` is _not_ blank (`null != 200`); the second value is blank (`200 == 200`). Per the expectations of `FillForward`, the existing value of `null` should be carried forward to replace the `200` value. However, `is (true, { } someSeed)` rejects a `null` value, even if `true`, so we get the value of `200`.

Related: viceroypenguin/SuperLinq#372

atifaziz:
This PR fixes #999.

### DIFF
--- a/MoreLinq/FillForward.cs
+++ b/MoreLinq/FillForward.cs
@@ -112,10 +112,10 @@ namespace MoreLinq
             {
                 if (predicate(item))
                 {
-                    yield return seed is (true, { } someSeed)
+                    yield return seed is (true, var theSeed)
                                ? fillSelector != null
-                                 ? fillSelector(item, someSeed)
-                                 : someSeed
+                                 ? fillSelector(item, theSeed)
+                                 : theSeed
                                : item;
                 }
                 else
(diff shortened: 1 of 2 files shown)


viceroypenguin (approved):

MoreLinq/FillForward.cs
```diff
@@ -112,10 +112,10 @@ static IEnumerable<T> FillForwardImpl<T>(IEnumerable<T> source, Func<T, bool> pr
             {
                 if (predicate(item))
                 {
-                    yield return seed is (true, { } someSeed)
+                    yield return seed is (true, var theSeed)
```
viceroypenguin, line 115:
```suggestion
                    yield return (seed, fillSelector) switch
                    {
                      ((true, var theSeed), { } f) => f(theSeed),
                      ((true, var theSeed), _) => theSeed,
                      _ => item,
                    };
```

atifaziz: I'm going to skip this suggestion since I want the changes for this bug fix to be as minimal and obvious as possible. Changing to a `switch` expression with pattern-matching is a separate issue of whether it reads better than nested conditionals. It belongs in a separate issue/PR.

PR #1011, opened 2023-07-03
refactored slice to avoid twice iteration

leandromoh:


viceroypenguin:
While there are two enumerable methods involved in the previous version, I don't see how it would have iterated the sequence twice - both `Skip` and `Take` are lazy, and neither will pull items unnecessarily. It would be good to update the tests for `Slice` to confirm this, but 

Furthermore, for non-list sequence, `.Skip().Take()` is already heavily optimized by the corelib. It turns into a single `EnumerablePartition<TSource>()` which handles both already. Take a look at the code [here](https://github.com/dotnet/runtime/blob/779d53611efd10526ca2bc3e94ce13ee18ad493a/src/libraries/System.Linq/src/System/Linq/Partition.SpeedOpt.cs) (`Skip` and `Take` are available in the same directory, and you can see that `Skip` creates the `EnumerablePartition<>` and `Take` calls `EnumerablePartition<>.Take()`. [note: since net6.0, `Slice` is actually a synonym for the bcl `Take(IEnumerable<TSource> source, Range range)` operator]

atifaziz (commented):
> While there are two enumerable methods involved in the previous version, I don't see how it would have iterated the sequence twice - both `Skip` and `Take` are lazy, and neither will pull items unnecessarily.

@leandromoh I agree with @viceroypenguin here that there isn't any re-iteration going on here.

> It would be good to update the tests for `Slice` to confirm this, though.

Yup, this should be done first and anyway. This was the idea behind PR #759. Either a commit against it or a separate one for just `Slice` would be welcome.

atifaziz:
Closing this assuming it's unnecessary and abandoned.

PR #1015, opened 2023-10-14
🚧 Add simpler `Unfold` overload

atifaziz:
This PR adds a simpler overload of [`Unfold`] that requires only a single function instead of 3. Example usage:

```c#
var xs = MoreEnumerable.Unfold(1, s => s <= 100 ? (true, s * 2, s) : default);
Console.WriteLine(string.Join(", ", xs));
```

Prints:

    1, 2, 4, 8, 16, 32, 64

Pending work:

- [ ] Add unit tests

[`Unfold`]: https://morelinq.github.io/3.4/ref/api/html/M_MoreLinq_MoreEnumerable_Unfold__3.htm


viceroypenguin:
How is this different than `MoreEnumerable.Generate(1, s => s *2).TakeWhile(s => s <= 100)`?

atifaziz:
> How is this different than `MoreEnumerable.Generate(1, s => s *2).TakeWhile(s => s <= 100)`?

That specific example is not, but `Unfold` anyway is `Generate` + `TakeWhile` + `Select` rolled into one. If `TState` and `TResult` are identical then `Select` can be dropped. Historically, `Generate` was there first, back in 2009. `Unfold` was added later in 2017. Either can be written in terms of the other.

For comparison, below is a slightly more involved example that generates the Fibonacci sequence where all numbers are less than 100. Using `Generate` + `TakeWhile` + `Select`:

```c#
var fibonacciNumbersLowerThan100 =
    MoreEnumerable.Generate((Curr: 0, Next: 1), s => (s.Next, s.Curr + s.Next))
                  .TakeWhile(s => s.Curr < 100)
                  .Select(s => s.Curr);
```

Using `Unfold`:

```c#
var fibonacciNumbersLowerThan100 =
    MoreEnumerable.Unfold((Curr: 0, Next: 1),
                          s => (s.Curr, Next: (Curr: s.Next, Next: s.Curr + s.Next)),
                          s => s.Curr < 100,
                          s => s.Next,
                          s => s.Curr);
```

Using the overload proposed here:

```c#
var fibonacciNumbersLowerThan100 =
    MoreEnumerable.Unfold((Curr: 0, Next: 1),
                          s => s.Curr < 100 ? (true, (s.Next, s.Curr + s.Next), s.Curr) : default);
```

viceroypenguin:
nit: for this example, this would be cleaner as it does not rely on a ternary or `default`
```cs
var fibonacciNumbersLowerThan100 =
    MoreEnumerable.Unfold((Curr: 0, Next: 1),
                          s => (s.Curr < 100, (s.Next, s.Curr + s.Next), s.Curr));
```

That said, I'm curious to whom you see this as a potential benefit.
* I don't see this version of `Unfold` (or even the original one) as simpler than `Generate().TakeWhile().Select()`; on the contrary, I'd argue that `Generate()` is more advantageous because it relies on composability which is both more flexible and more consistent in reading.
* I'm not sure that I see this version of `Unfold` as using less lambdas, because the implementation requires three lambdas, so it is still a four-lambda solution in the end.
* I might be able to buy a performance argument, but I'd have to see a benchmark to confirm the validity.

atifaziz:
> nit: for this example, this would be cleaner as it does not rely on a ternary or `default`

Yes, but with the ternary (for illustration) and more generally, you avoid wasteful computation (like even `s.Curr + s.Next`) if not needed because the break condition's been met.

> That said, I'm curious to whom you see this as a potential benefit.

People used to _unfold_ from other languages and/or runtimes/libraries (such as someone used to [`Seq.unfold`](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-seqmodule.html#unfold) in F#) will find it more aligned and natural.

> * I don't see this version of `Unfold` (or even the original one) as simpler than `Generate().TakeWhile().Select()`; on the contrary, I'd argue that `Generate()` is more advantageous because it relies on composability which is both more flexible and more consistent in reading.

It's not trying to be simpler than `Generate` + `TakeWhile` + `Select`. It's overloading the existing `Unfold` with a simpler version.

> * I'm not sure that I see this version of `Unfold` as using less lambdas, because the implementation requires three lambdas, so it is still a four-lambda solution in the end.

The implementation is immaterial as it can change so the point I was makign is how many lambdas the developer has to author.

> * I might be able to buy a performance argument, but I'd have to see a benchmark to confirm the validity.

There's no performance argument being made for now.

PR #1033, opened 2023-11-04
Add `AggregateOrDefault`

skarllot:
Add the extension method `AggregateOrDefault` allowing to aggregate over empty collections by returning a default value.

### DIFF
--- a/MoreLinq.Test/AggregateOrDefaultTest.cs
+++ b/MoreLinq.Test/AggregateOrDefaultTest.cs
@@ -1,0 +1,70 @@
+﻿namespace MoreLinq.Test
+{
+    using System;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class AggregateOrDefaultTest
+    {
+        [Test]
+        public void AggregateOrDefaultWhenCollectionIsNull()
+        {
+            string[]? collection = null;
+
+            var e1 = Assert.Throws<ArgumentNullException>(
+                () => collection!.AggregateOrDefault((curr, next) => curr + next));
+            var e2 = Assert.Throws<ArgumentNullException>(
+                () => collection!.AggregateOrDefault((curr, next) => curr + next, "<null>"));
+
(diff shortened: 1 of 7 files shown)


atifaziz (commented):
Pskarllot Thanks for this contribution, but isn't this essentially redundant with [`DefaultIfEmpty`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.defaultifempty) + [`Aggregate`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.aggregate)? Taking the examples in your tests:

```c#
var q =
    from input in new[]
    {
        new[] { "john" },
        new[] { "john", "jane", "alex" },
        Enumerable.Empty<string>(),
    }
    select new
    {
        Input = input,
        Result = input.DefaultIfEmpty("<null>")
                      .Aggregate((curr, next) => curr + next)
    };
               
foreach (var e in q)
    Console.WriteLine($"[{string.Join(", ", e.Input)}] = {e.Result}");
```

Prints:

    [john] = john
    [john, jane, alex] = johnjanealex
    [] = <null>

skarllot:
@atifaziz yes, you are right. The only difference that I can see is less allocation (do not create `DefaultIfEmptyIterator` instance).

atifaziz:
> The only difference that I can see is less allocation (do not create `DefaultIfEmptyIterator` instance).

That's true, but it's going to be hard to avoid that with most LINQ compositions/chains that run deeper than a single call. With every addition to the project, one has to the weigh the cost of maintaining the code, documentation and tests (often long after the original contributor may have moved on) and I find that this one isn't worth it on the count of avoiding one allocation. I wish you had opened an issue to declare your intent to add this so I could've avoided you the trouble of through the effort of coding, testing and submitting a PR in the first place. I will close this now, but I hope this does not discourage you from contributing in the future especially because (and for what it's worth stating) this was a good quality PR.

skarllot:
No problem. I thought to open a issue before, but I could use this PR as training to understand this project particularities for future contributions.

PR #1038, opened 2023-11-15
Enable `ImplicitUsings`

viceroypenguin:
This PR enables `ImplicitUsings` on the MoreLinq project and removes (now-)unnecessary `using`s on relevant files. This is a pre-requisite for #947 

Open questions:
* Should we enable `ImplicitUsings` on the other projects as well?
* Are there any non-default `using`s we should add to/remove from the implicit set? I could make the argument to remove `System.Linq` from the set, or to add `System.Collections` to the set of implicit usings, for example.
* Should we use a `GlobalUsings.cs` file instead of enabling `ImplicitUsings`? I prefer `ImplicitUsings`, because I don't like random files with no code, but `GlobalUsings.cs` would be more explicit.

### DIFF
--- a/MoreLinq/Acquire.cs
+++ b/MoreLinq/Acquire.cs
@@ -17,9 +17,6 @@
 
 namespace MoreLinq
 {
-    using System;
-    using System.Collections.Generic;
-
     static partial class MoreEnumerable
     {
         /// <summary>
--- a/MoreLinq/AggregateRight.cs
+++ b/MoreLinq/AggregateRight.cs
@@ -17,10 +17,6 @@
 
 namespace MoreLinq
 {
-    using System;
-    using System.Collections.Generic;
-    using System.Linq;
-
     static partial class MoreEnumerable
     {
         /// <summary>
--- a/MoreLinq/Append.cs
+++ b/MoreLinq/Append.cs
@@ -17,9 +17,6 @@
 
 namespace MoreLinq
 {
-    using System;
-    using System.Collections.Generic;
-
     static partial class MoreEnumerable
     {
         /// <summary>
--- a/MoreLinq/AssemblyInfo.cs
+++ b/MoreLinq/AssemblyInfo.cs
@@ -15,7 +15,6 @@
 // limitations under the License.
 #endregion
 
-using System;
 using System.Reflection;
 
 [assembly: AssemblyTitle("MoreLINQ")]
--- a/MoreLinq/Assert.cs
+++ b/MoreLinq/Assert.cs
@@ -17,9 +17,6 @@
 
 namespace MoreLinq
 {
-    using System;
-    using System.Collections.Generic;
-
     static partial class MoreEnumerable
     {
         /// <summary>
--- a/MoreLinq/AssertCount.cs
+++ b/MoreLinq/AssertCount.cs
@@ -17,9 +17,6 @@
 
 namespace MoreLinq
 {
-    using System;
-    using System.Collections.Generic;
-
     static partial class MoreEnumerable
     {
         /// <summary>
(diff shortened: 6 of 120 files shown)


viceroypenguin:
NB: All of the CodeFactor issues are pre-existing and should be addressed separately.

atifaziz (commented):
> This is a pre-requisite for #947

Are you sure? How? I don't think implicit usings are a great idea, especially because they make it difficult to understand the code base and exclude embedding scenarios. They may have their place in some projects, but don't think this is one of those.

viceroypenguin:
> > This is a pre-requisite for #947
> 
> 
> 
> Are you sure? How? I don't think implicit usings are a great idea, especially because they make it difficult to understand the code base and exclude embedding scenarios. They may have their place in some projects, but don't think this is one of those.
> 
> 

Noted in [this comment](https://github.com/morelinq/MoreLINQ/pull/947#issuecomment-1812614090).

viceroypenguin:
I have a few more minutes to respond, so here goes:

### Problems with `ImplicitUsings`
> exclude embedding scenarios. 
I'll disagree with this for two reasons:
1. Embedding scenarios were common in the late '00s and early '10s, before the rise of nuget. It was a necessary way to include code, because using randomly built DLLs across the internet was unsafe and untrusted. Now that nuget packages are commonplace, people are not embedding code anymore, and we should not be concerned with whether people are embedding MoreLinq code anymore. MoreLinq is clearly popular enough that people are referencing the package directly, so it is unlikely that people are doing copy/paste from MoreLinq into other files.
2. A majority of projects now have `ImplicitUsings` enabled; this is because all of the default templates since net6+ have it enabled in the .csproj, for the convenience of top-level programs. As such, _if_ anyone were to copy files from MoreLinq into their projects, they are more likely to be working with the default `ImplicitUsings`.

> it difficult to understand the code base
What does removing `using System;` make harder to understand? I think most developers can agree that `IDisposable` and `ArgumentNullException` come from the `System` namespace, and `IList<>` comes from the `System.Collection.Generic` namespace. And if they don't know that, then a) are they working with the MoreLinq codebase in the first place, and b) do they _care_? The default implicit usings contain the following namespaces:
```
System
System.Collections.Generic
System.Linq
System.Text
System.Threading.Tasks
```

Which classes from these namespaces are difficult to understand where they come from or how they are used? And which of these classes are likely to overlap with any code written by us or other developers? 

* If this is a concern of knowing exactly which namespace a class comes from, then we should start using `System.Collections.Generic.IEnumerable<>` on every return type for clarity. 
* If this is a concern for clashing, then we should avoid clashing with classes anyway (and there's a CAXXXX for that too). 
* If this is a concern for Intellisense, then modern intellisense automatically includes everything from these five namespaces regardless of whether they're `using` or not - and if you use the class, VS automatically includes a `using` statement for that namespace in the file. 

### Why `ImplicitUsings` is necessary
That all said, the problem with not using `ImplicitUsings` or `global using System;` is that we cannot use a `global using` for `ArgumentNullException` if `using System;` is present at the top of the file. The file and/or namespace `using`s have higher priority than the `global using`s. As such, if each file has `using System;`, then each file must separately have the following code at the top fo the file:
```cs
namespace MoreLinq
{
   using System;

#if !NET6_0_OR_GREATER
   using ArgumentNullException = MoreLinq.Exceptions.ArgumentNullException;
#endif
#if !NET8_0_OR_GREATER
   using ArgumentOutOfRangeException = MoreLinq.Exceptions.ArgumentOutOfRangeException;
#endif

   // code
}
```

I think that this is untenable to have across the project. 

### Options:
1. As written, with the open questions listed above
2. Add a `GlobalUsings.cs` file with the necessary `global using`s, for `System` and for the aliases.
3. use the two `#ifdef`s in a majority of files.

PR #1039, opened 2023-11-15
Update MoreLinq to net8.0

viceroypenguin:
This PR updates MoreLinq to net8.0 and addresses unit test issues created by BCL changes.

Issue #1012, opened 2023-08-05
Hide `ToDictionary()` in net8+


viceroypenguin:
net8.0 has added `ToDictionary()` overloads that overlap the MoreLinq versions. We should hide them in net8+. Similar to #749.

See: https://github.com/dotnet/runtime/pull/85811

PR #1041, opened 2023-11-15
Use .NET 8 SDK to add .NET 8 target

atifaziz:
This PR adds the .NET 8 target (requiring SDK upgrade to 8.0 too), primarily to address #1012.

This PR overlaps with #1039 from @viceroypenguin, which was done in parallel. Any differences will be consolidated here.


MoreLinq.Test/SliceTest.cs
```diff
@@ -122,7 +131,9 @@ public void TestSliceLongerThanSequence()
             var sequenceA = Enumerable.Range(1, count);
             var sequenceB = sequenceA.ToList();
             var resultA = sequenceA.Slice(count / 2, count);
-            var resultB = sequenceB.Slice(count / 2, count);
+            // .NET 8 introduced `Slice` on `List<T>` so use `AsEnumerable` to force the extension
+            // method to be invoked.
+            var resultB = sequenceB.AsEnumerable().Slice(count / 2, count);
```
viceroypenguin, line 136: Would recommend consistency in the various tests here. Either all should use .AsEnumerable() or all should disable ide0057. Up to you which you think is cleaner.

viceroypenguin (commented):
Will read some more later.

MoreLinq.Test/SliceTest.cs
atifaziz: Agree. Lots of code added over time by different contributors needs a consistency refresh.

viceroypenguin: > Agree. Lots of code added over time by different contributors needs a consistency refresh.

I'm talking about the code you added here. Please pick one method or the other for all three changes in this file.

MoreLinq/RandomSubset.cs
```diff
@@ -76,8 +76,11 @@ static IEnumerable<T> RandomSubsetImpl<T>(IEnumerable<T> source, Random rand, Fu
 
             if (array.Length < subsetSize)
             {
+#pragma warning disable CA2208 // Instantiate argument exceptions correctly
+                               // TODO Throw InvalidOperationException instead?
```
viceroypenguin, line 80: `ArgumentOutOfRangeException` is valid here. CA2208 is a good check, but cannot be coded to evaluate this scenario. disabling it for this line is the most correct approach, imo

MoreLinq.Test/SliceTest.cs
atifaziz: The others are marked as not applicable (N/A). Smells like a bug in the analyser? There is no overload of [`Slice` in MoreLINQ](https://morelinq.github.io/4.0/ref/api/html/M_MoreLinq_MoreEnumerable_Slice__1.htm) that allow use of the range operator.

MoreLinq/RandomSubset.cs
atifaziz: Do you know of LINQ operators from the BCL that do argument validate at time of iteration/evaluation?

MoreLinq.Test/SliceTest.cs
viceroypenguin: > The others are marked as not applicable (N/A). Smells like a bug in the analyser? There is no overload of [`Slice` in MoreLINQ](https://morelinq.github.io/4.0/ref/api/html/M_MoreLinq_MoreEnumerable_Slice__1.htm) that allow use of the range operator.

They are correct, and not a bug in the analyzer. You have to switch to net8 to see why they're correct. The issue is the same as described in the comment here, `List<>` added a `.Slice()` method in net8, which is accessible via the range operator. There are four tests in this file that all do `var x = y.ToList(); x.Slice(..);`. In each case, either IDE0057 needs to be suppressed in net8+, or resultB needs to use `sequenceB.AsEnumerable().Slice(...)` to switch to the MoreLinq method instead of the `List<>` method. 

Either is correct; but the same method should be used for all four tests in this file. I would recommend `.AsEnumerable()`, which is what you see in four lines in #1039.

MoreLinq/RandomSubset.cs
viceroypenguin: No, but none of the BCL LINQ operators have any arguments that could be invalid, but are only known to be invalid at iteration.

MoreLinq.Test/SliceTest.cs
atifaziz: I was switched to .NET 8 target in the IDE, but I guess I misinterpreted something when working in a rush to get to a successful build state yesterday. Good to have two eyeballs (and fixed with 6abc4715ea1dd19d71736663c9b2963432bef24a), but going back and reading the tests in more detail today, I think there's lots wrong still. For one thing, `Enumerable.Range` in .NET 8 no longer returns a pure sequence, but [an object that is effectively a list](https://github.com/dotnet/runtime/blob/v8.0.0/src/libraries/System.Linq/src/System/Linq/Range.SpeedOpt.cs#L13) and therefore may be accidentally exercising the optimised path for many operators. That's the case here with `Slice` too. For example, the following test is supposed to test that the optimised list path behaves the same as the one where the source is a pure sequence, but both, `sequenceA` and `sequenceB`, end up being seen by `Slice` as lists effectively:

```c#
/// <summary>
/// Verify that a slice from 0 to sequence.Count() yields the source sequence.
/// NOTE: Since there are two different implementations of Slice() - one optimized that's
///       for lists and one that operates on any <c>IEnumerable{T}</c>, this test examines both.
/// </summary>
[Test]
public void TestSliceIdentity()
{
    const int count = 100;
    var sequenceA = Enumerable.Range(1, count);
    var sequenceB = sequenceA.ToList().AsEnumerable();

    var resultA = sequenceA.Slice(0, count);
    var resultB = sequenceB.Slice(0, count);

    Assert.That(resultA, Is.EqualTo(sequenceA));
    Assert.That(resultB, Is.EqualTo(sequenceB));
}
```

In fact, none of the tests except `TestSliceIsLazy` hit the following line:

https://github.com/morelinq/MoreLINQ/blob/97dbe9874072cdf149c21daa663c22a9de9779c6/MoreLinq/Slice.cs#L60

viceroypenguin: All issues with `Slice` should be addressed via #1043.

viceroypenguin: For `Slice` at least, we can test using the following:
```cs
    var sequenceA = Enumerable.Range(1, count).Select(x => x);
    var sequenceB = sequenceA.ToList();

    var resultA = MoreEnumerable.Slice(sequenceA, 0, count);
    var resultB = MoreEnumerable.Slice(sequenceB, 0, count);
```

The identity `Select` will use one of the `SelectIterator`s, which do implement some performance improvements, but do not implement `IList<>`. Using the `MoreEnumerable.Slice()` pattern will force calling the MoreLinq `.Slice()` method regardless of which methods may or may not exist on the sequence objects.

atifaziz: Tracking this issue in #1044. It doesn't need to be solved in this PR since `Enumerable.Range(1, 10) is IList<int>` returns `false` up to .NET 7 and tests against the various targets will cover the other code path.

MoreLinq/RandomSubset.cs
atifaziz: It can also be left as undefined behaviour then. I don't think `ArgumentOutOfRangeException` is the right call as it could be triggered far from the call site.

atifaziz: For now I think we can close this and revisit when the to-do comment is being address because there's no pressing need to make a binary breaking change here. It's been this way for a while.

viceroypenguin:
[a3b88be](https://github.com/morelinq/MoreLINQ/pull/1041/commits/a3b88be71ffd2205ca8cc5300fc65a80bc78fd5c) I thought about this too. I think I would qualify it as shipped, because even though there's no net8 version released, net8 consumers are consuming the net6 version which does have these methods. The important thing for the shipped/unshipped files is to have awareness via PR which things are added for the consumer in the next update. net8 consumers won't see these functions as new in net8, they'll just have net8 improvements.

atifaziz:
@viceroypenguin Technically speaking, there's no difference but it just seems more honest and less confusing (especially for newcomers) to say it's unshipped for now.

> The important thing for the shipped/unshipped files is to have awareness via PR which things are added for the consumer in the next update.

You might be saying the same thing, but I think it's more to avoid accidentally leaking something into the Public API. It could be eventually used for tacking too.

PR #1046, opened 2023-11-17
Update to using C# 12

atifaziz:
This PR allows the use of the C# 12 going forward and updates the code base to take advantage of primary constructors and collection expressions to reduce ceremony.


viceroypenguin (commented):
Sorry, I didn't get a chance to review this yesterday. LGTM. I would add these four errors to the .editorconfig, though:
```
# Primary Constructors
dotnet_diagnostic.CS9107.severity = error               # CS9107: Parameter is captured into the state of the enclosing type and its value is also passed to the base constructor.
dotnet_diagnostic.CS9113.severity = error               # CS9113: Parameter is unread.
dotnet_diagnostic.CS9124.severity = error               # CS9124: Parameter is captured into the state of the enclosing type and its value is also used to initialize a field, property, or event.
dotnet_diagnostic.CS9179.severity = error               # CS9179: Primary constructor parameter is shadowed by a member from base.
```

This gives confidence in correct behavior with regards to the use of parameter values being used only once and not being overwritten incorrectly.

atifaziz:
> Sorry, I didn't get a chance to review this yesterday. LGTM.

No worries and feel free to comment still if you see something, or better yet, open an issue to track.



> I would add these four errors to the .editorconfig, though:

I don't it's needed since [those warnings](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/constructor-errors#primary-constructor-declaration), or in fact all warnings, will be promoted to errors by the following default:

https://github.com/morelinq/MoreLINQ/blob/80ab81c9e06af0409e621ddeef7f83e4f4ebf0b9/Directory.Build.props#L5

PR #1064, opened 2024-04-25
Removed closures in favour of private static methods to close issue #906

### LINKED ISSUE #906: Avoid closures due to implementations in local functions

### LINKED ISSUE #1065: `Rank` does not iterate source on re-iterations
The following code demonstrates this bug:

```c#
var xs = Enumerable.Range(1, 5)
                   .Select(x => x * 10)
                   .ToArray();

IEnumerable<int> Unstable()
{
    try
    {	        
        foreach (var x in xs)
            yield return x;
    }
    finally
    {
        // rotate array elements by 1
        var x0 = xs[0];
        Array.Copy(xs, 1, xs, 0, xs.Length - 1);
        xs[^1] = x0;
    }
}

var result = Unstable().Rank();
foreach (var _ in xs)
    Console.WriteLine(result.ToDelimitedString(", "));
```

The output should have been:

    5, 4, 3, 2, 1
    4, 3, 2, 1, 5
    3, 2, 1, 5, 4
    2, 1, 5, 4, 3
    1, 5, 4, 3, 2

Instead, it is:

    5, 4, 3, 2, 1
    5, 4, 3, 2, 1
    5, 4, 3, 2, 1
    5, 4, 3, 2, 1
    5, 4, 3, 2, 1

It seems that [`Rank`](https://morelinq.github.io/4.1/ref/api/html/Overload_MoreLinq_MoreEnumerable_Rank.htm) caches the source on the first iteration and reuses it on subsequent re-iterations. The culprit appears to be the following line:

https://github.com/morelinq/MoreLINQ/blob/9e8073d8b502fbc91697d3cdb25b97a23d71a6c4/MoreLinq/Rank.cs#L84

which modifies the original captured `source`.

This bug is present in the [latest release (4.2)](https://www.nuget.org/packages/morelinq/4.2.0).

JacobSilasBentley:
This PR:

- primarily closes #906 where closures are to be removed in favour of using private static methods.
- and incidentally fixes #1065 

One unit test had to be updated. This was to change the max enumerations of the test sequence from 1 to 2. The test enumerates the sequence twice so this should have always been the case. However I think previously using closures meant that test was enumerated once when it should have been twice. Happy to be corrected if I have misunderstood.

This is my first PR for MoreLINQ so apologies if I have missed any guidance on submitting PRs and will happily take onboard feedback

### DIFF
--- a/MoreLinq/CountDown.cs
+++ b/MoreLinq/CountDown.cs
@@ -58,12 +58,16 @@ namespace MoreLinq
             if (resultSelector == null) throw new ArgumentNullException(nameof(resultSelector));
 
             return source.TryAsListLike() is { } listLike
-                   ? IterateList(listLike)
+                   ? IterateList(listLike, source, count, resultSelector)
                    : source.TryAsCollectionLike() is { } collectionLike
-                     ? IterateCollection(collectionLike)
-                     : IterateSequence();
+                     ? IterateCollection(collectionLike, source, count, resultSelector)
+                     : IterateSequence(source, count, resultSelector);
 
-            IEnumerable<TResult> IterateList(ListLike<T> list)
+            static IEnumerable<TResult> IterateList(
+                ListLike<T> list,
+                IEnumerable<T> source,
+                int count,
+                Func<T, int?, TResult> resultSelector)
             {
                 var listCount = list.Count;
                 var countdown = Math.Min(count, listCount);
@@ -72,14 +76,21 @@ namespace MoreLinq
                     yield return resultSelector(list[i], listCount - i <= count ? --countdown : null);
             }
 
-            IEnumerable<TResult> IterateCollection(CollectionLike<T> collection)
+            static IEnumerable<TResult> IterateCollection(
+                CollectionLike<T> collection,
+                IEnumerable<T> source,
+                int count,
+                Func<T, int?, TResult> resultSelector)
             {
                 var i = collection.Count;
                 foreach (var item in collection)
                     yield return resultSelector(item, i-- <= count ? i : null);
             }
 
-            IEnumerable<TResult> IterateSequence()
+            static IEnumerable<TResult> IterateSequence(
+                IEnumerable<T> source,
+                int count,
+                Func<T, int?, TResult> resultSelector)
             {
                 var queue = new Queue<T>(Math.Max(1, count + 1));
 
--- a/MoreLinq/CountDown.cs
+++ b/MoreLinq/CountDown.cs
@@ -58,14 +58,13 @@ namespace MoreLinq
             if (resultSelector == null) throw new ArgumentNullException(nameof(resultSelector));
 
             return source.TryAsListLike() is { } listLike
-                   ? IterateList(listLike, source, count, resultSelector)
+                   ? IterateList(listLike, count, resultSelector)
                    : source.TryAsCollectionLike() is { } collectionLike
-                     ? IterateCollection(collectionLike, source, count, resultSelector)
+                     ? IterateCollection(collectionLike, count, resultSelector)
                      : IterateSequence(source, count, resultSelector);
 
             static IEnumerable<TResult> IterateList(
                 ListLike<T> list,
-                IEnumerable<T> source,
                 int count,
                 Func<T, int?, TResult> resultSelector)
             {
@@ -78,7 +77,6 @@ namespace MoreLinq
 
             static IEnumerable<TResult> IterateCollection(
                 CollectionLike<T> collection,
-                IEnumerable<T> source,
                 int count,
                 Func<T, int?, TResult> resultSelector)
             {
--- a/MoreLinq/EndsWith.cs
+++ b/MoreLinq/EndsWith.cs
@@ -78,14 +78,14 @@ namespace MoreLinq
             return second.TryAsCollectionLike() is { Count: var secondCount }
                    ? first.TryAsCollectionLike() is { Count: var firstCount } && secondCount > firstCount
                      ? false
-                     : Impl(second, secondCount, first, comparer)
-                   : Impl(secondList = second.ToList(), secondList.Count, first, comparer);
+                     : Impl(first, second, secondCount, comparer)
+                   : Impl(first, secondList = second.ToList(), secondList.Count, comparer);
 #pragma warning restore IDE0075 // Simplify conditional expression
 
-            static bool Impl(IEnumerable<T> snd, int count, IEnumerable<T> first, IEqualityComparer<T> comparer)
+            static bool Impl(IEnumerable<T> first, IEnumerable<T> second, int count, IEqualityComparer<T> comparer)
             {
                 using var firstIter = first.TakeLast(count).GetEnumerator();
-                return snd.All(item => firstIter.MoveNext() && comparer.Equals(firstIter.Current, item));
+                return second.All(item => firstIter.MoveNext() && comparer.Equals(firstIter.Current, item));
             }
         }
     }
--- a/MoreLinq/EndsWith.cs
+++ b/MoreLinq/EndsWith.cs
@@ -78,11 +78,11 @@ namespace MoreLinq
             return second.TryAsCollectionLike() is { Count: var secondCount }
                    ? first.TryAsCollectionLike() is { Count: var firstCount } && secondCount > firstCount
                      ? false
-                     : Impl(second, secondCount)
-                   : Impl(secondList = second.ToList(), secondList.Count);
+                     : Impl(second, secondCount, first, comparer)
+                   : Impl(secondList = second.ToList(), secondList.Count, first, comparer);
 #pragma warning restore IDE0075 // Simplify conditional expression
 
-            bool Impl(IEnumerable<T> snd, int count)
+            static bool Impl(IEnumerable<T> snd, int count, IEnumerable<T> first, IEqualityComparer<T> comparer)
             {
                 using var firstIter = first.TakeLast(count).GetEnumerator();
                 return snd.All(item => firstIter.MoveNext() && comparer.Equals(firstIter.Current, item));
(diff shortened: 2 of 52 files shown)


MoreLinq/CountDown.cs
```diff
@@ -58,12 +58,16 @@ static partial class MoreEnumerable
...
                    : source.TryAsCollectionLike() is { } collectionLike
-                     ? IterateCollection(collectionLike)
-                     : IterateSequence();
+                     ? IterateCollection(collectionLike, source, count, resultSelector)
+                     : IterateSequence(source, count, resultSelector);
 
-            IEnumerable<TResult> IterateList(ListLike<T> list)
+            static IEnumerable<TResult> IterateList(
+                ListLike<T> list,
+                IEnumerable<T> source,
```
atifaziz, line 68: `source` isn't used or needed here.

MoreLinq/CountDown.cs
```diff
@@ -72,14 +76,21 @@ IEnumerable<TResult> IterateList(ListLike<T> list)
                     yield return resultSelector(list[i], listCount - i <= count ? --countdown : null);
             }
 
-            IEnumerable<TResult> IterateCollection(CollectionLike<T> collection)
+            static IEnumerable<TResult> IterateCollection(
+                CollectionLike<T> collection,
+                IEnumerable<T> source,
```
atifaziz, line 81: `source` isn't used or needed here.

atifaziz (changes requested):
> This is my first PR for MoreLINQ so apologies if I have missed any guidance on submitting PRs and will happily take onboard feedback

First and foremost, welcome and thank you for your initiative to help with #906.

> One unit test had to be updated. This was to change the max enumerations of the test sequence from 1 to 2. The test enumerates the sequence twice so this should have always been the case. However I think previously using closures meant that test was enumerated once when it should have been twice. Happy to be corrected if I have misunderstood.

Actually you've uncovered a bug that I've documented in #1065! It seems that the sequence was accidentally getting cached by an update of the captured closure. Now that that's not happening with this PR, the test failed. I've proposed the correct fix to the test, which is to materialise the result of `Rank` before asserting. I've also edited the description of this PR to say it fixes #1065.

While you seem to have addressed the bulk of methods, the following may have slipped:

- [`Cartesian`](https://github.com/morelinq/MoreLINQ/blob/9e8073d8b502fbc91697d3cdb25b97a23d71a6c4/MoreLinq/Cartesian.g.tt#L118)
- [`Subsets`](https://github.com/morelinq/MoreLINQ/blob/9e8073d8b502fbc91697d3cdb25b97a23d71a6c4/MoreLinq/Subsets.cs#L116-L120)
- [`Await`](https://github.com/morelinq/MoreLINQ/blob/9e8073d8b502fbc91697d3cdb25b97a23d71a6c4/MoreLinq/Experimental/Await.cs#L431)
- [`Batch`](https://github.com/morelinq/MoreLINQ/blob/9e8073d8b502fbc91697d3cdb25b97a23d71a6c4/MoreLinq/Experimental/Batch.cs#L148) (Experimental)
- [`Memoize`](https://github.com/morelinq/MoreLINQ/blob/9e8073d8b502fbc91697d3cdb25b97a23d71a6c4/MoreLinq/Experimental/Memoize.cs)
- [`Merge`](https://github.com/morelinq/MoreLINQ/blob/9e8073d8b502fbc91697d3cdb25b97a23d71a6c4/MoreLinq/Experimental/Async/Merge.cs#L90)

If we can address those too, as well as some of the feedback, then we should be in a good position to get this merged.

MoreLinq/SortedMerge.cs
```diff
@@ -109,7 +109,7 @@ public static IEnumerable<TSource> SortedMerge<TSource>(this IEnumerable<TSource
             //
             // The algorithm used here will perform N*(K1+K2+...Kn-1) comparisons, where <c>N => otherSequences.Count()+1.</c>
 
-            IEnumerable<TSource> Impl(IEnumerable<IEnumerable<TSource>> sequences)
+            static IEnumerable<TSource> Impl(Func<TSource, TSource, bool> precedenceFunc, IEnumerable<IEnumerable<TSource>> sequences)
```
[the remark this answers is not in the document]
JacobSilasBentley: This has been updated.

MoreLinq/EndsWith.cs
```diff
@@ -78,11 +78,11 @@ public static bool EndsWith<T>(this IEnumerable<T> first, IEnumerable<T> second,
...
                    ? first.TryAsCollectionLike() is { Count: var firstCount } && secondCount > firstCount
                      ? false
-                     : Impl(second, secondCount)
-                   : Impl(secondList = second.ToList(), secondList.Count);
+                     : Impl(second, secondCount, first, comparer)
+                   : Impl(secondList = second.ToList(), secondList.Count, first, comparer);
 #pragma warning restore IDE0075 // Simplify conditional expression
 
-            bool Impl(IEnumerable<T> snd, int count)
+            static bool Impl(IEnumerable<T> snd, int count, IEnumerable<T> first, IEqualityComparer<T> comparer)
```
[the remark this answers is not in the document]
JacobSilasBentley: First has been moved and 'snd' renamed to 'second'

MoreLinq/CountDown.cs
JacobSilasBentley: Removed the unused argument.

MoreLinq/CountDown.cs
JacobSilasBentley: Removed the unused argument.

JacobSilasBentley:
Thank you for the warm welcome and the feedback!

I have hopefully addressed the requested changes and updated the methods that I missed in the initial PR.
 
'Memoize' might take a little bit more attention as unlike the other methods this method is a member of a non-static class. I'm not entirely sure if there is an advantage to making the method static in this scenario. At least one advantage could just be consistency. I would be interested in your opinion. For now, I have made the method static.

Let me know if there are any further issues for me to address.

atifaziz (changes requested):
> I have hopefully addressed the requested changes and updated the methods that I missed in the initial PR.

Thank you for all the follow-up!

> 'Memoize' might take a little bit more attention as unlike the other methods this method is a member of a non-static class. I'm not entirely sure if there is an advantage to making the method static in this scenario. At least one advantage could just be consistency. I would be interested in your opinion. For now, I have made the method static.

Thanks for bringing this case to my attention. I think the right call here may be to simply remove the local function and inline its body. I believe it was separated for clarity rather than consistency and one doesn't need a local function here since `GetEnumerator` has no arguments and therefore no eager validation is needed.

> Let me know if there are any further issues for me to address.

I addressed some minor formatting changes directly (with commits aca7772f0fa569b40d227acb721bd15a81b7d9bf and b6608c448604545f7249d54e267b297a6f4105aa) as they were awkward to point out in the PR review UI.

If we can address `Memoize` then this is good to merge so nearly there! 🏁

viceroypenguin:
> 'Memoize' might take a little bit more attention as unlike the other methods this method is a member of a non-static class. I'm not entirely sure if there is an advantage to making the method static in this scenario. At least one advantage could just be consistency. I would be interested in your opinion. For now, I have made the method static.

When re-implementing this class/iterator, I found that splitting the initialization and enumeration code useful. With two methods, neither of which is static, the whole thing became easier to reason about. There are also some corner cases that are not currently covered, such as a reset and re-initialization happening between two `MoveNext()` calls. You can see my implementation [here](https://github.com/viceroypenguin/SuperLinq/blob/master/Source/SuperLinq/Memoize.cs#L59-L198) and feel free to adapt whatever you feel is best.

JacobSilasBentley:
> Thanks for bringing this case to my attention. I think the right call here may be to simply remove the local function and inline its body. I believe it was separated for clarity rather than consistency and one doesn't need a local function here since `GetEnumerator` has no arguments and therefore no eager validation is needed.

@atifaziz I've changed the method so it is no longer static and following @viceroypenguin suggestion, I've split the initialisation logic and enumeration logic into two different private non-static methods. This has the benefit of the method not being too long. Happy to inline both methods if you would prefer that approach.

@viceroypenguin also mentions: 
>There are also some corner cases that are not currently covered, such as a reset and re-initialization happening between two MoveNext() calls

I'm happy to look into this but I think it would be more appropriate to do this on a separate issue and separate PR

atifaziz:
> Happy to inline both methods if you would prefer that approach.

@JacobSilasBentley Actually, I would completely revert the change and leave `Memoize` alone in this PR. The length of the method isn't a real concern and any complexity is incidental (the split even adds more ceremony). I don't think the closure can be avoid here and exists even with the split. The bigger problem with inlining entirely is that it changes the semantics and behaviour and we don't have a test covering that otherwise it would've failed (would be good to eventually add the test). I feel this part of `Memoize` may need more thought, whether initialisation out to be lazy as it usually is with most iterator methods, but I'd say #906 doesn't apply in its case. It's not where we're looking to save a closure.

> I'm happy to look into this but I think it would be more appropriate to do this on a separate issue and separate PR

Agree, it should be a separate PR, but it would be good to first file those corner cases as demonstratable issues that need prioritising.

JacobSilasBentley:
@atifaziz I have reverted all changes made to the Memoize method made in this PR. 

Let me know if there are any remaining changes required before this code is merged in.

atifaziz (approved):
LGTM and thank you for all your help with contributing this!

Issue #1024, opened 2023-10-25
Review NUnit1030


atifaziz:
In PR #1023, updating the **[NUnit.Analyzers](https://www.nuget.org/packages/NUnit.Analyzers)** to version 3.8.0 introduced [NUnit1030] warnings/errors:

    A:\MoreLINQ\main\MoreLinq.Test\NullArgumentTest.cs(33,31,33,58): error NUnit1030: The TestCaseSource provides type 'NUnit.Framework.Interfaces.ITestCaseData', but the Test method expects type 'System.Action' for parameter 'testCase' (https://github.com/nunit/nunit.analyzers/tree/master/documentation/NUnit1030.md)
    A:\MoreLINQ\main\MoreLinq.Test\NullArgumentTest.cs(37,31,37,60): error NUnit1030: The TestCaseSource provides type 'NUnit.Framework.Interfaces.ITestCaseData', but the Test method expects type 'System.Action' for parameter 'testCase' (https://github.com/nunit/nunit.analyzers/tree/master/documentation/NUnit1030.md)
    A:\MoreLINQ\main\MoreLinq.Test\ReturnTest.cs(151,25,151,51): error NUnit1030: The TestCaseSource provides type 'NUnit.Framework.Interfaces.ITestCaseData', but the Test method expects type 'System.Action' for parameter 'unsupportedAction' (https://github.com/nunit/nunit.analyzers/tree/master/documentation/NUnit1030.md)

Since all tests pass regardless, the severity of NUnit1030 was demoted to suggestion:

https://github.com/morelinq/MoreLINQ/blob/91d3996df623934cf5a666bd34a4b039056aaa98/MoreLinq.Test/.editorconfig#L41-L42

It should be reviewed and ideally restored to a warning.

[NUnit1030]: https://github.com/nunit/nunit.analyzers/blob/master/documentation/NUnit1030.md

PR #1086, opened 2024-10-28
Cleanup NUnit1030

baranyaimate:
Fixes the NUnit1030 warnings.

More details are available in [Issue #1024](https://github.com/morelinq/MoreLINQ/issues/1024).


atifaziz (changes requested):
@baranyaimate I looked into this quickly and it seems like that the only change that's needed is to remove the following two changes to severity:

https://github.com/morelinq/MoreLINQ/blob/03a160ceaa0942f46cf585894a3c6e53df040076/MoreLinq.Test/.editorconfig#L44-L48

and then simply replace `ITestCaseData` with `TestCaseData` in all flagged cases.

PR #1110, opened 2026-03-04
Interleave: add parameterless overload

### LINKED ISSUE #697: [Proposal] Add Interleave overloads.
Given an input of type `IEnumerable<IEnumerable<T>>` its hard to call `Interleave` on it:
```C#
Enumerable.Empty<T>.Interleave(input);
input.First().Interleave(input.Skip(1));
```

I propose to add an `Interleave` overload as a classic static method:
```C#
 public static IEnumerable<T> Interleave<T>(IEnumerable<IEnumerable<T>> otherSequences);
```

Or as an extension method:
```C#
 public static IEnumerable<T> Interleave<T>(this IEnumerable<IEnumerable<T>> otherSequences);
```

We can also add an overload with the `SelecMany` pattern:
```C#
 public static IEnumerable<TResult> Interleave<TSource, TResult>(this IEnumerable<TSource> source, Func<TSource, IEnumerable<TResult>> selector);
```

fowl2:
fixes https://github.com/morelinq/MoreLINQ/issues/697

### DIFF
--- a/MoreLinq/Extensions.g.cs
+++ b/MoreLinq/Extensions.g.cs
@@ -3081,6 +3081,32 @@ namespace MoreLinq.Extensions
     [GeneratedCode("MoreLinq.ExtensionsGenerator", "1.0.0.0")]
     public static partial class InterleaveExtension
     {
+
+        /// <summary>
+        /// Interleaves the elements of a sequence of sequences into a single sequence.
+        /// </summary>
+        /// <typeparam name="T">The type of the elements of the source sequences.</typeparam>
+        /// <param name="sequences">The sequences to interleave</param>
+        /// <returns>A sequence of interleaved elements from all of the source sequences.</returns>
+        /// <remarks>
+        /// <para>
+        /// Interleave combines sequences by visiting each in turn, and returning the first element
+        /// of each, followed by the second, then the third, and so on. So, for example:</para>
+        /// <code><![CDATA[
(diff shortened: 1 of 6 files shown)


atifaziz (changes requested):
Thanks, @fowl2.

Unfortunately, I think that the overload introduces ambiguity, which requires an explicit type argument for resolution, e.g.:

```c#
var sequences = new[]
{
    new[] { 0, 0, 0, 0, 0, 0 },
    new[] { 1, 1, 1, 1 }
};
IEnumerable<IEnumerable<int>> result1 = sequences.Interleave();
IEnumerable<int> result2 = sequences.Interleave<int>();
```

However, when `T` is an anonymous type, the type argument becomes impossible to express.

In hindsight, perhaps the `params` overload should have been defined to require at least one other sequence (because just calling `Interleave` on a sequence alone would be silly):

```c#
public static IEnumerable<T> Interleave<T>(this IEnumerable<T> first,
                                           IEnumerable<T> second,
                                           params IEnumerable<T>[] others)
```

Adding such an overload now would not help and adjusting the existing one would be a breaking change so would need to be deferred to the next major version.

The most workable solution would be adding the following overload that @Orace also suggested in #697:

```c#
public static IEnumerable<TResult> Interleave<T, TResult>(this IEnumerable<T> source,
                                                          Func<T, IEnumerable<TResult>> selector)
```

It would require always providing an identity function, a lambda `s => s`, when `source` is already a sequence of sequences.

The CI is failing because the generated extensions need to be refreshed by running either `build.cmd` or `build.sh` (depending on your OS/shell) and committing the change.

Also, tests would need to be added for the new overload before this can be considered for merging.

